PR #4, opened 2018-11-28
feat: add plugin parameter auth url

karniwl:
adding plugin origin parameter in url for analytics purposes.


Kirill89:
LGTM, but first let’s finish and merge PR on server-side.

Kirill89:
And maybe make this argument a bit more universal? Something like ‘&from=intellijPlugin’?

adrukh:
Suggest to merge it sooner than later to get on the install train - the sooner this is out, the more users will have it in any given time in the future. :)

PR #18, opened 2020-01-27
Feat: add red underline annotator for maven vulnerabilities

aldanchenko:
This will add a red underline for Maven vulnerabilities.

Screen Shot 2020-01-27 at 6 51 09 PM

### DIFF
--- a/src/main/scala/io/snyk/plugin/ui/annotators/SnykMavenRedUnderlineAnnotator.scala
+++ b/src/main/scala/io/snyk/plugin/ui/annotators/SnykMavenRedUnderlineAnnotator.scala
@@ -1,0 +1,41 @@
+package io.snyk.plugin.ui.annotators
+
+import com.intellij.lang.annotation.{AnnotationHolder, Annotator}
+import com.intellij.openapi.diagnostic.Logger
+import com.intellij.psi.PsiElement
+import io.snyk.plugin.IntellijLogging.ScalaLogger
+import io.snyk.plugin.datamodel.SecurityVuln
+import io.snyk.plugin.ui.state.SnykPluginState
+
+/**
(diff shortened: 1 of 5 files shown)


src/main/scala/io/snyk/plugin/ui/annotators/SnykMavenRedUnderlineAnnotator.scala
```diff
@@ -0,0 +1,41 @@
...
+
+      val vulnerabilityName = securityVulnerability.name
+      val vulnerabilityNameParts = vulnerabilityName.split(":")
+
+      vulnerabilityNameParts.foreach(name => {
+        if (element.getText == name) {
+          import com.intellij.openapi.util.TextRange
+
+          val range = new TextRange(element.getTextRange.getStartOffset, element.getTextRange.getEndOffset)
+          holder.createErrorAnnotation(range, "Vulnerability issue")
```
yuliabaron: Where do I see this error text?

aldanchenko: When a user hovers over dependency text.

Screen Shot 2020-01-28 at 4 19 43 PM

I think we can add a more useful description. For example, in this case, show "XML External Entity (XXE) Injection" message.

yuliabaron: Maybe then `Vulnerable package`?

aldanchenko: Ok. I'll change it.

PR #20, opened 2020-01-29
Feat: replace maven api call with snyk cli call

aldanchenko:
This version replaces Maven API call using JSON payload of all dependencies with Snyk CLI call.

### DIFF
--- a/src/main/scala/io/snyk/plugin/client/ApiClient.scala
+++ b/src/main/scala/io/snyk/plugin/client/ApiClient.scala
@@ -1,9 +1,14 @@
 package io.snyk.plugin.client
 
+import java.io.FileNotFoundException
 import java.net.URI
+import java.nio.charset.Charset
 import java.time.OffsetDateTime
+import java.util
 import java.util.UUID
 
+import com.intellij.execution.configurations.GeneralCommandLine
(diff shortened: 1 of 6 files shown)


src/main/scala/io/snyk/plugin/client/ApiClient.scala
```diff
@@ -72,41 +79,33 @@ private final class StandardApiClient(tryConfig: => Try[SnykConfig]) extends Api
...
-    val jsonReq = stringifyWithoutNulls(SnykClientSerialisation.encodeRoot(treeRoot, config.org))
+  private def runRaw(project: Project, treeRoot: SnykMavenArtifact): Try[String] = tryConfig flatMap { config =>
+    log.debug("ApiClient: run Snyk CLI")
 
-    log.debug("ApiClient: Built JSON Request")
-    log.debug(jsonReq)
+    val projectPath = project.getBasePath
 
-    val apiEndpoint = config.endpointOrDefault
-    val apiToken = config.api
```
pavel-github: we don't use the token anymore, so a user has to run `snyk auth` first, correct?

aldanchenko: Good point Pavel. Thanks!

PR #24, opened 2020-02-25
Fix: add display cli error message

aldanchenko:
Fixed display of long text errors.
Added message how can user can fix the error in the case of Maven projects.

bold and italic example


aldanchenko:
@pavel-snyk made a new, cleaner PR.

PR #25, opened 2020-03-01
Fix: add display cli error message

aldanchenko:
Fixed display of long text errors.
Added message how can user can fix the error in the case of Maven projects.

75274212-1d703980-580b-11ea-97c2-3a24fec2ab0a


aldanchenko:
@pavel-snyk could you, please, check this PR? It does not have issues with formatting (I hope).

PR #35, opened 2020-03-19
Feat: add gradle annotator

aldanchenko:
Added Gradle annotator.
Fixed display of red underlines for Maven.

Снимок экрана 2020-03-19 в 1 03 29 PM


src/main/scala/io/snyk/plugin/ui/annotators/SnykMavenRedUnderlineAnnotator.scala
```diff
@@ -1,49 +1,164 @@
...
+    annotationInfos.foreach(annotationInfo => {
+      val parentPsiElement = annotationInfo.psiElement
+
+      if (parentPsiElement != null) {
+        parentPsiElement.accept(new PsiRecursiveElementWalkingVisitor() {
+          override def visitElement(element: PsiElement): Unit = {
+            if (element.isInstanceOf[XmlText]) {
+              val elementText = element.getText
+
+              if (annotationInfo.isArtifactIdMatch(elementText)
```
yuliabaron: why is OR and not AND? Don't you need exact match of artifact and its version?

PR #44, opened 2020-04-28
docs: add information in overview that CLI must be installed for v2

pavel-snyk:
This PR adds additional information in plugin overview about CLI installation for v2.

### DIFF
--- a/src/main/resources/META-INF/plugin.xml
+++ b/src/main/resources/META-INF/plugin.xml
@@ -30,6 +30,13 @@
         <li>Maven</li>
         <li>Gradle</li>
     </ul>
+
+    <h2><strong>Prerequisites:</strong></h2>
+    <p>
+        To use the Snyk Vulnerability Scanner plugin, you must have Snyk CLI installed on your machine.
+        <a href="https://support.snyk.io/hc/en-us/articles/360003812538-Install-the-Snyk-CLI">See how to install</a>
+        Snyk CLI on our documentation page.
+    </p>
     ]]></description>
 
     <change-notes><![CDATA[


src/main/resources/META-INF/plugin.xml
```diff
@@ -30,6 +30,13 @@
         <li>Maven</li>
         <li>Gradle</li>
     </ul>
+
+    <h2><strong>Prerequisites:</strong></h2>
+    <p>
+        To use the Snyk Vulnerability Scanner plugin, you must have Synk CLI installed on your machine.
```
JackuB, line 36:
```suggestion
        To use the Snyk Vulnerability Scanner plugin, you must have Snyk CLI installed on your machine.
```

src/main/resources/META-INF/plugin.xml
```diff
@@ -30,6 +30,13 @@
...
+        <a href="https://support.snyk.io/hc/en-us/articles/360003812538-Install-the-Snyk-CLI">Click here</a>
```
JackuB, line 37: Just a nit: it's usually a bad practice to use "[click] here" as a link text. You should describe what the link is - See https://www.w3.org/QA/Tips/noClickHere / https://uxmovement.com/content/why-your-links-should-never-say-click-here/

src/main/resources/META-INF/plugin.xml
pavel-snyk: 🤦 , shame on me.

pavel-snyk: fixed with 049e8eb

src/main/resources/META-INF/plugin.xml
pavel-snyk: done with 049e8eb

PR #49, opened 2020-05-28
Feat: add settings

aldanchenko:
Added settings dialog.
Tests fixed.

Снимок экрана 2020-05-28 в 1 04 29 PM

### DIFF
--- a/src/main/scala/io/snyk/plugin/ui/settings/SettingsDialog.scala
+++ b/src/main/scala/io/snyk/plugin/ui/settings/SettingsDialog.scala
@@ -1,0 +1,198 @@
+package io.snyk.plugin.ui.settings
+
+import java.net.URL
+
+import com.intellij.openapi.ui.{ComponentValidator, ValidationInfo}
+import javax.swing.{JCheckBox, JComponent, JLabel, JPanel, JTextField}
+import java.awt.{Dimension, Insets}
+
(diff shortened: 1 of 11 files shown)


src/main/scala/io/snyk/plugin/ui/settings/SettingsDialog.scala
```diff
@@ -0,0 +1,198 @@
...
+  private def isIgnoreUnknownCAModified =
+    ignoreUnknownCACheckBox.isSelected != persistentState.isIgnoreUnknownCA()
+
+  private def setupUI(): Unit = {
+    val gridLayoutManager = new UIGridLayoutManager(5, 2, new Insets(0, 0, 0, 0), -1, -1)
+
+    rootPanel.setLayout(gridLayoutManager)
+
+    val customEndpointLabel = new JLabel
+    customEndpointLabel.setText("Custom Endpoint:")
```
pavel-snyk, line 70: Should it be "Custom endpoint:" without capitalization the second word?
I checked UI guidelines from JetBrains about capitalization https://jetbrains.github.io/ui/text/capitalization/#sentence and I think this is our case.

aldanchenko: Good point @pavel-snyk 👍

PR #66, opened 2020-09-14
chore: patch plugin.xml with latest changelog item

pavel-snyk:
This PR adds latest changelog item in `plugin.xml` during `patchChangeLog` task.

Example screenshot:
Screen Shot 2020-09-14 at 12 01 32

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Snyk Vulnerability Scanner Changelog
 
-## [Unreleased]
+## [2.0.1]
+### Added
+- Propagate integration name for JetBrain IDEs
 
 ## [2.0.0]
 ### Added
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # Snyk Vulnerability Scanner Changelog
 
-## [Unreleased]
+## [2.0.1]
```
JackuB, line 3: Version numbers are something that we need to maintain manually for now?

JackuB (approved):

CHANGELOG.md
pavel-snyk: yes for now. https://github.com/JetBrains/intellij-platform-plugin-template uses it, so I don't want to change official way to handle changelog.
p.s.: perhaps later we'll add some automation.

PR #67, opened 2020-09-16
Fix: replaced text field with password field for token.

aldanchenko:
Replaced token text field with a password field.

token settings


pavel-snyk (approved):
LGTM

PR #68, opened 2020-09-22
chore: improved Icons class to use in plugin configuration files.

aldanchenko:
Moved Icons class to icons package and rename to SnykIcons to use in plugin configuration files.


pavel-snyk (approved):
LGTM,
@aldanchenko , good catch with `JvmStatic`.

PR #74, opened 2020-11-14
feat: Naively split the additionalParameters value so the CommandLineRunner sends them correctly as separate arguments

vicmac:
Had a bunch of issues with Snyk plugin for Android Studio since our setup requires the usage of several flags (--configuration-matches, --severity-threshold and --sub-project)

Discussed this with Trevor McCord over several emails since we're evaluating getting on board with Snyk. Discovered that SnykCLIService bundles them up in a single argument and apparently only the first arg is correctly parsed and the rest gets ignored or throws an error.

Android-studio log
`2020-11-12 10:27:02,635 [60825005]   INFO - plugin.services.SnykCliService - Cli parameters: [/Users/vmacias/Library/Application Support/Google/AndroidStudio4.1/plugins/snyk-intellij-plugin/snyk-macos, --json, --org=billpocket, --severity-threshold=low --sub-project=billpocket --configuration-matching=".*BillpocketProdSyncImages.*$", test]`

This PR naively splits the additionalParameters string at every whitespace. Didn't want to introduce dependencies for arg parsing.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,5 +1,5 @@
 pluginVersion=2.0.3
-pluginSinceBuild=202
+pluginSinceBuild=201
 pluginUntilBuild=202.*
 #
 platformVersion=2020.2
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,5 +1,5 @@
 pluginVersion=2.0.3
-pluginSinceBuild=201
+pluginSinceBuild=202
 pluginUntilBuild=202.*
 #
 platformVersion=2020.2
--- a/src/main/kotlin/io/snyk/plugin/services/SnykCliService.kt
+++ b/src/main/kotlin/io/snyk/plugin/services/SnykCliService.kt
@@ -112,11 +112,8 @@ class SnykCliService(val project: Project) {
 
         val additionalParameters = settings.getAdditionalParameters(project)
 
-        if (additionalParameters != null && additionalParameters.isNotEmpty()) {
-
-            val additionalParamsArray = splitParameters(additionalParameters)
-            commands.addAll(additionalParamsArray)
-
+        if (additionalParameters != null && additionalParameters.trim().isNotEmpty()) {
+            commands.addAll(additionalParameters.trim().split(" "))
         }
 
         commands.add("test")
@@ -148,10 +145,4 @@ class SnykCliService(val project: Project) {
     }
 
     fun isPackageJsonExists(): Boolean = File(project.basePath!!, "package.json").exists()
-
-    fun splitParameters(params: String) : List<String> {
-
-        return params.split(" ")
-
-    }
 }
--- a/src/main/kotlin/io/snyk/plugin/services/SnykCliService.kt
+++ b/src/main/kotlin/io/snyk/plugin/services/SnykCliService.kt
@@ -113,7 +113,10 @@ class SnykCliService(val project: Project) {
         val additionalParameters = settings.getAdditionalParameters(project)
 
         if (additionalParameters != null && additionalParameters.isNotEmpty()) {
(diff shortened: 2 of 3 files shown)


gradle.properties
```diff
@@ -1,5 +1,5 @@
 pluginVersion=2.0.3
-pluginSinceBuild=202
+pluginSinceBuild=201
```
pavel-snyk, line 2: We have two plugin versions deployed in JB plugin portal: first is until 2020.1 (201) included and the second is from 2020.2 (202+). So we cannot just decrease compatibility matrix.
I guess this change is because of Android Studio (based on 2020.1). What I suggest whether wait for a new AS based on 2020.2 (should be available earlier next year), or build s specific version with this changes locally.

pavel-snyk (changes requested):
👋 @vicmac,
thanks for your PR, it's great to see contributions in our repo 😉 .
I left some comments please take a look on it.

@aldanchenko , can you take a look on code too?

src/main/kotlin/io/snyk/plugin/services/SnykCliService.kt
```diff
@@ -113,7 +113,10 @@ class SnykCliService(val project: Project) {
         val additionalParameters = settings.getAdditionalParameters(project)
 
         if (additionalParameters != null && additionalParameters.isNotEmpty()) {
-            commands.add(additionalParameters)
+
+            val additionalParamsArray = splitParameters(additionalParameters)
```
aldanchenko, line 117: @vicmac thanks to pointing us on very important issue 👍! The only thing I could suggest is to simplify this part and just pass additionalParameters.split(" ") as parameter. Like 

```
if (additionalParameters != null && additionalParameters.isNotEmpty()) {
      commands.addAll(additionalParameters.split(" "))
}
```

Thanks!

vicmac: Thanks @aldanchenko. I resolved to put that into a separate function in case in the future you might want to:

1. Apply more complex logic to the parameter-splitting routine
2. Test the routine independently

If you think these are not a concern right now, I'll update the PR.

I was also thinking about updating the split expression to "` --`" since all params use the `--` prefix, I think that way it might be less prone to break and send a weird error messages if any extra spaces are incorrectly added to the additionalParams field.

gradle.properties
vicmac: Thanks @pavel-snyk 

Got it! I thought it would be great to backport this feature for current Snyk users in Android Studio. I understand the compatibility matrix issue and will revert the change in an update

pavel-snyk: 👍.
p.s: I checked release cycle by AS, it looks like next version would be Jan-Feb 2021.

src/main/kotlin/io/snyk/plugin/services/SnykCliService.kt
pavel-snyk: two cents from my side: CLI argument parsing is tricky (e.g. `--debug` can be `-d`). My suggestions would be:
1) split by whitespace
2) lets do the simplified version (without a separate function). We plan to add some improvements across all our Java integrations (this will resolve issue with args parsing). And without complex logic testing standard split function is not necessary.

Thanks both @vicmac and @aldanchenko.

vicmac: Makes sense, let's keep the whitespace char as delimiter

vicmac (commented):
Comments and suggestions were addressed to inline the arg-splitting logic  and keep it using whitespace as delimiter

pavel-snyk (approved):

PR #82, opened 2021-01-08
fix: handle custom endpoint option correctly

pavel-snyk:
This PR fixes handling custom endpoint (`--api`) option. The behavior was changed in CLI (https://github.com/snyk/snyk/blob/master/src/lib/config.ts#L10), it should be capitalized or use envvar.


aldanchenko (approved):
LGTM 👍

PR #85, opened 2021-02-01
Deepcode addition (as SnykCode)

ArtsiomCh:


ArtsiomCh:
For now (2.0.5-dev7) to build plugin: https://github.com/snyk/snyk-code-client (ex deepcode java-client ) required as fatJar (run shadowJar task for snyk-code-client) placed into `/libs` dir in the root of snyk-intellij-plugin

src/main/java/io/snyk/plugin/ui/toolwindow/SnykAuthPanel.form
```diff
@@ -0,0 +1,77 @@
...
+            <properties>
+              <icon value="icons/logo_snyk.png"/>
+              <text value=""/>
+            </properties>
+          </component>
+        </children>
+      </grid>
+    </children>
+  </grid>
+</form>
```
pavel-snyk, line 77: Do we need this file?

src/main/java/io/snyk/plugin/ui/toolwindow/SnykAuthPanel.java
```diff
@@ -0,0 +1,32 @@
...
+      project.getMessageBus().syncPublisher(SnykCliDownloadListener.Companion.getCLI_DOWNLOAD_TOPIC())
+        .checkCliExistsFinished();
+    });
+  }
+
+  public JPanel getRoot() {
+    return rootPanel;
+  }
+
+}
```
pavel-snyk, line 32: Do we need this file?

src/main/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanel.kt
```diff
@@ -36,239 +51,475 @@ import javax.swing.tree.DefaultTreeModel
...
 
-            override fun scanError(cliError: CliError) =
-                ApplicationManager.getApplication().invokeLater { displayError(cliError) }
-        })
+                override fun scanningSnykCodeFinished(snykCodeResults: SnykCodeResults) =
+                    ApplicationManager.getApplication().invokeLater { displaySnykCodeResults(snykCodeResults) }
+
+                override fun scanError(cliError: CliError) =
+                    ApplicationManager.getApplication().invokeLater {
+                        if (cliError.message == "Authentication failed. Please check the API token on https://snyk.io") {
```
pavel-snyk, line 158: This will work for producation SaaS only (and not for dev or on-prem instances).
How about don't match `https://snyk.io`?

src/main/kotlin/io/snyk/plugin/cli/Vulnerability.kt
```diff
@@ -33,4 +33,12 @@ class Vulnerability {
             ""
         }
     }
+
+    fun getSeverityIndex(): Int =
+        when (severity) {
+            "high" -> 3
+            "medium" -> 2
+            "low" -> 1
+            else -> 0
```
pavel-snyk, line 42: I see we use it for sorting later. How about using comparator for tree nodes? Issue I see atm: this code cannot handle other issues e.g. `critical` if this will come later.

.github/workflows/build.yml
```diff
@@ -16,6 +16,10 @@ jobs:
     name: Test
     needs: gradleValidation
     runs-on: ubuntu-latest
+    env:
+      USERNAME_GITHUB: ${{ github.actor }}
+      READ_TOKEN_GITHUB: ${{ github.token }}
+      DEEPCODE_API_KEY: ${{ secrets.DEEPCODE_API_KEY }}
```
pavel-snyk, line 22: Why do we need Deepcode api key?

src/main/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanel.kt
ArtsiomCh: check CLI output with custom endpoint

src/main/kotlin/io/snyk/plugin/SnykBulkFileListener.kt
```diff
@@ -0,0 +1,159 @@
...
+    // proceed added and changed files
+    override fun after(events: MutableList<out VFileEvent>) {
+//        cleanCaches(
+//            events, listOf(
+//                VFileContentChangeEvent::class.java,
+//                VFileMoveEvent::class.java,
+//                VFileCopyEvent::class.java,
+//                VFileCreateEvent::class.java
+//            )
+//        )
```
pavel-snyk, line 28: Let's don't merge commented code

src/main/kotlin/io/snyk/plugin/SnykBulkFileListener.kt
```diff
@@ -0,0 +1,159 @@
...
+                            project,
+                            "Updating cache on server (for ${filesToRemoveFromCache.size} locally changed files)..."
+                        ) { progress ->
+                            AnalysisData.instance.updateCachedResultsForFiles(
+                                project,
+                                Collections.emptyList(),
+                                filesToRemoveFromCache,
+                                progress);
+                        }
+*/
```
pavel-snyk, line 105: The same as previous with commented code.

src/main/kotlin/io/snyk/plugin/cli/Vulnerability.kt
ArtsiomCh: Don't we need to handle the case with `critical` severity anyway inside those comparator? 
Then why to add one more level of abstraction here?

.github/workflows/build.yml
ArtsiomCh: looks like it was used before when (hardcoded) deepcode token was needed to init the plugin
probably could be removed (together with github tokens as well, used for java-client library resolving from guthub afaik) 
but you better check, you know better how that magic works :)

src/main/kotlin/io/snyk/plugin/SnykBulkFileListener.kt
ArtsiomCh: we probably will need it in the future and if not commit commented code it will be lost. 
would be happy if there is better way to not lost this code.

pavel-snyk: git stash or local branch with changes?

aldanchenko (approved):
LGTM :)

src/main/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanel.kt
ArtsiomCh: fixed. Previously it seems there always have be the link to `https://snyk.io` in that error message even with custom endpoint. It's not like that now, probably I did something wrong.

src/main/kotlin/io/snyk/plugin/cli/Vulnerability.kt
ArtsiomCh: left //todo there for the future

.github/workflows/build.yml
ArtsiomCh: did remove that, hopefully nothing will be broken :)

src/main/java/io/snyk/plugin/ui/toolwindow/SnykAuthPanel.form
ArtsiomCh: yes for now, will rewrite if later

src/main/java/io/snyk/plugin/ui/toolwindow/SnykAuthPanel.java
ArtsiomCh: yes for now, will rewrite if later

pavel-snyk (approved):
Friendly reminder: as we agreed next time the goal is to merge simple features/bugfixes instead of big chunks of code.

PR #92, opened 2021-03-25
feat: analytics integration

pavel-snyk:
This PR adds integration with Segment analytics.

### DIFF
--- a/src/main/kotlin/io/snyk/plugin/analytics/EventPropertiesProvider.kt
+++ b/src/main/kotlin/io/snyk/plugin/analytics/EventPropertiesProvider.kt
@@ -1,0 +1,60 @@
+package io.snyk.plugin.analytics
+
+import ai.deepcode.javaclient.core.SuggestionForFile
+import io.snyk.plugin.cli.CliResult
+import io.snyk.plugin.cli.Vulnerability
+import io.snyk.plugin.services.SnykApplicationSettingsStateService
+import io.snyk.plugin.snykcode.SnykCodeResults
+import io.snyk.plugin.snykcode.severityAsString
+
+object EventPropertiesProvider {
+    fun getSelectedProducts(settings: SnykApplicationSettingsStateService): EventProperties {
+        val selectedProducts = mutableListOf<String>()
+
+        if (settings.cliScanEnable) selectedProducts += "Snyk Open Source"
+        if (settings.snykCodeSecurityIssuesScanEnable) selectedProducts += "Snyk Code Security"
+        if (settings.snykCodeQualityIssuesScanEnable) selectedProducts += "Snyk Code Quality"
+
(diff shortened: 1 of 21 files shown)


src/main/kotlin/io/snyk/plugin/analytics/internal/Snyk.kt
```diff
@@ -0,0 +1,48 @@
...
+
+import okhttp3.OkHttpClient
+import retrofit2.Retrofit
+import retrofit2.converter.gson.GsonConverterFactory
+import java.util.concurrent.TimeUnit
+
+/**
+ * Main entrypoint for using Snyk API.
+ */
+class Snyk private constructor(
```
aldanchenko, line 11: Just as a suggestion: maybe  this name is too wide mining? Maybe something like something SnykSegment or SnykAnalytics will be more specific?

src/main/kotlin/io/snyk/plugin/analytics/EventPropertiesProvider.kt
```diff
@@ -0,0 +1,60 @@
+package io.snyk.plugin.analytics
+
+import ai.deepcode.javaclient.core.SuggestionForFile
+import io.snyk.plugin.cli.CliResult
+import io.snyk.plugin.cli.Vulnerability
+import io.snyk.plugin.services.SnykApplicationSettingsStateService
+import io.snyk.plugin.snykcode.SnykCodeResults
+import io.snyk.plugin.snykcode.severityAsString
+
+object EventPropertiesProvider {
```
aldanchenko, line 10: Just a suggestion: maybe use just functions? Like we have here  [Utils.kt](https://github.com/snyk/snyk-intellij-plugin/blob/master/src/main/kotlin/io/snyk/plugin/Utils.kt)

aldanchenko (approved):
LGTM 😃 Great job! 👍

src/main/kotlin/io/snyk/plugin/analytics/internal/Snyk.kt
pavel-snyk: Good catch, totally agree with you.

src/main/kotlin/io/snyk/plugin/analytics/EventPropertiesProvider.kt
pavel-snyk: Discussed together about context. We leave it for now until analytics will be extracted into a separate library.
Thanks @aldanchenko

src/main/kotlin/io/snyk/plugin/analytics/internal/Snyk.kt
pavel-snyk: resolved with d1b7bb2

PR #104, opened 2021-04-24
fix: .dcignore and .gitignore parsing and usage (ROADRUNNER-55)

ArtsiomCh:
- to fully match https://git-scm.com/docs/gitignore#_pattern_format and glob syntax rules (https://docs.oracle.com/javase/8/docs/api/java/nio/file/FileSystem.html#getPathMatcher-java.lang.String-);
- caches invalidation and re-parsing on those files change;
- generic .dcignore added on project opening if no .gitignore/.dcignore exist;
- functional tests for IgnoreInfoHolder related basic functionality.

### DIFF
--- a/src/main/kotlin/io/snyk/plugin/SnykBulkFileListener.kt
+++ b/src/main/kotlin/io/snyk/plugin/SnykBulkFileListener.kt
@@ -91,14 +91,7 @@ class SnykBulkFileListener() : BulkFileListener {
                 }
 
             // clean .dcignore caches if needed
-            val ignoreFilesChanged = virtualFilesAffected
-                .filter { it.name == ".dcignore" || it.name == ".gitignore" }
-                .distinct()
-            for (virtualFile in ignoreFilesChanged) {
-                val ignoreFile = PsiManager.getInstance(project).findFile(virtualFile) ?: continue
-                SnykCodeIgnoreInfoHolder.instance.remove_ignoreFileContent(ignoreFile)
-                AnalysisData.instance.removeProjectFromCaches(project)
-            }
+            SnykCodeIgnoreInfoHolder.instance.cleanIgnoreFileCachesIfAffected(project, virtualFilesAffected)
         }
     }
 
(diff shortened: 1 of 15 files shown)


amir-mehler:
I dont know why im reviewing it. Got no clue here. But I trust you.

amir-mehler (approved):

src/main/resources/full.dcignore
```diff
@@ -0,0 +1,1547 @@
+# Write glob rules for ignored files.
```
deepcodeg, line 1: @ArtsiomCh thanks for the pr, I recalled chatting about getting only Java, JS and Pythong ignores initially. From what I understand I see all ignores here. Could you please shed some light on this?

src/main/kotlin/io/snyk/plugin/SnykBulkFileListener.kt
```diff
@@ -71,14 +81,49 @@ class SnykBulkFileListener() : BulkFileListener {
...
+                project.service<SnykTaskQueueService>().scheduleRunnable("Snyk Code is updating caches...") {
                     if (filesToRemoveFromCache.size > 10) {
                         // bulk files change event (like `git checkout`) - better to drop cache and perform full rescan later
                         AnalysisData.instance.removeProjectFromCaches(project)
                     } else {
                         AnalysisData.instance.removeFilesFromCache(filesToRemoveFromCache)
                     }
                 }
+
+            // clean .dcignore caches if needed
```
aldanchenko, line 93: Maybe it would be better to "hide" all logic related to dc ignore functionality in a separate class? Something like DcIgnoreService?

src/main/kotlin/io/snyk/plugin/SnykBulkFileListener.kt
```diff
@@ -71,14 +81,49 @@ class SnykBulkFileListener() : BulkFileListener {
...
+            if (project.isDisposed) continue
+
+            val virtualFilesAffected = getAffectedVirtualFiles(
+                events,
+                fileFilter = Predicate { true },
+                classesOfEventsToFilter = classesOfEventsToFilter
+            )
+
+            // update .dcignore caches if needed
+            val ignoreFilesChanged = virtualFilesAffected
```
aldanchenko, line 116: Same suggestion here

src/main/kotlin/io/snyk/plugin/SnykPostStartupActivity.kt
```diff
@@ -23,9 +28,22 @@ class SnykPostStartupActivity : StartupActivity.DumbAware {
             listenersActivated = true
         }
 
+        createDcIgnoreIfNeeded(project)
```
aldanchenko, line 31: Same suggestion for this method here :)

src/main/resources/full.dcignore
ArtsiomCh: Found only Java and Python "default" .gitignore here https://github.com/github/gitignore
Looks like for JS for each particular framework - specific .gitignore is used.
Also if we create our own "default" for IntelliJ plugin .dcignore it will be different from the one currently used in VSCE and that might confuse users.
Let me know what you think: shell we stay with current default .dcignore (same as VSCE) or create our own and if latter then which specific JS .gitignore should we use?

ArtsiomCh: @deepcodeg

ArtsiomCh: so i'm leaving default .dcignore as is. That seems to be most reasonable solution.

deepcodeg: Didn't get any messages from github before the merge, strange. Let's see how it goes with the default one and we could change it later if needed.

aldanchenko (approved):
Great work! 👍

PR #105, opened 2021-04-28
fix: `missing file parameter` exception [ROADRUNNER-111]

ArtsiomCh:
will fix https://snyksec.atlassian.net/browse/ROADRUNNER-111

### DIFF
--- a/src/main/kotlin/io/snyk/plugin/snykcode/core/PDU.kt
+++ b/src/main/kotlin/io/snyk/plugin/snykcode/core/PDU.kt
@@ -30,12 +30,17 @@ class PDU private constructor() : PlatformDependentUtilsBase() {
     override fun getProjectBasedFilePath(file: Any): String {
         val psiFile = toPsiFile(file)
         // looks like we don't need ReadAction for this (?)
-        var absolutePath = psiFile.virtualFile.path
+        val absolutePath = psiFile.virtualFile.path
         val projectPath = psiFile.project.basePath
-        if (projectPath != null) {
-            absolutePath = absolutePath.replace(projectPath, "")
+            ?: throw IllegalStateException("No Project base Path found for file $psiFile")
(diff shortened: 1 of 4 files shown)


src/main/kotlin/io/snyk/plugin/snykcode/core/PDU.kt
```diff
@@ -30,12 +30,17 @@ class PDU private constructor() : PlatformDependentUtilsBase() {
...
         val psiFile = toPsiFile(file)
         // looks like we don't need ReadAction for this (?)
-        var absolutePath = psiFile.virtualFile.path
+        val absolutePath = psiFile.virtualFile.path
         val projectPath = psiFile.project.basePath
-        if (projectPath != null) {
-            absolutePath = absolutePath.replace(projectPath, "")
+            ?: throw IllegalStateException("No Project base Path found for file $psiFile")
+        // `ignoreCase = true` needed due to https://youtrack.jetbrains.com/issue/IDEA-268081
+        val projectBasedPath = absolutePath.replaceFirst(projectPath, "", ignoreCase = true)
```
aldanchenko, line 37: Just curious. Could this produce unexpected issues if user OS is case sensitive and he has, for example,

/home/user/projects/testproj/
/home/user/projects/TestProj/

?

aldanchenko (approved):
LGTM. Great job Artsiom! 👍

src/main/kotlin/io/snyk/plugin/snykcode/core/PDU.kt
ArtsiomCh: It should not, as we comparing projectPath against projectPathWithFilePathInside, so technically speaking every time when those comparison fail is an Exception (that's why we have IllegalStateException and not just log.warn)

aldanchenko: Great work! Thanks for clarifications! 👍

PR #106, opened 2021-04-29
fix: error when messageBus on disposed Project is called [ROADRUNNER-110]

ArtsiomCh:
Same thing is needed for every `project.service<>()` call in any background task (due to Project could be disposed anytime before our code invocation) 
And probably for most(all?) Project method call. 
Will be handled in other PRs

### DIFF
--- a/src/main/kotlin/io/snyk/plugin/Utils.kt
+++ b/src/main/kotlin/io/snyk/plugin/Utils.kt
@@ -3,6 +3,7 @@ package io.snyk.plugin
 import com.intellij.openapi.application.PathManager
 import com.intellij.openapi.components.service
 import com.intellij.openapi.project.Project
+import com.intellij.util.messages.Topic
 import io.snyk.plugin.cli.Platform
 import io.snyk.plugin.services.SnykCliService
 import io.snyk.plugin.services.SnykApplicationSettingsStateService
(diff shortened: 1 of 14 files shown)


src/main/kotlin/io/snyk/plugin/Utils.kt
```diff
@@ -26,6 +27,12 @@ fun getPluginPath() = PathManager.getPluginsPath() + "/snyk-intellij-plugin"
 
 fun isProjectSettingsAvailable(project: Project?) = nonNull(project) && !project!!.isDefault
 
+fun <L> getSyncPublisher(project: Project, topic: Topic<L>): L? {
```
aldanchenko, line 30: As a suggestion: we could hide this function in service there also could be hidden different publish tokens. Like this

`@Service
class SnykSyncPublisherService(val project: Project) {

    private fun <L> getSyncPublisher(topic: Topic<L>): L? {
        val messageBus = project.messageBus
        if (messageBus.isDisposed) return null
        return messageBus.syncPublisher(topic)
    }

    fun scanningSnykCodeError(cliError: CliError) = getSyncPublisher(SnykScanListener.SNYK_SCAN_TOPIC)?.scanningSnykCodeError(cliError)

    fun scanningStarted() = getSyncPublisher(SnykScanListener.SNYK_SCAN_TOPIC)?.scanningStarted()
}`

And if add simple method to Utils like 

`fun getSyncPublisher(project: Project): SnykSyncPublisherService = project.service()`

we could use it like 

`getSyncPublisher(prj).scanningSnykCodeError(
                CliError(false, message, prj.basePath ?: "")
            )`
            
What do you think?

ArtsiomCh: May have sense for hiding our Topics from the api caller, but also will reduce flexibility in usage other (Idea's) Topics, as we'll have to create separate method for every such Topic.
IMHO such an extra layer of abstraction (proxy) bring a little value but increase the complexity of the code.
Also `project.service()` usage is not safe and will need the same disposability check (see PR's notes) - that means even more complexity to understand for the code readers.
I would probably try to use Services only if it's really necessary and convenient. And use other approach (simple-stupid singleton or util's method like here) if it's possible.

aldanchenko:
Thanks, Artsiom! Great work!

PR #108, opened 2021-04-30
fix: mark obsolete CLI result when cache dropped [ROADRUNNER-120]

ArtsiomCh:
UI presentation fix: Show OSS results as `obsolete` when it's outdated


aldanchenko (approved):
LGTM :)

PR #112, opened 2021-05-14
Feat: add critical severity [ROAD-4]

aldanchenko:
Added support for critical severity.


aldanchenko:
@ArtsiomCh @pavel-snyk I made small changes. For example, look please at io.snyk.plugin.Severity.kt. Waiting your feedback. Thanks!

src/main/kotlin/io/snyk/plugin/cli/Vulnerability.kt
```diff
@@ -42,13 +44,7 @@ class Vulnerability {
 
     // todo: (by Pavel)
     //  How about using comparator for tree nodes? Issue I see atm: this code cannot handle other issues e.g. critical if this will come later.
```
ArtsiomCh, line 46: that todo could probably be removed now as requested abstraction is now implemented in the next line? cc @pavel-snyk

src/main/kotlin/io/snyk/plugin/ui/actions/SnykTreeSeverityFilterActionBase.kt
```diff
@@ -9,12 +9,18 @@ abstract class SnykTreeSeverityFilterActionBase : ToggleAction() {
...
-            (!settings.highSeverityEnabled && settings.mediumSeverityEnabled && !settings.lowSeverityEnabled) ||
-            (!settings.highSeverityEnabled && !settings.mediumSeverityEnabled && settings.lowSeverityEnabled)
+
+        val onlyOneEnabled = arrayOf(
+            settings.criticalSeverityEnabled,
+            settings.highSeverityEnabled,
+            settings.mediumSeverityEnabled,
+            settings.lowSeverityEnabled
+        ).count { it } == 1
+
```
ArtsiomCh, line 19: Cool solution!

ArtsiomCh (approved):
LGTM

[minor] `Severity` class probably could be more efficeantly reperesented as `enum` class, something like:
```
enum class Severity(val text: String) {
  CRITICAL("critical"),
    ...
  ;

  override fun toString() = text
  
  ... // toName and getIndex here
}
```
But doesn't really metter. Current implementatiion is good and expressive enougth.

src/main/kotlin/io/snyk/plugin/Severity.kt
```diff
@@ -0,0 +1,39 @@
...
+            when (severity) {
+                CRITICAL -> 4
+                HIGH -> 3
+                MEDIUM -> 2
+                LOW -> 1
+                else -> 0
+            }
+
+        fun toName(index: Int): String =
+            when (index) {
```
ArtsiomCh, line 23: CRITICAL is foggoten here

src/main/kotlin/io/snyk/plugin/Severity.kt
```diff
@@ -0,0 +1,39 @@
...
+                3 -> HIGH
+                2 -> MEDIUM
+                1 -> LOW
+                else -> LOW
```
ArtsiomCh, line 27: probbaly we should create special type of severity for Unknown index values... might be out of this PR scope

ArtsiomCh (changes requested):

aldanchenko:
> LGTM
> 
> [minor] `Severity` class probably could be more efficeantly reperesented as `enum` class, something like:
> 
> ```
> enum class Severity(val text: String) {
>   CRITICAL("critical"),
>     ...
>   ;
> 
>   override fun toString() = text
>   
>   ... // toName and getIndex here
> }
> ```
> 
> But doesn't really metter. Current implementatiion is good and expressive enougth.

Good idea. I'll take a look :)

src/main/kotlin/io/snyk/plugin/Severity.kt
aldanchenko: @ArtsiomCh I don't added it yet here because as I understand critical severity not implemented for Snyk code for now. And this method used in Snyk code source code part. So just afraid to broke some Snyk code logic. But if you confident that this not broke anything it will be good to add. Thanks! Good catch! 👍

ArtsiomCh: AFAIK there been some rumors about Critical severity implementation in Snyk Code already, but I'm not sure.
Anyway, here it's just an number, so if we never get severity with index 4 from Snyk Code, then we just not show it :)
So adding `4 -> CRITICAL` is safe and should prevent brakeage later when Snyk Code actually will start giving us critical issues.

aldanchenko:
> LGTM
> 
> [minor] `Severity` class probably could be more efficeantly reperesented as `enum` class, something like:
> 
> ```
> enum class Severity(val text: String) {
>   CRITICAL("critical"),
>     ...
>   ;
> 
>   override fun toString() = text
>   
>   ... // toName and getIndex here
> }
> ```
> 
> But doesn't really metter. Current implementatiion is good and expressive enougth.

Tried to convert it to enum. But it seems that the greatest benefit of it can be obtained if we will receive severity as enum from cli. Like CliResult.Vulnerability.Severity <- enum type. Because right now we need always convert enum value to string. As an example in source code it will be like 

`Severity.HIGH.text`

Right now in source code its look like

`Severity.HIGH`

I really liked the idea, but while cli returns severity as string enum will complicate the code :(. 

Maybe we could make Vulnerability.severity as an enum type?

ArtsiomCh:
Hmm... Overrided toSting() method should allow us use just Severity.HIGH whenever String presentation of enum needed. So text val could be private and shouldn't be used outside the enum.
But again, I feel like we do "bike shading" and without enum class everything seems to be clear enough, so we should probably hold on our inner perfectionist here :)

aldanchenko:
> Hmm... Overrided toSting() method should allow us use just Severity.HIGH whenever String presentation of enum needed. So text val could be private and shouldn't be used outside the enum.
> But again, I feel like we do "bike shading" and without enum class everything seems to be clear enough, so we should probably hold on our inner perfectionist here :)

src/main/kotlin/io/snyk/plugin/Severity.kt
ArtsiomCh: Made a quick small fix for that use case: #113 @aldanchenko please have a look

ArtsiomCh (approved):

pavel-snyk (approved):
👍

PR #123, opened 2021-06-09
fix: set maximal attempts for ask for sast feature

pavel-snyk:
This PR fix the issue with infinite loop to ask API for enabling/disabling SAST feature. Maximal attempt count is 20.

### DIFF
--- a/src/main/kotlin/io/snyk/plugin/ui/settings/ScanTypesPanel.kt
+++ b/src/main/kotlin/io/snyk/plugin/ui/settings/ScanTypesPanel.kt
@@ -84,14 +84,17 @@ class ScanTypesPanel(
                         url = getSnykCodeSettingsUrl(),
                         forceShow = true)
 
-                    // check sastEnablement every 1 sec.
+                    // check sastEnablement every 2 sec.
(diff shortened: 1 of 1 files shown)


davidleger95 (approved):

michelkaporin (approved):

ArtsiomCh (changes requested):
That will lead to the same problem we have with Auth: https://snyksec.atlassian.net/browse/ROAD-19?atlOrigin=eyJpIjoiNWViNTUxMmRmMDE0NDVjZjg2Mjk0NzJkZmQzMWRkYTciLCJwIjoiaiJ9
I.e. hardcoding the specific number of attempts (timeframe) will cuase to brake the flow for the "slow" users.
Also don't think we really need that as far as the loop is not infinite, it will be stopped and disposed when `parentDisposable` (parent Settings/Onboarding Panel) is disposed.

ArtsiomCh (approved):

ArtsiomCh:
Accept as workaround for now, but need to be reviewed later.
I see now the problem for which this fix is coming. Would be nice if we have ticket with context created before and then linked to this PR.

PR #125, opened 2021-06-13
fix: network related exceptions and notifications. Related to [ROAD-35]

ArtsiomCh:
- Should fix https://snyk.zendesk.com/agent/tickets/10881
- Explicit notification to the user in case of connection/network errors
- Unification for option/condition checks:
  - ~~Change sastEnable check from check in loop to explicit user's check to avoid servers overload and time outing for slow users~~ loop start only after link to SnykCode option web page opened
  - Change "endpoint support" check from endpoint field listener to explicit user's check to avoid collisions with other checks

### DIFF
--- a/src/main/kotlin/io/snyk/plugin/ui/SnykBalloonNotifications.kt
+++ b/src/main/kotlin/io/snyk/plugin/ui/SnykBalloonNotifications.kt
@@ -12,9 +12,11 @@ import com.intellij.openapi.ui.popup.Balloon
 import com.intellij.openapi.ui.popup.JBPopupFactory
 import com.intellij.ui.LightColors
 import com.intellij.ui.awt.RelativePoint
+import com.intellij.util.Alarm
 import io.snyk.plugin.getApplicationSettingsStateService
 import io.snyk.plugin.getSnykCodeSettingsUrl
 import io.snyk.plugin.settings.SnykProjectSettingsConfigurable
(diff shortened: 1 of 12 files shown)


src/main/kotlin/io/snyk/plugin/ui/SnykBalloonNotifications.kt
```diff
@@ -41,15 +47,33 @@ class SnykBalloonNotifications {
...
+                NotificationAction.createSimpleExpiring("Snyk > Settings > Snyk Code") {
+                    BrowserUtil.browse(getSnykCodeSettingsUrl())
+                    startSastEnablementCheckLoop(project)
+                }
+            )
+            var currentAttempt = 1
+            val maxAttempts = 200
+            lateinit var checkIfSastEnabled: () -> Unit
+            checkIfSastEnabled = {
+                if (getApplicationSettingsStateService().sastOnServerEnabled == true) {
```
pavel-snyk, line 66: @ArtsiomCh , do we execute here an API call?

ArtsiomCh: Just asking settings locally in background, no net call. So presume it's safe.
PS I did check that twice after last week incident :)

pavel-snyk (approved):

PR #129, opened 2021-06-24
fix: welcome and auth screen was missed to display [ROAD-175]

ArtsiomCh:
fix: welcome and auth screen was missed to display if toolWindow was closed when project opening
fix: Run action availability before auth passed


src/main/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanel.kt
```diff
@@ -92,7 +89,7 @@ class SnykToolWindowPanel(val project: Project) : JPanel(), Disposable {
 
         initializeUiComponents()
 
-        displayTreeAndDescriptionPanels()
+        chooseMainPanelToDisplay()
```
pavel-snyk, line 92: good catch and re-use `chooseMainPanelToDisplay()` here and in `checkCliExistsFinished`

pavel-snyk (approved):

PR #131, opened 2021-06-28
fix: ProcessNotCreatedException if try to run CLI command while CLI is still downloading [ROAD-177]

ArtsiomCh:
Also include blocking Auth execution possibility while CLI is still downloading.
More info at https://snyksec.atlassian.net/browse/ROAD-177

PS not to be included into (i.e. blocking) upcoming 2.1.8 release

### DIFF
--- a/src/main/kotlin/io/snyk/plugin/services/SnykCliService.kt
+++ b/src/main/kotlin/io/snyk/plugin/services/SnykCliService.kt
@@ -37,15 +37,12 @@ class SnykCliService(val project: Project) {
 
             val commands = buildCliCommandsList(applicationSettings)
 
-            val projectPath = project.basePath!!
+            val projectPath = project.basePath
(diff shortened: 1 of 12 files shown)


src/main/kotlin/io/snyk/plugin/services/SnykCliService.kt
```diff
@@ -37,15 +37,12 @@ class SnykCliService(val project: Project) {
 
             val commands = buildCliCommandsList(applicationSettings)
 
-            val projectPath = project.basePath!!
+            val projectPath = project.basePath
+                ?: throw IllegalStateException("Scan should not be performed on Default project (with `null` project base dir)")
```
pavel-snyk, line 41: @ArtsiomCh , this message will be propagated to users via notificaitions, right?

pavel-snyk (approved):
LGTM

PR #133, opened 2021-06-30
test: fix `CLI command execution while CLI is downloading` test

ArtsiomCh:
Previous version of test relay on relatively long CLI downloading process (with 100Mb/s channel it takes ~10 sec). 
But while performing tests on GitHub CI, the CLI is downloading (from GitHub) very fast and that should be considered.


pavel-snyk (approved):
good catch!

PR #135, opened 2021-07-06
Feat: show advisor score, phase 1 [ROAD-152]

ArtsiomCh:
phase1 : show scores (for NPM packages) in the Editor + some minor fixes + AdvisorApiClient error-handling + some tests

To do in phase 2 : 
1. Advisor enable/disable checkbox in the Settings and Welcome screen; 
2. hint/popup with "more details" in the Editor


src/main/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanel.kt
```diff
@@ -117,14 +117,15 @@ class SnykToolWindowPanel(val project: Project) : JPanel(), Disposable {
                 override fun scanningCliError(cliError: CliError) {
                     currentCliResults = null
                     ApplicationManager.getApplication().invokeLater {
+                        SnykBalloonNotifications.showError(cliError.message, project)
                         if (cliError.message.startsWith("Authentication failed. Please check the API token on ")) {
+                            getApplicationSettingsStateService().token = null
```
pavel-snyk, line 122: Why do we want to remove existing token? Sure, it's invalid because of auth error, but e.g. endpoint can be changed.

src/main/kotlin/snyk/advisor/SnykAdvisorModel.kt
```diff
@@ -0,0 +1,98 @@
...
+
+    @TestOnly
+    fun getPackagesRequested() = packagesRequested
+
+    override fun dispose() {}
+
+    companion object {
+        private const val H24_IN_MILLISECONDS = 24 * 60 * 60 * 1000
+        const val IGNORE_REQUESTS_DELAY: Long = 1000 //ms
+        const val SCORE_REQUESTS_BATCHING_DELAY = 1000 //ms
```
pavel-snyk, line 96: How about adding a suffix to show that it's milliseconds? It will be simpler to read.

src/main/kotlin/snyk/advisor/api/AdvisorApiClient.kt
```diff
@@ -13,7 +16,27 @@ class AdvisorApiClient private constructor(
     private lateinit var retrofit: Retrofit
     private lateinit var scoreServiceEndpoint: ScoreService
 
-    fun scoreService(): ScoreService {
+    fun getPackagesInfo(packageManager: AdvisorPackageManager, packageNames: List<String>): List<PackageInfo>? {
+        try {
+            log.debug("Executing request to Advisor api")
```
pavel-snyk, line 21: Do we need this line? It will not very helpful by debugging or issue investigation. `HttpLoggingInterceptor` will log requests and response.

src/test/kotlin/snyk/advisor/api/AdvisorApiClientTest.kt
```diff
@@ -163,10 +165,43 @@ class AdvisorApiClientTest {
...
+        assertThat(actualInfos, notNullValue())
+        assertThat(actualInfos, hasSize(expectedInfos.size))
+
+        actualInfos?.forEachIndexed { index, actual ->
+            val expected = expectedInfos[index]
+            assertThat(actual.name, equalTo(expected.name))
+            assertThat(actual.score, equalTo(expected.score))
+            assertThat(actual.pending, equalTo(expected.pending))
+            assertThat(actual.error, equalTo(expected.error))
+            assertThat(actual.labels, equalTo(expected.labels))
```
pavel-snyk, line 198: assert DTO instead of each field? if unit test will fail you will see more meaningful messages

src/main/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanel.kt
ArtsiomCh: We have checks like `token != null` in various places (for example SnykRunScanAction#update, AdvisorClient creation. etc) to enable/allow some action to perform. And if we keep invalid token then many actions wil produce errors or have undefined behaviour.

src/main/kotlin/snyk/advisor/api/AdvisorApiClient.kt
ArtsiomCh: HttpLoggingInterceptor is not used yet. Also even if it will, there will be only warn events. We might consider to review such logging after start usage of HttpLoggingInterceptor.

pavel-snyk (approved):
Good job, smart idea with Model :)!

PR #141, opened 2021-07-20
chore: move getting plugin info into utility class

pavel-snyk:
This PR extract getting plugin information logic into `PluginInformation.kt` class.
We'll need this information by setting:
- envvars
- user-agent header
- sentry alerting (later)


aldanchenko (approved):
LGTM 👍

PR #143, opened 2021-07-23
feat: use amplitude experiment api [ROAD-122]

pavel-snyk:
This PR adds a Retrofit client to use Amplitude Experiment API.

Amplitude Experiment REST API: https://developers.experiment.amplitude.com/reference. Note, that this documentation is outdated and correct methods and endpoints can be found in [experiment-android-client](https://github.com/amplitude/experiment-android-client/blob/main/sdk/src/main/java/com/amplitude/experiment/DefaultExperimentClient.kt#L122-L129)

All classes mimic Amplitude Android SDK names (except Android runtime dependencies) to simplify migration later when JVM Kotlin SDK for experiments will be released.


src/main/kotlin/snyk/amplitude/api/AmplitudeExperimentApiClient.kt
```diff
@@ -0,0 +1,81 @@
...
+            if (!response.isSuccessful) {
+                log.info("Error response: $response")
+                return emptyMap()
+            }
+
+            val variants = response.body()
+            log.debug("Received variants: $variants")
+            variants ?: emptyMap()
+        } catch (e: IOException) {
+            log.error("Could not fetch variants because of network communication error with amplitude server", e)
```
ArtsiomCh, line 39: log.error will be shown to the user as exception

src/main/kotlin/snyk/amplitude/api/AmplitudeExperimentApiClient.kt
```diff
@@ -0,0 +1,81 @@
...
+import okhttp3.OkHttpClient
+import retrofit2.Retrofit
+import retrofit2.converter.gson.GsonConverterFactory
+import snyk.net.HttpClient
+import java.io.IOException
+
+class AmplitudeExperimentApiClient private constructor(
+    private val baseUrl: String,
+    private val httpClient: OkHttpClient
+) {
```
ArtsiomCh, line 13: This is our 3rd ApiClient already :( May be we need abstract/generic ApiClient class to inherite from?

src/main/kotlin/snyk/amplitude/api/ExperimentUser.kt
```diff
@@ -0,0 +1,15 @@
+package snyk.amplitude.api
+
+import com.google.gson.annotations.SerializedName
+
+class ExperimentUser(
```
ArtsiomCh, line 5: data class?

src/main/kotlin/snyk/amplitude/AmplitudeExperimentService.kt
```diff
@@ -0,0 +1,44 @@
...
+import snyk.amplitude.api.ExperimentUser
+import snyk.amplitude.api.Variant
+import java.util.concurrent.ConcurrentHashMap
+
+@Service
+class AmplitudeExperimentService : Disposable {
+    private val log = logger<AmplitudeExperimentService>()
+
+    private val apiClient: AmplitudeExperimentApiClient =
+        AmplitudeExperimentApiClient.create(apiKey = "client-bQAwhmzLhgpgqAhKAqRzIjnS915YrAqj")
```
ArtsiomCh, line 17: hardcoded key?

src/main/kotlin/snyk/amplitude/api/ExperimentUser.kt
pavel-snyk: The purpose of this class is not to hold data only, but handle changing userId (atm not implemented).

src/main/kotlin/snyk/amplitude/AmplitudeExperimentService.kt
pavel-snyk: The envkey is hardcoded to simplify development, maybe later we'll make it configurable. Compare https://github.com/snyk/vscode-extension/blob/feat/experiment/src/snyk/experiments/amplitude.ts#L11

src/main/kotlin/snyk/amplitude/api/AmplitudeExperimentApiClient.kt
pavel-snyk: Let's sync offline about it.

src/main/kotlin/snyk/amplitude/api/AmplitudeExperimentApiClient.kt
pavel-snyk: good catch!

src/main/kotlin/snyk/amplitude/api/AmplitudeExperimentApiClient.kt
ArtsiomCh: Will be replaced with Experiment Kotlin SDK, so ok for now

ArtsiomCh (approved):
LGTM

PR #152, opened 2021-08-03
feat: update health score event when user is clicking the Advisor tooltip [ROAD-205]

pavel-snyk:
This PR adds tracking of `Heath Score is Clicked` iteratively event by clicking on advisor detail popup.

_Note for reviewers:_
Adding tests here would be tricky since mouse event listener is inner class and mockito cannot access it, let's think about UI tests and including it later.


src/main/kotlin/snyk/advisor/AdvisorScoreProvider.kt
```diff
@@ -178,6 +190,13 @@ class AdvisorScoreProvider(
                     """.trimIndent(),
                     e.mouseEvent
                 )
+                service<SnykAnalyticsService>().logHealthScoreIsClicked(
+                    HealthScoreIsClicked.builder()
+                        .ecosystem(packageManager.getEcosystem())
+                        .ide(HealthScoreIsClicked.Ide.JETBRAINS)
+                        .packageName(info.name)
+                        .build()
+                )
```
ArtsiomCh, line 199: Move up before `val labels` initialization? 
We can fall on `val labels` initialization or later on `SnykBalloonNotifications.showAdvisorMoreDetailsPopup` invocation and our `click` event should be registered before those possible falls.

pavel-snyk: If we would fall on `SnykBalloonNotifications.showAdvisorMoreDetailsPopup`, should we "count" it as click. A user will not see the popup, and we'll register this as successful click 🤔 .

ArtsiomCh (approved):
LGTM
Just clarification with `clicked` vs `viewed` event need to addressed.

src/main/kotlin/snyk/advisor/AdvisorScoreProvider.kt
pavel-snyk: clarified in JIRA ticket => resolved.

PR #160, opened 2021-08-19
fix: NullPointerException when no release info available [ROAD-187]

michelkaporin:
Screenshot 2021-08-19 at 16 42 21

### DIFF
--- a/src/main/kotlin/io/snyk/plugin/services/SnykCliDownloaderService.kt
+++ b/src/main/kotlin/io/snyk/plugin/services/SnykCliDownloaderService.kt
@@ -2,20 +2,24 @@ package io.snyk.plugin.services
 
 import com.google.gson.Gson
 import com.google.gson.annotations.SerializedName
+import com.intellij.ide.BrowserUtil
+import com.intellij.notification.NotificationAction
 import com.intellij.openapi.application.ApplicationManager
 import com.intellij.openapi.components.Service
 import com.intellij.openapi.progress.ProgressIndicator
+import com.intellij.openapi.project.Project
 import com.intellij.util.io.HttpRequests
 import io.snyk.plugin.cli.Platform
 import io.snyk.plugin.events.SnykCliDownloadListener
 import io.snyk.plugin.getApplicationSettingsStateService
 import io.snyk.plugin.getCliFile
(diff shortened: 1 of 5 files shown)


src/main/kotlin/io/snyk/plugin/ui/SnykBalloonNotifications.kt
```diff
@@ -28,6 +28,7 @@ import io.snyk.plugin.services.SnykTaskQueueService
 import io.snyk.plugin.settings.SnykProjectSettingsConfigurable
 import io.snyk.plugin.snykToolWindow
 import io.snyk.plugin.startSastEnablementCheckLoop
+import org.jetbrains.annotations.Nullable
```
pavel-snyk, line 31: @michelkaporin, can you remove the unused import? (detekt-warning)

michelkaporin: @pavel-snyk detekt does its job, removed :)

ArtsiomCh:
FYI notification.notify() with Null project is allowed and suppose to show balloon in all opened projects, but been broken few times already and imho could not be relayed on. (See cooresponent bugs on youtrack)

pavel-snyk (approved):

src/main/kotlin/io/snyk/plugin/services/SnykCliDownloaderService.kt
```diff
@@ -118,8 +141,11 @@ class SnykCliDownloaderService {
     }
 
     fun isNewVersionAvailable(currentCliVersion: String?, newCliVersion: String?): Boolean {
-        if (currentCliVersion == null || newCliVersion == null
-            || currentCliVersion.isEmpty() || currentCliVersion.isEmpty()) {
+        if (currentCliVersion == null ||
```
michelkaporin, line 144: Let's fix this detekt warning ("Complex conditions should be simplified and extracted into well-named methods if necessary")

michelkaporin (commented):
@pavel-snyk I see you have approved this PR, however the mentioned bug with the ToolWindow displaying "Downloading CLI..." after "Retry" is probably still there, unless I don't see how you've fixed it?

PR #165, opened 2021-08-31
feat(ui): enhance tree with container feature [ROAD-256]

pavel-snyk:


src/main/kotlin/io/snyk/plugin/services/SnykTaskQueueService.kt
```diff
@@ -166,8 +179,41 @@ class SnykTaskQueueService(val project: Project) {
...
+        })
+    }
+
+    private fun scheduleContainerScan() {
+        taskQueueIacContainer.run(object : Task.Backgroundable(project, "Snyk Container is scanning", true) {
+            override fun run(indicator: ProgressIndicator) {
+                val toolWindowPanel = project.service<SnykToolWindowPanel>()
+                if (toolWindowPanel.currentContainerResult != null) return
+
+                val iacResult = toolWindowPanel.currentIacResult
```
aldanchenko, line 193: It seems we use ToolWindowPanel also in other parts of this service. My doubts that ToolWindowPanel is ui part and we are tied hard background service with ui here. Maybe it will be better to save results in other place. But it's just my thoughts about this. Added here to not forget :)

src/main/kotlin/snyk/container/ContainerService.kt
```diff
@@ -2,18 +2,100 @@ package snyk.container
...
+
+                if (lineWithImage.first != null) {
+                    // extract image name
+                    val imageName = lineWithImage.second.trim().replace("image:", "").trim()
+                    LOG.info("Found image tag for container scan: $imageName")
+
+                    var commands = buildCliCommandsList()
+                    val mutCommands = commands.toMutableList()
+                    mutCommands.add(imageName)
+                    commands = mutCommands.toList()
```
aldanchenko, line 54: [nitpick]: Maybe we could simplify this like `commands = listOf(imageName, *commands.toTypedArray())`

aldanchenko (approved):
LGTM 👍 Great work @pavel-snyk!

src/main/kotlin/snyk/container/ContainerService.kt
pavel-snyk: good catch :)

src/main/kotlin/io/snyk/plugin/services/SnykTaskQueueService.kt
pavel-snyk: yes, you're right. At the moment this is *the way* how results are shared, but we have to refactor/modify the whole tree.

PR #166, opened 2021-09-03
[ROAD-243] Feat: add run plugin verifier task.

aldanchenko:
Add runPluginVerifier Gradle task and add Github action for it.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -51,6 +51,8 @@ jobs:
       - name: Fetch Sources
         uses: actions/checkout@v2
 
+      # PluginVerifier has been compiled by a more recent version of the Java - Java 11.
+      # That's why, we are using not version 8 here, but 11.
       - name: Setup Java
         uses: actions/setup-java@v1
         with:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -42,3 +42,28 @@ jobs:
 
       - name: Run Tests
         run: ./gradlew clean check -x detekt
+
+  pluginVerifier:
+    name: Plugin Verifier
+    needs: gradleValidation
+    runs-on: ubuntu-18.04
+    steps:
+      - name: Fetch Sources
+        uses: actions/checkout@v2
+
+      - name: Setup Java
+        uses: actions/setup-java@v1
+        with:
+          java-version: 11
+
+      - name: Setup Gradle Cache
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.gradle/caches
+            ~/.gradle/wrapper
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle') }}
+          restore-keys: ${{ runner.os }}-gradle
+
+      - name: Run Plugin Verifier
+        run: ./gradlew runPluginVerifier
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -136,6 +136,17 @@ tasks {
       ideDir.set(File(properties("localIdeDirectory")))
     }
   }
+
+  runPluginVerifier {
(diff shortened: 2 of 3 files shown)


aldanchenko:
@pavel-snyk I added Github action, but have some questions around it. 

For example, `failureLevel`. You could check available settings [here](https://github.com/JetBrains/gradle-intellij-plugin/blob/ea5eb75548af4b368f2bd981d8c2d338edb3208d/src/main/groovy/org/jetbrains/intellij/tasks/RunPluginVerifierTask.groovy#L29).

And right now reports saved to `$rootDir.path/reports/pluginVerifier`. I tried to send them via email, but haven’t yet been able to solve some tricky parts with path.

.github/workflows/build.yml
```diff
@@ -40,5 +40,8 @@ jobs:
       - name: Verify Plugin
         run: ./gradlew verifyPlugin
 
+      - name: Run Plugin Verifier
+        run: ./gradlew runPluginVerifier
+
```
pavel-snyk, line 45: I agree that this task should be the part of "build" workflow. Let's move it into separate job (atm we have `gradleValidation` and `test`) so we can control PR checks and see it separately.

aldanchenko: @pavel-snyk could you please review last changes?

.github/workflows/build.yml
```diff
@@ -42,3 +42,31 @@ jobs:
...
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.gradle/caches
+            ~/.gradle/wrapper
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle') }}
+          restore-keys: ${{ runner.os }}-gradle
+
+      - name: Verify Plugin
+        run: ./gradlew verifyPlugin
```
pavel-snyk, line 69: @aldanchenko , we can omit this lines since we already run `verifyPlugin` command in **test** job

pavel-snyk (approved):

.github/workflows/build.yml
aldanchenko: Good point @pavel-snyk. Thanks!

.github/workflows/build.yml
```diff
@@ -42,3 +42,28 @@ jobs:
...
+    needs: gradleValidation
+    runs-on: ubuntu-18.04
+    steps:
+      - name: Fetch Sources
+        uses: actions/checkout@v2
+
+      - name: Setup Java
+        uses: actions/setup-java@v1
+        with:
+          java-version: 11
```
michelkaporin, line 57: Why is java-version 11 here, whereas we test with 1.8 version in the test job? Can we leave an inline comment perhaps or is it me not knowing the verification specifics coming from a different tech stack?

aldanchenko:
@pavel-snyk I squashed commits, Could you please check the result?

.github/workflows/build.yml
aldanchenko: Good question @michelkaporin!. We need Java 11 because of this error: Error: 

`Exception in thread "main" java.lang.UnsupportedClassVersionError: **_com/jetbrains/pluginverifier/PluginVerifierMain has been compiled by a more recent version of the Java Runtime_** (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0`.

michelkaporin: I see, can you perhaps leave a comment with explanation for this? Will be valuable in future, when we forget 👍

aldanchenko: @michelkaporin added comment for this ;). Thanks! 👍

michelkaporin (approved):

PR #167, opened 2021-09-13
feat: implement Sentry error reporter as error submitter [ROAD-261]

pavel-snyk:
This PR implements the extension point `com.intellij.errorHandler` to integrate the plugin with the IDE's dialog to report fatal errors.

### Additional info
<details>
  <summary>Click to see example screenshots</summary>
  image
</details>

### DIFF
--- a/src/main/kotlin/snyk/errorHandler/SentryErrorReporter.kt
+++ b/src/main/kotlin/snyk/errorHandler/SentryErrorReporter.kt
@@ -106,8 +106,11 @@ object SentryErrorReporter {
     fun captureException(throwable: Throwable): SentryId {
         val settings = getApplicationSettingsStateService()
         return if (settings.crashReportingEnabled) {
-            Sentry.captureException(throwable)
+            val sentryId = Sentry.captureException(throwable)
+            LOG.info("Sentry event reported: $sentryId")
+            sentryId
         } else {
+            LOG.info("Sentry crash reporting is disabled")
             SentryId.EMPTY_ID
         }
     }
--- a/src/main/kotlin/snyk/errorHandler/SentryErrorReporter.kt
+++ b/src/main/kotlin/snyk/errorHandler/SentryErrorReporter.kt
@@ -15,6 +15,7 @@ import io.sentry.protocol.Message
 import io.sentry.protocol.OperatingSystem
 import io.sentry.protocol.SentryId
 import io.sentry.protocol.SentryRuntime
+import io.snyk.plugin.getApplicationSettingsStateService
 import snyk.PropertyLoader
 import snyk.pluginInfo
 
@@ -94,6 +95,20 @@ object SentryErrorReporter {
                 LOG.warn("Unable to submit Sentry error information")
                 consumer.consume(SubmittedReportInfo(SubmissionStatus.FAILED))
             }
+        }
+    }
+
+    /**
+     * Captures exception with Sentry. This method should be used in `try-catch` blocks.
+     *
+     * Note: [io.snyk.plugin.services.SnykApplicationSettingsStateService.crashReportingEnabled] is respected
+     */
+    fun captureException(throwable: Throwable): SentryId {
+        val settings = getApplicationSettingsStateService()
+        return if (settings.crashReportingEnabled) {
+            Sentry.captureException(throwable)
+        } else {
+            SentryId.EMPTY_ID
         }
     }
 }
--- a/src/main/kotlin/snyk/errorHandler/SentryErrorReporter.kt
+++ b/src/main/kotlin/snyk/errorHandler/SentryErrorReporter.kt
@@ -1,0 +1,99 @@
+package snyk.errorHandler
+
+import com.intellij.openapi.diagnostic.Attachment
+import com.intellij.openapi.diagnostic.SubmittedReportInfo
+import com.intellij.openapi.diagnostic.SubmittedReportInfo.SubmissionStatus
+import com.intellij.openapi.diagnostic.logger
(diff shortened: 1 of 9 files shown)


src/main/resources/application.properties
```diff
@@ -1,3 +1,5 @@
 amplitude.experiment.api-key=@amplitude.experiment.api-key@
+environment=@environment@
 iteratively.analytics.environment=@iteratively.analytics.environment@
 segment.analytics.write-key=@segment.analytics.write-key@
+sentry.dsn=https://33e9484001e8411693945d7d1a8a2f64@o30291.ingest.sentry.io/5956531
```
michelkaporin, line 5: Should DSN key be exposed like that? Does it also mean everyone knowing the public key can submit reports to our project in Sentry?

ArtsiomCh: We could "hide" it under github secret in insert during final gradle build task, but it still be visible in final binary and could be extracted form there.
I think we have to provide some kind of public endpoint to get reports. Otherwise we probably also should hide all our company emails from public appearance to avoid be spammed.

michelkaporin: @ArtsiomCh analytics write key will be in the final binary too, however we do hide it :)

pavel-snyk: @ArtsiomCh , @michelkaporin , Sentry DSN is not a secret in comparison of  Segment write key, take a look on other examples:

- https://github.com/asciidoctor/asciidoctor-intellij-plugin/blob/main/src/main/java/org/asciidoc/intellij/errorHandler/SentryErrorReporter.java#L48
- https://github.com/intellij-solidity/intellij-solidity/blob/4a9669bb22f509db54ee77ea8456d1effd2db618/src/main/kotlin/me/serce/solidity/ide/errors.kt#L27

Since we don't have different projects to send crash reports, there is no reason to make it too complicated.

michelkaporin:
@pavel-snyk does this PR satisfy all acceptance criteria from the ticket?
`Handled errors from try/catch blocks. We should invest in this way of handling errors and report errors from all catch blocks.  ` seems not to be addressed here, unless I understand it wrongly.

ArtsiomCh:
> @pavel-snyk does this PR satisfy all acceptance criteria from the ticket?
> `Handled errors from try/catch blocks. We should invest in this way of handling errors and report errors from all catch blocks. ` seems not to be addressed here, unless I understand it wrongly.

@michelkaporin captureException() is responsible for that. We probably need to go over all try-catch blocks in our codebase and replace/add log.warn() with that fun. 
@pavel-snyk  Shell we do that within this PR or a separate PR/tickets needed?

michelkaporin:
@pavel-snyk yeap. Current catch blocks won't be captured via IntelliJ error dialog either. Agree that it's not part of the PR but it's within this ticket scope, probably worth amending the scope and communicating the changes to the team, or creating new ticket then.

src/main/kotlin/snyk/errorHandler/SnykErrorReportSubmitter.kt
```diff
@@ -0,0 +1,45 @@
...
+
+            if (event.data is AbstractMessage) {
+                throwable = (event.data as AbstractMessage).throwable
+                attachments = (event.data as AbstractMessage).includedAttachments
+            }
+            if (throwable is PluginException && throwable.cause != null) {
+                // unwrap PluginManagerCore.createPluginException
+                throwable = throwable.cause
+            }
+            SentryErrorReporter.submitErrorReport(
```
michelkaporin, line 36: This function always returns true. Should we return false when `SentryErrorReporter.submitErrorReport` fails? What's the behaviour when this.submit returns `false`?

ArtsiomCh: Just checked behavior when this.submit returns `false` and surprisingly nothing happened. I expect Idea at least notify user that submission was unsuccessful...
But anyway we probably should return from SentryErrorReporter.submitErrorReport if we actually been able to submit. And I see we already do log such event.

pavel-snyk: Other plugins return 'true' always :), but let's not relay on others.

If submit returns `false` then your error for IntelliJ is still "active" and will/can be shown to users. I suggest that we show a notification whether submitting Sentry report was successful or not. Wdyt?

michelkaporin: I guess they return `true` for a better UX. If user gets a notification that it wasn't successful, what can he do with it? Probably nothing, additionally being annoyed that he cannot even submit an error report 😄 Maybe leave it as-is for the time being then :)

ArtsiomCh (approved):
LGTM
PS can't wait to have in prod, we've been waiting for it for a while....

src/main/kotlin/snyk/errorHandler/SentryErrorReporter.kt
```diff
@@ -0,0 +1,117 @@
...
+                runtime.name = pluginInfo.integrationEnvironment
+                runtime.version = pluginInfo.integrationEnvironmentVersion
+                event.contexts.setRuntime(runtime)
+
+                event.setTag("java.vendor", SystemInfo.JAVA_VENDOR)
+                event.setTag("java.version", SystemInfo.JAVA_VERSION)
+                event.setTag("java.runtime.version", SystemInfo.JAVA_RUNTIME_VERSION)
+
+                // clear the server name
+                event.serverName = null
```
michelkaporin, line 52: @pavel-snyk why did you need this?

PR #169, opened 2021-09-15
fix: prevent multiple execution of Iteratively load method [ROAD-296]

pavel-snyk:
This PR extracts calling `Itly.getInstance().load` method into singleton object `Iteratively`.

The fix is needed because sometimes IntelliJ creates light services from different classloaders.


ArtsiomCh (approved):
LGTM
The repetitive 
```
if (!settings.usageAnalyticsEnabled || userId.isBlank()) {
            return
        }
```
could/should be addressed later.

PR #170, opened 2021-09-16
fix(experiment): add setting to show scanner reminder only once

pavel-snyk:
This PR adds a settings to show scanner reminder only once (since it's not possible to set it from Amplitude side).


ArtsiomCh (approved):
LGTM

PR #171, opened 2021-09-19
fix: for collecting all project files, our own recursive dirs scan -> Idea api (visitor) [ROAD-269]

ArtsiomCh:


src/main/kotlin/io/snyk/plugin/snykcode/core/SnykCodeUtils.kt
```diff
@@ -21,27 +21,29 @@ class SnykCodeUtils private constructor() : DeepCodeUtilsBase(
...
         val project = PDU.toProject(projectO)
-        return RunUtils.computeInReadActionInSmartMode(
+        val progressIndicator = ProgressManager.getInstance().progressIndicator ?: null
+        val allVirtualFiles = mutableSetOf<VirtualFile>()
+
+        val cancelled = !ProjectRootManager.getInstance(project).fileIndex.iterateContent {
+            if (!it.isDirectory) {
+                allVirtualFiles.add(it)
+            }
+            return@iterateContent progressIndicator?.isCanceled != true
```
michelkaporin, line 33: should we have another log entry here, if we return before :34 ?

src/integTest/kotlin/io/snyk/plugin/snykcode/core/IgnoreInfoHolderPlatformTestCase.kt
```diff
@@ -14,6 +15,11 @@ import java.io.FileNotFoundException
 // See: https://plugins.jetbrains.com/docs/intellij/light-and-heavy-tests.html
 class IgnoreInfoHolderPlatformTestCase : HeavyPlatformTestCase() {
 
+    override fun setUp() {
+        super.setUp()
+        ModuleRootModificationUtil.addContentRoot(module, project.basePath!!)
```
pavel-snyk, line 20: @ArtsiomCh, why we need this change?

src/main/kotlin/io/snyk/plugin/snykcode/core/SnykCodeUtils.kt
```diff
@@ -21,27 +21,29 @@ class SnykCodeUtils private constructor() : DeepCodeUtilsBase(
     fun allProjectFilesCount(project: Project): Int = allProjectFiles(project).size
 
     override fun allProjectFiles(projectO: Any): Collection<Any> {
+        scLogger.logInfo("allProjectFiles requested")
         val project = PDU.toProject(projectO)
-        return RunUtils.computeInReadActionInSmartMode(
+        val progressIndicator = ProgressManager.getInstance().progressIndicator ?: null
+        val allVirtualFiles = mutableSetOf<VirtualFile>()
+
+        val cancelled = !ProjectRootManager.getInstance(project).fileIndex.iterateContent {
```
pavel-snyk, line 29: @ArtsiomCh , do we iterate over all project files (included, excluded, ignored etc)? Could  you elaborate the difference between previous and current method to get files?

pavel-snyk (commented):
left some comments

src/integTest/kotlin/io/snyk/plugin/snykcode/core/IgnoreInfoHolderPlatformTestCase.kt
ArtsiomCh: @pavel-snyk HeavyPlatformTestCase does not create ContentRoot for default project. And now we use IntelliJ api to iterate over project files that internally relay on ContentRoot setting.

src/main/kotlin/io/snyk/plugin/snykcode/core/SnykCodeUtils.kt
ArtsiomCh: @michelkaporin Here we not return from outer fun but inner iterator ContentIterator.processFile() fun: "Returns: false if files processing should be stopped, true if it should be continued" I.e we just stop iterator in case user cancel task before it actually finished.
We'll pass through logging at :38 in any case.

michelkaporin: @ArtsiomCh got it, thanks!

src/main/kotlin/io/snyk/plugin/snykcode/core/SnykCodeUtils.kt
pavel-snyk: I remember we talked about it.
This is my point: we have now 3 levels where we filter out source files:
- .gitignore
- .dciignore
- IntelliJ excluded/ignored (from gitignore) files/folders.

I guess it's ok and we should use IntelliJ API.

src/integTest/kotlin/io/snyk/plugin/snykcode/core/IgnoreInfoHolderPlatformTestCase.kt
pavel-snyk: got it, thanks for clarification.

pavel-snyk (approved):
👍

PR #174, opened 2021-09-21
fix: cache invalidation for OssResults happened for other projects build file change [ROAD-291]

ArtsiomCh:


src/integTest/kotlin/io/snyk/SnykBulkFileListenerTest.kt
```diff
@@ -0,0 +1,56 @@
...
+import com.intellij.testFramework.PsiTestUtil
+import com.intellij.testFramework.fixtures.BasePlatformTestCase
+import io.snyk.plugin.ui.toolwindow.SnykToolWindowPanel
+import org.junit.Test
+import snyk.oss.OssResult
+
+class SnykBulkFileListenerTest : BasePlatformTestCase() {
+
+    @Test
+    fun testBuildFileChanged() {
```
pavel-snyk, line 17: neat pick: I recommend to use following schema for test methods: `test<methodUnderTest>_<testedStated>_<expectedBehaviour>`.
Since IntelliJ platform uses JUnit3 as test framework `test*` prefix for method is necessary.

For this case it could be something like that: `testCurrentOssResults_shouldDropCachedResult_whenBuildFileChanged`

pavel-snyk (approved):
LGTM.

src/integTest/kotlin/io/snyk/SnykBulkFileListenerTest.kt
ArtsiomCh: Like the schema, despite Idea's inspection is complaining about naming, it make tests more "speaking".

PR #178, opened 2021-09-28
fix: set production Sentry environment when running release workflow

pavel-snyk:
This PR sets Sentry environment ("PRODUCTION") when running `release` GH workflow.
Otherwise it will be "DEVELOPMENT".


aldanchenko (approved):
LGTM :) Thanks Pavel 👍🚀

PR #182, opened 2021-10-04
fix: request Idea restart after plugin update [ROAD-390]

ArtsiomCh:
due to some hard to catch issues with dynamic services unload/load


pavel-snyk (approved):
small PR -> good PR :).

PR #184, opened 2021-10-05
feat: add missing iteratively events [ROAD-267]

pavel-snyk:


ArtsiomCh:
Shell we address the case when user install/uninstall plugin without opening the project?

PR #185, opened 2021-10-05
feat: timeout for Snyk Code scan `waiting for results` phase increase [ROAD-389]

ArtsiomCh:
To 10 minutes (to match OSS scan timeout). Could be changed through 'snyk.timeout.results.waiting' Idea registry key.

Could be reviewed and merged only after https://github.com/snyk/code-sdk-java/pull/25 merge and snyk-code-client:2.1.11 published in Maven


src/main/resources/META-INF/plugin.xml
```diff
@@ -23,6 +23,13 @@
     <editorFactoryListener implementation="snyk.advisor.AdvisorEditorFactoryListener"/>
 
     <errorHandler implementation="snyk.errorHandler.SnykErrorReportSubmitter"/>
+
+    <registryKey
+      key="snyk.timeout.results.waiting"
+      defaultValue="720000"
+      description="Snyk timeout (milliseconds) to wait for results during scan"
```
pavel-snyk, line 30: @ArtsiomCh, let set `restartRequired=true` parameter too.

src/main/kotlin/io/snyk/plugin/cli/ConsoleCommandRunner.kt
```diff
@@ -63,7 +64,16 @@ open class ConsoleCommandRunner {
...
-        val processOutput = ScriptRunnerUtil.getProcessOutput(processHandler, ScriptRunnerUtil.STDOUT_OUTPUT_KEY_FILTER, 720000)
+        val timeout = getWaitForResultsTimeout()
+        val processOutput = try {
+            ScriptRunnerUtil.getProcessOutput(
+                processHandler,
+                ScriptRunnerUtil.STDOUT_OUTPUT_KEY_FILTER,
+                timeout
+            )
+        } catch (e: ExecutionException) {
+            "Execution timeout [${timeout / 1000} sec] is reached with NO results produced"
```
pavel-snyk, line 75: We should not swallow the exception here. How about using `SentryErrorReporter.captureException(e)`?

ArtsiomCh: Why we should not? We intentionally swallow ExecutionException here to show timeout error notification instead of fail with exception.
PS will fix: https://sentry.io/organizations/snyk/issues/2695118718/?project=5956531&referrer=slack

src/main/resources/META-INF/plugin.xml
ArtsiomCh: why do we need restart here? we request that value dynamically through getWaitForResultsTimeout()
So no restart needed. Am I missing something?

src/main/kotlin/io/snyk/plugin/cli/ConsoleCommandRunner.kt
ArtsiomCh: Actually have sense to log it in Sentry but not pass it to the user.

src/main/resources/META-INF/plugin.xml
pavel-snyk: right, our settings-bug put me into "always-restart" mode :).

src/main/kotlin/io/snyk/plugin/cli/ConsoleCommandRunner.kt
pavel-snyk: > Why we should not?
Exception swallowing is an anti-pattern.
We wanna to address issues in case we'll have an increased rate of such exceptions + we will have a full stacktrace.

PR #203, opened 2021-10-21
feat: ✨Add ability to ignore IaC issue (IaC) [ROAD-383]

bastiandoetsch:
As a developer I'd like to be able to ignore the IaC issues, so that I only see issues that are important to me.
Implementation details:
- Uses snyk ignore id=<issueId> and the .snyk file to ignore
- After clicking Ignore This Issue, the button is diabled and the text changes to Issue now ignored
- The tree node of the issue is NOT removed at this point, but would disappear in the next scan
- The button does NOT stay disabled when switching between TreeNodes
- It works. Additional details for user workflows would need to be added as separate stories

### DIFF
--- a/src/main/kotlin/io/snyk/plugin/services/CliAdapter.kt
+++ b/src/main/kotlin/io/snyk/plugin/services/CliAdapter.kt
@@ -12,7 +12,7 @@ import org.jetbrains.annotations.TestOnly
 /**
  * Wrap work with Snyk CLI.
  */
-abstract class CliAdapter<R>(val project: Project, private val cliCommands: List<String>) {
+abstract class CliAdapter<R>(val project: Project) {
 
     private var consoleCommandRunner = ConsoleCommandRunner()
 
@@ -26,10 +26,10 @@ abstract class CliAdapter<R>(val project: Project, private val cliCommands: List
         return getCliFile().exists()
     }
 
-    fun scan(): R = try {
-        val commands = buildCliCommandsList()
+    fun execute(commands: List<String>): R = try {
+        val cmds = buildCliCommandsList(commands)
         val apiToken = pluginSettings().token ?: ""
-        val rawResultStr = consoleCommandRunner.execute(commands, projectPath, apiToken, project)
+        val rawResultStr = consoleCommandRunner.execute(cmds, projectPath, apiToken, project)
         convertRawCliStringToCliResult(rawResultStr)
     } catch (exception: CliNotExistsException) {
         getErrorResult(exception.message ?: "Snyk CLI not installed.")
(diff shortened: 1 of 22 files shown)


src/main/kotlin/snyk/iac/IacIssue.kt
```diff
@@ -0,0 +1,15 @@
...
+    val severity: String = "",
+    val publicId: String = id,
+    val documentation: String = "",
+    val lineNumber: Int = 0,
+    val issue: String = id,
+    val impact: String = "",
+    val resolve: String? = null,
+    val references: List<String> = emptyList(),
+    val path: List<String> = emptyList()
+)
```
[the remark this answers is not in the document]
bastiandoetsch: wasn't sure about the lists - should I remove them, too?

bastiandoetsch:
@pavel-snyk @ArtsiomCh could one of you have another look?

src/main/kotlin/io/snyk/plugin/services/download/CliDownloader.kt
```diff
@@ -22,32 +22,31 @@ class CliDownloader {
...
         HttpRequests
             .request(url)
             .productNameAsUserAgent()
             .saveToFile(cliFile, indicator)
 
-        verifyCLIChecksum(cliFile)
+        verifyChecksum(expectedSha, cliFile.readBytes())
         cliFile.setExecutable(true)
         return cliFile
     }
```
ArtsiomCh, line 52: don't we have all changes above in another PR? Might it conflict?

src/main/kotlin/io/snyk/plugin/services/CliAdapter.kt
```diff
@@ -84,7 +82,7 @@ abstract class CliService<R>(val project: Project, private val cliCommands: List
 
         logger.debug("Cli parameters: $commands")
 
-        return commands.toList()
+        return commands.toMutableList()
```
ArtsiomCh, line 85: toList() should be fine here

ArtsiomCh:
Generally looks good. Few fundamental issue need to be resolved before that feature GA:
1. After ignoring we need at least drop IaC cashed results to let start new scan next time user press Run button. Otherwise we'll always show cashes results with already ignores issue. Better solution would be remove it from the tree.
2. After ignoring we need to do force update for `.snyk` file representation (VirtualFileSystsem and probably Psi) to display changes to user if they have opened it in the editor.
Shell we address it whiting this PR or in separate ticket/PR?

src/main/kotlin/io/snyk/plugin/services/download/CliDownloader.kt
bastiandoetsch: I'll take care of it, if it's conflicting during merge. Don't think it will be a problem.

ArtsiomCh (approved):
LGTM
PS Also might have sense to `dismiss` notification balloon after re-download start, but that is not so vital

PR #209, opened 2021-11-01
fix: 🐛 Network shouldn't be accessed in EDT or inside read action

bastiandoetsch:
Moved CLI download retry logic into background thread

### DIFF
--- a/src/main/kotlin/io/snyk/plugin/services/download/CliDownloaderErrorHandler.kt
+++ b/src/main/kotlin/io/snyk/plugin/services/download/CliDownloaderErrorHandler.kt
@@ -2,7 +2,9 @@ package io.snyk.plugin.services.download
 
 import com.intellij.ide.BrowserUtil
 import com.intellij.notification.NotificationAction
+import com.intellij.openapi.diagnostic.logger
 import com.intellij.openapi.progress.ProgressIndicator
+import com.intellij.openapi.progress.runBackgroundableTask
 import com.intellij.openapi.project.Project
 import com.intellij.util.io.HttpRequests
 import io.snyk.plugin.getCliFile
@@ -14,6 +16,9 @@ class CliDownloaderErrorHandler {
         SnykBalloonNotificationHelper.showError(message, project,
             NotificationAction.createSimple("Retry CLI download") {
                 project.getService(SnykCliDownloaderService::class.java).downloadLatestRelease(indicator, project)
+                runBackgroundableTask("Retry Snyk CLI Download", project, false) {
+                    project.getService(SnykCliDownloaderService::class.java).downloadLatestRelease(it, project)
+                }
             },
             NotificationAction.createSimple("Contact support...") {
                 BrowserUtil.browse("https://snyk.io/contact-us/?utm_source=JETBRAINS_IDE")
@@ -21,9 +26,20 @@ class CliDownloaderErrorHandler {
     }
 
     fun handleIOException(exception: IOException, indicator: ProgressIndicator, project: Project) {
-        val downloader = project.getService(SnykCliDownloaderService::class.java).downloader
-        downloader.downloadFile(getCliFile(), indicator)
-        showErrorWithRetryAndContactAction(getNetworkErrorNotificationMessage(exception), indicator, project)
+        retryDownload(project, indicator, getNetworkErrorNotificationMessage(exception))
+    }
+
+    private fun retryDownload(project: Project, indicator: ProgressIndicator, message: String) {
+        try {
+            runBackgroundableTask("Retry CLI Download", project, false) {
+                val downloader = project.getService(SnykCliDownloaderService::class.java).downloader
+                downloader.downloadFile(getCliFile(), indicator)
+            }
+        } catch (throwable: Throwable) { // we must catch throwable as IntelliJ could throw AssertionError
+            // IntelliJ throws an exception if we log an error, and in this case it is just the retry that failed
+            logger<CliDownloaderErrorHandler>().warn("Retry of downloading the Snyk CLI failed.", throwable)
+            showErrorWithRetryAndContactAction(message, indicator, project)
+        }
     }
 
     fun getNetworkErrorNotificationMessage(exception: IOException) =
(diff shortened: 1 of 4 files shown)


src/main/kotlin/io/snyk/plugin/services/download/CliDownloaderErrorHandler.kt
```diff
@@ -14,16 +16,30 @@ class CliDownloaderErrorHandler {
...
+    }
+
+    private fun retryDownload(project: Project, indicator: ProgressIndicator, message: String) {
+        try {
+            runBackgroundableTask("Retry CLI Download", project, false) {
+                val downloader = project.getService(SnykCliDownloaderService::class.java).downloader
+                downloader.downloadFile(getCliFile(), indicator)
+            }
+        } catch (throwable: Throwable) { // we must catch throwable as IntelliJ could throw AssertionError
+            // IntelliJ throws an exception if we log an error, and in this case it is just the retry that failed
```
michelkaporin, line 39: Love the comments ❤️

src/test/kotlin/io/snyk/plugin/services/download/CliDownloaderErrorHandlerTest.kt
```diff
@@ -4,79 +4,57 @@ import com.intellij.notification.NotificationAction
...
+    private lateinit var retryActionSlot: CapturingSlot<NotificationAction>
+    private lateinit var contactActionSlot: CapturingSlot<NotificationAction>
+    private lateinit var messageSlot: CapturingSlot<String>
+    private lateinit var projectSlot: CapturingSlot<Project>
+
     @Before
     fun setUp() {
-        cut = CliDownloaderErrorHandler()
+        clearAllMocks()
+        unmockkAll()
```
michelkaporin, line 35: Is there a need to call `unmockkAll` in both before and after function?

bastiandoetsch: Yes, because after the last test we would like to unmock, too, and we want to start with a clean slate for each test.

michelkaporin: Isn't it what tearDown function already does for you? 👀

bastiandoetsch: In general that's true - but not if you hard-stop the testing. In this case, if no VM is kept running it would probably be fine. I've seen it with database and filesystem changes though, that you're better off cleaning up before and after a test, to minimize potential for flakiness.

michelkaporin: Fair point, thanks for explaining. Resolving as it doesn't harm anyway :)

michelkaporin (approved):
LGTM

src/main/kotlin/io/snyk/plugin/services/download/CliDownloaderErrorHandler.kt
```diff
@@ -14,16 +16,30 @@ class CliDownloaderErrorHandler {
         SnykBalloonNotificationHelper.showError(message, project,
             NotificationAction.createSimple("Retry CLI download") {
                 project.getService(SnykCliDownloaderService::class.java).downloadLatestRelease(indicator, project)
+                runBackgroundableTask("Retry Snyk CLI Download", project, false) {
+                    project.getService(SnykCliDownloaderService::class.java).downloadLatestRelease(it, project)
```
ArtsiomCh, line 20: Kptlin has nice inline fun for that: project.service<T>()

src/main/kotlin/io/snyk/plugin/services/download/CliDownloaderErrorHandler.kt
```diff
@@ -14,16 +16,30 @@ class CliDownloaderErrorHandler {
...
-        showErrorWithRetryAndContactAction(getNetworkErrorNotificationMessage(exception), indicator, project)
+        retryDownload(project, indicator, getNetworkErrorNotificationMessage(exception))
+    }
+
+    private fun retryDownload(project: Project, indicator: ProgressIndicator, message: String) {
+        try {
+            runBackgroundableTask("Retry CLI Download", project, false) {
+                val downloader = project.getService(SnykCliDownloaderService::class.java).downloader
+                downloader.downloadFile(getCliFile(), indicator)
+            }
```
ArtsiomCh, line 37: duplicate of the runBackgroundableTask from above. Shell  we extract fun here?

src/main/kotlin/io/snyk/plugin/services/download/CliDownloaderErrorHandler.kt
```diff
@@ -14,16 +16,30 @@ class CliDownloaderErrorHandler {
...
+        } catch (throwable: Throwable) { // we must catch throwable as IntelliJ could throw AssertionError
```
ArtsiomCh, line 38: I would agree here with detekt and check if we can catch more specific type of exceptions here

ArtsiomCh:
We usually try to update CHANGELOG.md with each PR (under top level `## [Unreleased]`), so changes will be not lost at release's Change Notes

src/main/kotlin/io/snyk/plugin/services/download/CliDownloaderErrorHandler.kt
pavel-snyk: we are talking about this one: https://plugins.jetbrains.com/docs/intellij/plugin-services.html#retrieving-a-service

src/main/kotlin/io/snyk/plugin/services/download/CliDownloaderErrorHandler.kt
bastiandoetsch: Which class would you recommend? Unfortunately, AssertionError is in the error hierarchy, not in the exception hierarchy. Or would you prefer having two catch blocks?

src/main/kotlin/io/snyk/plugin/services/download/CliDownloaderErrorHandler.kt
bastiandoetsch: It's not duplicate - slight difference, added comment.

bastiandoetsch:
New PR on the way

PR #210, opened 2021-11-02
fix: IaC filtering issues by severity [ROAD-376]

ArtsiomCh:
chore: IntegTests refactoring:
We have to reset Settings for each tests otherwise changes in previous tests will affect currently running one. That leads to flakey, hard-to-debug tests.

### DIFF
--- a/src/integTest/kotlin/io/snyk/plugin/TestUtils.kt
+++ b/src/integTest/kotlin/io/snyk/plugin/TestUtils.kt
@@ -1,5 +1,12 @@
 package io.snyk.plugin
 
+import com.intellij.openapi.components.service
+import com.intellij.openapi.project.Project
+import io.snyk.plugin.services.SnykProjectSettingsStateService
+import java.time.Instant
+import java.util.Date
+import java.util.UUID
+
 fun setupDummyCliFile() {
     val cliFile = getCliFile()
 
@@ -7,4 +14,52 @@ fun setupDummyCliFile() {
         if (!cliFile.parentFile.exists()) cliFile.mkdirs()
         cliFile.createNewFile()
     }
+}
+
+fun removeDummyCliFile() {
+    val cliFile = getCliFile()
+    if (cliFile.exists()) {
+        cliFile.delete()
+    }
+}
+
+fun resetSettings(project: Project?) {
+    pluginSettings().apply {
+        token = null
+        customEndpointUrl = null
+        organization = null
+        ignoreUnknownCA = false
+        cliVersion = null
+
+        cliScanEnable = true
+
+        ossScanEnable = true
+        advisorEnable = true
+        snykCodeSecurityIssuesScanEnable = true
+        snykCodeQualityIssuesScanEnable = false
+        iacScanEnabled = false
+
+        sastOnServerEnabled = null
+        usageAnalyticsEnabled = true
+        crashReportingEnabled = true
+
+        lowSeverityEnabled = true
+        mediumSeverityEnabled = true
+        highSeverityEnabled = true
+        criticalSeverityEnabled = true
+
+        lastCheckDate = null
+        pluginFirstRun = true
+        pluginInstalled = false
+
+        pluginFirstInstallTime = Date.from(Instant.now())
+        lastTimeFeedbackRequestShown = Date.from(Instant.now())
+        showFeedbackRequest = true
+
+        scanningReminderWasShown = false
+
+        userAnonymousId = UUID.randomUUID().toString()
+    }
+
+    project?.service<SnykProjectSettingsStateService>()?.additionalParameters = null
 }
--- a/src/integTest/kotlin/io/snyk/plugin/TestUtils.kt
+++ b/src/integTest/kotlin/io/snyk/plugin/TestUtils.kt
@@ -1,11 +1,10 @@
 package io.snyk.plugin
 
-import com.intellij.openapi.components.service
+import com.intellij.openapi.application.ApplicationManager
 import com.intellij.openapi.project.Project
+import com.intellij.testFramework.replaceService
+import io.snyk.plugin.services.SnykApplicationSettingsStateService
 import io.snyk.plugin.services.SnykProjectSettingsStateService
-import java.time.Instant
-import java.util.Date
-import java.util.UUID
 
 fun setupDummyCliFile() {
     val cliFile = getCliFile()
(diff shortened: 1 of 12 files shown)


src/integTest/kotlin/io/snyk/plugin/cli/ConsoleCommandRunnerTest.kt
```diff
@@ -136,13 +137,15 @@ class ConsoleCommandRunnerTest : LightPlatformTestCase() {
...
         mockkObject(SentryErrorReporter)
         every { SentryErrorReporter.captureException(any()) } returns SentryId()
     }
 
     @After
     override fun tearDown() {
-        super.tearDown()
         unmockkAll()
+        resetSettings(project)
+        super.tearDown()
```
pavel-snyk, line 149: any reason to move `super.teadDown()` as last statement?

ArtsiomCh: after tearDown project will be Null. 
Not a big deal in that case but would be good to do any wrap up required project instance before that.

src/integTest/kotlin/io/snyk/plugin/TestUtils.kt
```diff
@@ -8,3 +15,51 @@ fun setupDummyCliFile() {
...
 }
+
+fun removeDummyCliFile() {
+    val cliFile = getCliFile()
+    if (cliFile.exists()) {
+        cliFile.delete()
+    }
+}
+
+fun resetSettings(project: Project?) {
```
bastiandoetsch, line 26: As discussed, we should have a look at just unregistering and re-registering each stateful service before each test. You could cast `getPicoContainer()` to `MutablePicoContainer` and then register/unregister. see https://www.programcreek.com/java-api-examples/?api=org.picocontainer.MutablePicoContainer

src/integTest/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanelTest.kt
```diff
@@ -0,0 +1,104 @@
...
+        getIacService(project).setConsoleCommandRunner(mockRunner)
+
+        // actual test run
+
+        project.service<SnykTaskQueueService>().scan()
+
+        PlatformTestUtil.waitWhileBusy(toolWindowPanel.getTree())
+
+        val rootIacIssuesTreeNode = toolWindowPanel.getRootIacIssuesTreeNode()
+        fun isMediumSeverityShown(): Boolean = rootIacIssuesTreeNode.children().asSequence()
```
bastiandoetsch, line 82: My preference would be to have that function defined on the class level, but that's just personal preference ;).

src/integTest/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanelTest.kt
```diff
@@ -0,0 +1,104 @@
...
+            ActionManager.getInstance().getAction("io.snyk.plugin.ui.actions.SnykTreeMediumSeverityFilterAction")
+                as SnykTreeMediumSeverityFilterAction
+        mediumSeverityFilterAction.setSelected(TestActionEvent(), false)
+
+        PlatformTestUtil.waitWhileBusy(toolWindowPanel.getTree())
+
+        assertFalse("Medium severity IaC results should NOT be shown after filtering", isMediumSeverityShown())
+
+        // restore modified Registry value
+        isIacEnabledRegistryValue.setValue(isIacEnabledOldValue)
```
bastiandoetsch, line 102: should this be in setup/teardown?

src/main/kotlin/io/snyk/plugin/services/SnykApplicationSettingsStateService.kt
```diff
@@ -16,6 +16,9 @@ import java.time.ZoneId
 import java.util.Date
 import java.util.UUID
 
+/**
+ * If default values changed then [io.snyk.plugin.TestUtilsKt.resetSettings] MUST be changed too!
+ */
```
bastiandoetsch, line 21: good to have that reminder, would prefer the other approach (see above), though.

src/main/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanel.kt
```diff
@@ -896,6 +898,12 @@ class SnykToolWindowPanel(val project: Project) : JPanel(), Disposable {
         (vulnerabilitiesTree.model as DefaultTreeModel).reload()
     }
 
+    @TestOnly
+    fun getRootIacIssuesTreeNode() = rootIacIssuesTreeNode
+
+    @TestOnly
+    fun getTree() = vulnerabilitiesTree
```
bastiandoetsch, line 905: This will be so helpful in other tests, too.

src/integTest/kotlin/io/snyk/plugin/TestUtils.kt
bastiandoetsch: Actually, Jetbrains provides a Utility Class for that :). https://upsource.jetbrains.com/idea-ce/file/idea-ce-8f0275fd7faaeafdb8900147eab3d256fe4221cb/platform/testFramework/src/com/intellij/testFramework/ServiceContainerUtil.kt?_ga=2.186386601.1807346972.1636115617-321360060.1633965402

ArtsiomCh: I like the idea. And thanks for finding ServiceContainerUtil (why do you delete that comment?). It looks like exactly what we need for such service's replacement.
I just feel such deep Integ tests refactoring is way out of scope for that ticket.
Also feel like we need to do some research how (and if) such replacement works. In the past we had problem with SnykApplicationSettingsStateService replacement. Probably due to it PersistentStateComponent nature, but not sure...
So would propose to explore that test's reset way in next tickets/PRs

src/integTest/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanelTest.kt
ArtsiomCh: That what I've been missing in Java after Pascal: internal functions. :)
Feel like here there're no needs to expose such fun on higher(class) level as it's needed only inside parent fun.
But if it harder to read the code for you that way, I can extract it to higher level.

src/integTest/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanelTest.kt
ArtsiomCh: That registry settings is changed explicitly in that test only. So I think we should revert it back here explicitly too and not in tearDown which executed for all (future) tests in that class.
If we later find the way to reset the whole App state for each test to it initial state, that would solve the problem and no such revert setting will be needed.

src/main/kotlin/io/snyk/plugin/services/SnykApplicationSettingsStateService.kt
ArtsiomCh: I like the other approach too. But it need some research and effort to implement.
So until we implement that better/proper way, current "workaround" for resetting Settings is better then what we had before (no reset at all and flaky test due of that)

src/integTest/kotlin/io/snyk/plugin/TestUtils.kt
bastiandoetsch: Dunno why it got away...

src/integTest/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanelTest.kt
bastiandoetsch: Na, it's okay. Just preference and not being used to it. Probably it's better style to have it where it is used...

src/integTest/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanelTest.kt
bastiandoetsch: Ok, sounds like a plan. Hoping for resetting the whole app state. Else it would be cleaner to have setup/teardown ensure that the expected pre-condition is met, without a test having to cleanup after itself (jumping out of a test with an exception has happened when debugging to me).

src/integTest/kotlin/io/snyk/plugin/TestUtils.kt
ArtsiomCh: Hmm... surprisingly it works. Hope we don't create another "time bomb" by doing such replacement...
PS @pavel-snyk  we should consider to clean up the same way in the future at least SnykToolWindowPanel as it holds state of cached scan results.

src/integTest/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanelTest.kt
ArtsiomCh: I'm afraid resetting the whole Registry might be out of our control as it's IntelliJ own class....

bastiandoetsch (approved):
LGTM

PR #213, opened 2021-11-04
IlligalStateException  due to different base paths for modules [ROAD-496]

ArtsiomCh:
fix: Snyk Code scan for multi-modules project cause `IlligalStateException` due to different base paths for modules.

PS also added check for `VirtualFile.isValid` whenever we call `PsiManager.findFile()` to prevent getting errors for invalid (deleted/disposed) files.


src/integTest/kotlin/io/snyk/plugin/snykcode/core/PduTest.kt
```diff
@@ -0,0 +1,32 @@
+package io.snyk.plugin.snykcode.core
+
+import com.intellij.openapi.util.io.FileUtil
+import com.intellij.openapi.vfs.LocalFileSystem
+import com.intellij.psi.PsiManager
+import com.intellij.testFramework.LightPlatformTestCase
+import com.intellij.testFramework.PlatformTestUtil
+import java.io.File
+
+class PduTest : LightPlatformTestCase() {
```
bastiandoetsch, line 10: Shouldn't that be PDUTest?

src/integTest/kotlin/io/snyk/plugin/snykcode/core/PduTest.kt
```diff
@@ -0,0 +1,32 @@
...
+
+    fun testOutOfProjectFile_shouldBeCorrectlyConvertingPath2PsiAndVs() {
+        val testFilePath = FileUtil.toSystemIndependentName(PlatformTestUtil.getCommunityPath()) + "/testFile"
+        File(testFilePath).createNewFile()
```
bastiandoetsch, line 14: For less manual cleanup, you could use File.createTempFile() and File.deleteOnExit() (see https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html)

src/main/kotlin/io/snyk/plugin/snykcode/core/PDU.kt
```diff
@@ -40,26 +40,23 @@ class PDU private constructor() : PlatformDependentUtilsBase() {
...
             ?: throw IllegalStateException("No Project base Path found for file $psiFile")
         // `ignoreCase = true` needed due to https://youtrack.jetbrains.com/issue/IDEA-268081
-        val projectBasedPath = absolutePath.replaceFirst(projectPath, "", ignoreCase = true)
-        if (projectBasedPath == absolutePath) {
-            throw IllegalStateException("projectBasedPath is not valid:\n" +
-                "  psiFile.virtualFile.path = $absolutePath\n" +
-                "  psiFile.project.basePath = $projectPath")
-        }
-        return projectBasedPath
+        return absolutePath.replaceFirst(projectPath, "", ignoreCase = true)
```
bastiandoetsch, line 43: Ignoring the case could lead to bugs on MacOS and Linux, as a file with different case would be a different file. How about covering this scenario with a test, so we are certain it works?

src/main/kotlin/io/snyk/plugin/snykcode/core/PDU.kt
```diff
@@ -40,26 +40,23 @@ class PDU private constructor() : PlatformDependentUtilsBase() {
...
+            ?: LocalFileSystem.getInstance().findFileByPath(path)
         if (virtualFile == null) {
-            SCLogger.instance.logWarn("VirtualFile not found for: $absolutePath")
+            SCLogger.instance.logWarn("VirtualFile not found for: $absolutePath or $path")
             return null
         }
         return RunUtils.computeInReadActionInSmartMode(
             prj,
-            Computable { PsiManager.getInstance(prj).findFile(virtualFile) }
+            Computable {
```
bastiandoetsch, line 57: Computable class doc says one should use java.util.function.Supplier instead.

src/integTest/kotlin/io/snyk/plugin/snykcode/core/PduTest.kt
pavel-snyk: It makes sense to use JUnit rule for TemporaryFile and Folder: https://junit.org/junit4/javadoc/4.12/org/junit/rules/TemporaryFolder.html

src/main/kotlin/io/snyk/plugin/snykcode/core/PDU.kt
ArtsiomCh: ignoring the case was added exactly to avoid bugs on Mac/Linux :) See comment above the line:
`ignoreCase = true` needed due to https://youtrack.jetbrains.com/issue/IDEA-268081

src/integTest/kotlin/io/snyk/plugin/snykcode/core/PduTest.kt
ArtsiomCh: Looks like deleteOnExit will not help us as it's javadocs says: "deleted when the virtual machine terminates" which has no sense in our case as we need it to be deleted after test method execution. So manual deletion needed here anyway. Thanks for pointing that out!

src/main/kotlin/io/snyk/plugin/snykcode/core/PDU.kt
ArtsiomCh: `Computable` is used under nice in IntelliJ API call: `DumbService#runReadActionInSmartMode(Computable<T>)`
Looks like no benefits with usage of Supplier here and then transform it to Computable for that API call.

src/integTest/kotlin/io/snyk/plugin/snykcode/core/PduTest.kt
bastiandoetsch: I like Pavel's JUnit stuff. Dunno if it's available in JUnit 3 (`LightPlatformTestcase`) though.

src/main/kotlin/io/snyk/plugin/snykcode/core/PDU.kt
bastiandoetsch: I still don't understand it - I looked at the issue, and github is reported with different case, it seems. Still, different case means, different file/directory on Linux/macOs, so it should not be treated case-insensitive. For me the question is why in the you track issue the virtual file is reporting a different path, and if that cannot be helped, if we should differentiate between platforms.

ArtsiomCh: The problem is that on Mac/Linux we have different case for the **same** dir returned from those api:
PsiFile.project.basePath
PsiFile.virtualFile.path
And we use the text of `basePath` to subtract it from full-file-path to get project-related-path.
If we don't do it case insensitive then basePath is treated as different paths and will not be properly subtracted.
You're right, in some edge cases: two files with the same basePath (just different in case) that might cause an issue. But it seems to be not realist in real world so far.
PS we are going to get off project.basePath anyway (due to issue with multi-module projects that have no basePath) so would leave it as is for now, and it will be fixed automatically after project.basePath issue fixed.

bastiandoetsch: Okay, let's go ahead. If we chuck project.basePath anyway, no use to cater to that special edge case right now. It's weird though, that project.basePath is not returning the case-correct path.

bastiandoetsch (approved):

src/integTest/kotlin/io/snyk/plugin/snykcode/core/PduTest.kt
ArtsiomCh: Just tried to do it through TemporaryFolder and for some reason neither `TemporaryFolder#before` (responsible for actual creation of temp folder) nor `TemporaryFolder#after` (responsible for deletion of all created temp files/dirs) are not executed. Any idea why that could be?...

ArtsiomCh: Seems like JUnit annotations (`@Rule` in our case) are ignored when run LightPlatformTestCase

PR #219, opened 2021-11-12
Refactor SnykToolWindowPanel unit test to not use jetbrains test base class

bastiandoetsch:
also contains the commit to only enable experiments if tracking is enabled

### DIFF
--- a/src/main/kotlin/snyk/amplitude/AmplitudeExperimentService.kt
+++ b/src/main/kotlin/snyk/amplitude/AmplitudeExperimentService.kt
@@ -68,6 +68,7 @@ class AmplitudeExperimentService : Disposable {
     }
 
     fun isShowScanningReminderEnabled(): Boolean {
+        pluginSettings().usageAnalyticsEnabled || return false
         val variant = storage["intellij-show-scanning-reminder"] ?: return false
 
         val settings = pluginSettings()
@@ -77,6 +78,7 @@ class AmplitudeExperimentService : Disposable {
     }
 
     fun isPartOfExperimentalWelcomeWorkflow(): Boolean {
+        pluginSettings().usageAnalyticsEnabled || return false
         val variant = storage[CHANGE_AUTHENTICATE_BUTTON] ?: return false
         return variant.value == TEST_GROUP
     }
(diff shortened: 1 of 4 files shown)


src/main/kotlin/snyk/amplitude/AmplitudeExperimentService.kt
```diff
@@ -76,7 +77,8 @@
~        return variant.value == "test" && !settings.scanningReminderWasShown
     }
```
ArtsiomCh, line 79: wow! didn't know that trick, how did you found it? can't find it in official docs....

src/main/kotlin/snyk/amplitude/AmplitudeExperimentService.kt
```diff
@@ -77,6 +78,7 @@ class AmplitudeExperimentService : Disposable {
...
 
     fun isPartOfExperimentalWelcomeWorkflow(): Boolean {
+        pluginSettings().usageAnalyticsEnabled || return false
```
ArtsiomCh, line 81: wow! didn't know that trick. How did you found it? Can't find it in official docs....

bastiandoetsch: I just wrote it as I would be writing it in shell scripting, and surprisingly it worked...

bastiandoetsch:
bump @ArtsiomCh @pavel-snyk . Please review (again).

ArtsiomCh:
As I said before, I'm against mocking Jetbrains classes and if that needed, imho, JB test framework should be used. So I'm not going to use that testing approach even if merged. Would be not happy to fix/change such tests when introducing new functionality, but I'm ok with that if @pavel-snyk found it useful.

bastiandoetsch:
@pavel-snyk please comment - btw., without this change, we would keep a test inheriting from LightPlatformTestCase in Unit tests.

ArtsiomCh (changes requested):
Besides I'm still against that approach, I feel @bastiandoetsch you need that for your TTD developing way. 
I'm willing to approve it after merge conflict resolving (and other code clean up if needed to match current `master` state)
PS Please use that mocks with caution and always think about fellow developers who will need to adjust such tests later.

bastiandoetsch:
I won't merge it now - it is failing now after rebasing with master. And the exception is very weird, so as it's not really and very important, I'll let it rest.

PR #220, opened 2021-11-12
fix: allow experiments before authentication [ROAD-472]

bastiandoetsch:
Previously, variants were only fetched from Amplitude when a user was already authenticate. This is not enough, when we want to experiment before authentication.

Now the variants are fetched on startup even if there is no token, using the anonymousID from segment as deviceId in the Amplitude Experiment API as additional parameter. Thus, when userID is empty, the deviceId is still filled and experiments can be assigned to users.


bastiandoetsch:
will fix this on monday

src/main/kotlin/snyk/amplitude/AmplitudeExperimentService.kt
```diff
@@ -19,11 +19,11 @@ private val LOG = logger<AmplitudeExperimentService>()
 class AmplitudeExperimentService : Disposable {
     companion object {
         const val CHANGE_AUTHENTICATE_BUTTON = "ide-jetbrains-change-authenticate-button-and-first-time-workflow"
-        const val TEST_GROUP = "TEST"
+        const val TEST_GROUP = "treatment"
```
michelkaporin, line 22: Why change it to `treatment` if we had it defined as `test` for the previous experiment? I'd rather rename the group on Amplitude side.

Also there's possibility that we will be having more groups when running A/B/C test in future.

src/main/kotlin/snyk/amplitude/AmplitudeExperimentService.kt
```diff
@@ -19,11 +19,11 @@ private val LOG = logger<AmplitudeExperimentService>()
...
+        private val storage: ConcurrentHashMap<String, Variant> = ConcurrentHashMap()
```
michelkaporin, line 23: Out of curiosity - what's the semantic difference between having storage outside and inside companion object? 👀

src/main/kotlin/snyk/amplitude/api/AmplitudeExperimentApiClient.kt
```diff
@@ -20,15 +20,15 @@ class AmplitudeExperimentApiClient private constructor(
      */
     @Suppress("TooGenericExceptionCaught")
     fun allVariants(user: ExperimentUser): Map<String, Variant> {
-        if (user.userId.isBlank()) {
-            log.warn("user id is empty; amplitude may not resolve identity")
+        if (user.userId.isBlank() && user.deviceId.isBlank()) {
```
michelkaporin, line 23: Why would we even try to fetch the variants if both properties are blank? This will lead to undefined behaviour and wouldn't count towards the experiment data.

src/main/resources/application.properties
```diff
@@ -1,4 +1,6 @@
-amplitude.experiment.api-key=@amplitude.experiment.api-key@
+amplitude.experiment.api-key=client-zkBHsOK3KSAyPyCCvIXlXgdlACNhTjHD
```
michelkaporin, line 1: ⛔

michelkaporin:
@bastiandoetsch let's link up the ticket here

src/main/kotlin/snyk/amplitude/AmplitudeExperimentService.kt
bastiandoetsch: Static storage vs local storage

src/main/kotlin/snyk/amplitude/api/AmplitudeExperimentApiClient.kt
bastiandoetsch: Because we still fetch it with a deviceID, so we don't need the userID.

src/main/resources/application.properties
bastiandoetsch: Argh.

src/main/kotlin/snyk/amplitude/AmplitudeExperimentService.kt
michelkaporin: thank you!

src/main/kotlin/snyk/amplitude/api/AmplitudeExperimentApiClient.kt
michelkaporin: let me ask you differently - what's the purpose of fetching it with a blank device id and blank userId?

src/main/kotlin/snyk/amplitude/AmplitudeExperimentService.kt
bastiandoetsch: Treatment is the default on Amplitude - and I'd rather stick to their default. But in the end, I don't care much.

src/main/kotlin/snyk/amplitude/api/AmplitudeExperimentApiClient.kt
bastiandoetsch: The deviceID is never blank, as it is automatically populated on instance creation with the default value.

build.gradle.kts
```diff
@@ -24,15 +24,15 @@ repositories {
...
+  implementation("io.sentry:sentry:5.3.0")
   implementation("io.snyk.code.sdk:snyk-code-client:2.1.12")
-  implementation("ly.iterative.itly:plugin-iteratively:1.2.10")
-  implementation("ly.iterative.itly:plugin-schema-validator:1.2.10") {
+  implementation("ly.iterative.itly:plugin-iteratively:1.2.11")
+  implementation("ly.iterative.itly:plugin-schema-validator:1.2.11") {
     exclude(group = "org.slf4j")
   }
-  implementation("ly.iterative.itly:sdk-jvm:1.2.10")
+  implementation("ly.iterative.itly:sdk-jvm:1.2.11")
```
ArtsiomCh, line 35: What's the reason to update versions above? 
Are we sure we don't get new bugs with such update?

src/main/kotlin/io/snyk/plugin/SnykPostStartupActivity.kt
```diff
@@ -71,16 +71,18 @@ class SnykPostStartupActivity : StartupActivity.DumbAware {
...
-            val publicUserId = service<SnykApiService>().userId ?: ""
-            val experimentUser = ExperimentUser(publicUserId)
-            service<AmplitudeExperimentService>().fetch(experimentUser)
-
-            if (service<AmplitudeExperimentService>().isShowScanningReminderEnabled()) {
-                SnykBalloonNotifications.showScanningReminder(project)
-                settings.scanningReminderWasShown = true
-            }
+            publicUserId = service<SnykApiService>().userId ?: ""
+        }
```
ArtsiomCh, line 77: minor: could be `val publicUserId = if ...`

ArtsiomCh (approved):
LGTM
Please address minor notes from my side.
Also please get approval from @michelkaporin before merging as he know better how all that Amplitude's experiment magic works.

build.gradle.kts
bastiandoetsch: Yeah, it's only a minor upgrade and I tested it also manually against amplitude and verified events are sent.

michelkaporin (approved):

PR #221, opened 2021-11-15
fix: IaC errors display in description panel [ROAD-486]

ArtsiomCh:
image

### DIFF
--- a/src/integTest/kotlin/UIComponentFinder.kt
+++ b/src/integTest/kotlin/UIComponentFinder.kt
@@ -1,0 +1,24 @@
+import java.awt.Component
+import java.awt.Container
+import kotlin.reflect.KClass
+import kotlin.reflect.cast
+
+object UIComponentFinder {
+
+    @OptIn(ExperimentalStdlibApi::class)
+    fun <T: Component> getComponentByName(parent: Container, clazz: KClass<T>, name: String): T? {
+        val components = parent.components
+        var found: T? = null
+        for (component in components) {
+            if (clazz.isInstance(component) && name == component.name) {
+                found = clazz.cast(component)
+            } else if (component is Container) {
+                found = getComponentByName(component, clazz, name)
+            }
+            if (found != null) {
+                break
+            }
+        }
+        return found
+    }
+}
--- a/src/integTest/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanelIntegTest.kt
+++ b/src/integTest/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanelIntegTest.kt
@@ -45,7 +45,6 @@ class SnykToolWindowPanelIntegTest : HeavyPlatformTestCase() {
         setupDummyCliFile()
         // restore modified Registry value
         isIacEnabledRegistryValue.setValue(isIacEnabledDefaultValue)
-        unmockkStatic("io.snyk.plugin.UtilsKt")
     }
 
     override fun tearDown() {
@@ -54,7 +53,6 @@ class SnykToolWindowPanelIntegTest : HeavyPlatformTestCase() {
         removeDummyCliFile()
         // restore modified Registry value
         isIacEnabledRegistryValue.setValue(isIacEnabledDefaultValue)
-        unmockkStatic("io.snyk.plugin.UtilsKt")
         super.tearDown()
     }
 
--- a/src/integTest/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanelIntegTest.kt
+++ b/src/integTest/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanelIntegTest.kt
@@ -6,9 +6,12 @@ import com.intellij.openapi.util.registry.Registry
 import com.intellij.testFramework.HeavyPlatformTestCase
 import com.intellij.testFramework.PlatformTestUtil
 import com.intellij.testFramework.TestActionEvent
+import com.intellij.util.ui.tree.TreeUtil
 import io.mockk.every
 import io.mockk.mockk
+import io.mockk.mockkStatic
 import io.mockk.unmockkAll
+import io.mockk.unmockkStatic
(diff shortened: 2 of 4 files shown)


src/integTest/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanelIntegTest.kt
```diff
@@ -98,7 +113,38 @@ class SnykToolWindowPanelIntegTest : HeavyPlatformTestCase() {
...
+        assertEquals(iacError, toolWindowPanel.currentIacError)
+
+        TreeUtil.selectNode(toolWindowPanel.getTree(), toolWindowPanel.getRootIacIssuesTreeNode())
+        PlatformTestUtil.waitWhileBusy(toolWindowPanel.getTree())
+
+        val descriptionComponents = toolWindowPanel.getDescriptionPanel().components.toList()
+        assertTrue(descriptionComponents.any { it is SnykErrorPanel })
+
+        // tear Down
+        unmockkStatic("io.snyk.plugin.UtilsKt")
```
bastiandoetsch, line 147: This should be done in teardown, as it is never called if the assertion fail.

src/integTest/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanelIntegTest.kt
```diff
@@ -98,7 +113,38 @@ class SnykToolWindowPanelIntegTest : HeavyPlatformTestCase() {
...
+        every { getIacService(project).isCliInstalled() } returns true
+        every { getIacService(project).scan() } returns iacResultWithError
+
+        // actual test run
+        project.service<SnykTaskQueueService>().scan()
+
+        val toolWindowPanel = project.service<SnykToolWindowPanel>()
+        PlatformTestUtil.waitWhileBusy(toolWindowPanel.getTree())
+
+        assertEquals(iacError, toolWindowPanel.currentIacError)
```
bastiandoetsch, line 138: Consider to use verify { ... } to make sure that a service call is actually happening and the description panel is populated by the result of the call, not from any other error.

src/integTest/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanelIntegTest.kt
```diff
@@ -98,7 +113,38 @@ class SnykToolWindowPanelIntegTest : HeavyPlatformTestCase() {
...
+
+        TreeUtil.selectNode(toolWindowPanel.getTree(), toolWindowPanel.getRootIacIssuesTreeNode())
+        PlatformTestUtil.waitWhileBusy(toolWindowPanel.getTree())
+
+        val descriptionComponents = toolWindowPanel.getDescriptionPanel().components.toList()
+        assertTrue(descriptionComponents.any { it is SnykErrorPanel })
```
bastiandoetsch, line 144: Please check for the actual error content, to make sure it's not a different error displayed :). You can get the Label with UIComponentFinder class.

src/integTest/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanelIntegTest.kt
ArtsiomCh: Here we check intermediate condition required for the test of the test to have sense.
Also I don't want to chain the test to any particulate service or method call. But instead would like to check that we have required final result: error shown in description panel.
PS But Thanks for clarification! will add check for the expected error text and not any random error.

src/integTest/kotlin/UIComponentFinder.kt
```diff
@@ -0,0 +1,24 @@
+import java.awt.Component
+import java.awt.Container
+import kotlin.reflect.KClass
+import kotlin.reflect.cast
+
+object UIComponentFinder {
+
+    @OptIn(ExperimentalStdlibApi::class)
```
bastiandoetsch, line 8: Interesting, what's that?

src/integTest/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanelIntegTest.kt
```diff
@@ -41,17 +43,23 @@ class SnykToolWindowPanelIntegTest : HeavyPlatformTestCase() {
         unmockkAll()
         resetSettings(project)
         setupDummyCliFile()
+        // restore modified Registry value
+        isIacEnabledRegistryValue.setValue(isIacEnabledDefaultValue)
+        unmockkStatic("io.snyk.plugin.UtilsKt")
```
bastiandoetsch, line 48: The unmockStatic looks redundant to unmockAll().

src/integTest/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanelIntegTest.kt
```diff
@@ -41,17 +43,23 @@ class SnykToolWindowPanelIntegTest : HeavyPlatformTestCase() {
...
     }
 
     override fun tearDown() {
         unmockkAll()
         resetSettings(project)
         removeDummyCliFile()
+        // restore modified Registry value
+        isIacEnabledRegistryValue.setValue(isIacEnabledDefaultValue)
+        unmockkStatic("io.snyk.plugin.UtilsKt")
```
bastiandoetsch, line 57: Same here, why do you unmock utils explicitly and not just call unmockAll() at the end of the method?

src/integTest/kotlin/UIComponentFinder.kt
ArtsiomCh: It's easy to find, just comment it and Idea will show you ;)
Actually it's needed for KClass<T>.cast() call as it's for some reason still an experimental function

bastiandoetsch (approved):

PR #222, opened 2021-11-16
feat: Identify running docker daemon [ROAD-456]

bastiandoetsch:
snyk container test <image-name> command requires running Docker daemon. The scope of this ticket is finding the way to identify running Docker daemon (maybe with optional IntelliJ Docker plugin?)

Implemented using the testcontainers library that has that functionality already.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -33,6 +33,7 @@ dependencies {
     exclude(group = "org.slf4j")
   }
   implementation("ly.iterative.itly:sdk-jvm:1.2.10")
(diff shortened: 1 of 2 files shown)


ArtsiomCh:
Referring to the conversation in Slack, I don't think we should introduce new dependency to the plugin for the functionality we don't really needed for `snyk container test` command.

ArtsiomCh:
Referring to the conversation in Slack, I don't think we should introduce new dependency to the plugin for the functionality we don't really needed for `snyk container test` command.

bastiandoetsch:
@pavel-snyk please review :)

PR #233, opened 2021-12-10
Oss plugin displays error if no supported package manager is found [ROAD-463]

bastiandoetsch:
- extracted SnykScanListener closure as variable to be able to directly call it
- added a constant that identifies the error message that should be an info
- added an if to check for that constant
- wrote an integration test verifying that this error message is displayed as info, others as error

### DIFF
--- a/src/integTest/kotlin/snyk/advisor/PackageNameProviderTest.kt
+++ b/src/integTest/kotlin/snyk/advisor/PackageNameProviderTest.kt
@@ -1,8 +1,19 @@
 package snyk.advisor
 
 import com.intellij.testFramework.fixtures.BasePlatformTestCase
+import io.snyk.plugin.pluginSettings
 
 class PackageNameProviderTest : BasePlatformTestCase() {
+
+    override fun setUp() {
+        super.setUp()
+        pluginSettings().fileListenerEnabled = false
+    }
+
+    override fun tearDown() {
+        super.tearDown()
+        pluginSettings().fileListenerEnabled = true
+    }
 
     private fun getResourceAsString(resourceName: String): String = javaClass.classLoader
         .getResource(resourceName)!!.readText(Charsets.UTF_8)
--- a/src/integTest/kotlin/snyk/iac/annotator/IacBaseAnnotatorCase.kt
+++ b/src/integTest/kotlin/snyk/iac/annotator/IacBaseAnnotatorCase.kt
@@ -4,6 +4,7 @@ import com.intellij.testFramework.fixtures.BasePlatformTestCase
 import com.intellij.testFramework.replaceService
 import io.mockk.mockk
 import io.mockk.unmockkAll
+import io.snyk.plugin.pluginSettings
 import io.snyk.plugin.ui.toolwindow.SnykToolWindowPanel
 import org.junit.Before
 import java.nio.file.Paths
@@ -25,6 +26,7 @@ abstract class IacBaseAnnotatorCase : BasePlatformTestCase() {
         super.setUp()
         unmockkAll()
         project.replaceService(SnykToolWindowPanel::class.java, toolWindowPanel, project)
+        pluginSettings().fileListenerEnabled = false
     }
 
     @Suppress("SwallowedException", "TooGenericExceptionCaught")
@@ -32,6 +34,7 @@ abstract class IacBaseAnnotatorCase : BasePlatformTestCase() {
         unmockkAll()
         try {
             super.tearDown()
+            pluginSettings().fileListenerEnabled = true
         } catch (e: Exception) {
             // when tearing down the test case, our File Listener is trying to react on the deletion of the test
             // files and tries to access the file index that isn't there anymore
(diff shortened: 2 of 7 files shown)


ArtsiomCh:
Well done with notification part! 
We just need also bring the tree and the description panel into correct state too:
image

src/main/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanel.kt
```diff
@@ -971,6 +988,7 @@
     }
```
ArtsiomCh, line 988: @TestOnly

src/main/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanel.kt
```diff
@@ -971,6 +988,7 @@ class SnykToolWindowPanel(val project: Project) : JPanel(), Disposable {
...
 
     fun getRootIacIssuesTreeNode() = rootIacIssuesTreeNode
+    fun getRootOssIssuesTreeNode() = rootOssTreeNode
```
ArtsiomCh, line 991: `@TestOnly`

src/main/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanel.kt
```diff
@@ -615,19 +627,22 @@ class SnykToolWindowPanel(val project: Project) : JPanel(), Disposable {
...
+                        count == NODE_INITIAL_STATE -> ""
+                        count == 0 -> NO_ISSUES_FOUND_TEXT
+                        count > 0 -> " - $count vulnerabilit${if (count > 1) "ies" else "y"}$addHMLPostfix"
+                        else -> throw IllegalStateException()
+                    }
                 }
-            }
         }
-        newOssTreeNodeText?.let { rootOssTreeNode.userObject = it }
+        newOssTreeNodeText.let { rootOssTreeNode.userObject = it }
```
ArtsiomCh, line 645: Code above will not show `(scanning...)` and `(error)` states.
Also logic become inconsistent with other nodes and not following the method KDocs. I.e. `null` value for `ossResultsCount` should left node text as-is, due to it changes in other places, like `VulnerabilityTreeCellRenderer` 
Would suggest instead of that breaking changes do call 
`updateTreeRootNodesPresentation(ossResultsCount = NODE_INITIAL_STATE)` from scanningOssError()

src/main/kotlin/io/snyk/plugin/SnykBulkFileListener.kt
```diff
@@ -45,7 +46,13 @@ class SnykBulkFileListener : BulkFileListener {
 
         if (isContainerEnabled()) {
             for (project in ProjectUtil.getOpenProjects()) {
-                getKubernetesImageCache(project).extractFromEvents(events)
+                try {
+                    getKubernetesImageCache(project).extractFromEvents(events)
+                } catch (e: ClassCastException) {
+                    // if this exception occurs, IntelliJ is in an undefined state.
+                    // Outside of tests, this should not occur
+                    Logger.getInstance(javaClass).warn(e)
```
ArtsiomCh, line 54: looks like ugly workariund... could we found the resaon for that ClassCastException in tetst instead?

bastiandoetsch: And doesn't work. I can't reproduce it locally - it doesn't happen even if I test the action via act -j test

bastiandoetsch: Feel free to have a look - it has nothing to do with the actual PR functionality.

bastiandoetsch: I guess, we have to add a flag to enable/disable the filelistener with default to enable, to be able to disable it for certain tests.

bastiandoetsch: Also, the reason is quite clear - we try to access ProjectRootManager when it is not available during tests. Maybe we need to make it dumb-aware or such.

src/integTest/kotlin/snyk/iac/annotator/IacBaseAnnotatorCase.kt
```diff
@@ -25,6 +26,7 @@ abstract class IacBaseAnnotatorCase : BasePlatformTestCase() {
         super.setUp()
         unmockkAll()
         project.replaceService(SnykToolWindowPanel::class.java, toolWindowPanel, project)
+        pluginSettings().fileListenerEnabled = false
```
ArtsiomCh, line 29: restore in tearDown

src/integTest/kotlin/snyk/advisor/PackageNameProviderTest.kt
```diff
@@ -1,9 +1,15 @@
 package snyk.advisor
 
 import com.intellij.testFramework.fixtures.BasePlatformTestCase
+import io.snyk.plugin.pluginSettings
 
 class PackageNameProviderTest : BasePlatformTestCase() {
 
+    override fun setUp() {
+        super.setUp()
+        pluginSettings().fileListenerEnabled = false
```
ArtsiomCh, line 10: restore in tearDown

ArtsiomCh (approved):
Ok, for now as workaround for that tests issue
@bastiandoetsch please address comments

PR #257, opened 2021-12-20
Facilitate Release Process

bastiandoetsch:
Automatically determine the right semantic version from git history, create a tag and push to origin with the next version when e.g. `./gradlew release -Prelease.versionIncrementer=incrementMajor` is executed. Executing just `./gradlew release` will increase patch level, using `./gradlew release -Prelease.versionIncrementer=incrementMinor` will [increment](https://axion-release-plugin.readthedocs.io/en/latest/configuration/version/) the middle number. This all is done using the [axion release plugin](https://github.com/allegro/axion-release-plugin). This way, we do not need to manually set the version in gradle.properties. Developer releases on branches use `-SNAPSHOT` convention, e.g. `2.4.9-SNAPSHOT`.

Automatically create github release when release workflow is triggered, done using the [automatic release action](https://github.com/marvinpinto/action-automatic-releases) recommended by github.

Updated editorconfig to recognise *.kts files and reformat build.gradle.kts. Now detekt plugin is not complaining anymore.

### DIFF
--- a/.github/workflows/weekly-release.yml
+++ b/.github/workflows/weekly-release.yml
@@ -2,7 +2,7 @@ name: Weekly Release
 
 on:
   schedule:
-    - cron: '0 10 * * 3'  # every wednesday at 10 am
+    - cron: '0 10 * * 1'  # every Monday at 10 am UTC
 
 jobs:
   create-release-tag:
--- a/.github/workflows/weekly-release.yml
+++ b/.github/workflows/weekly-release.yml
@@ -1,0 +1,41 @@
+name: Weekly Release
+
+on:
+  schedule:
+    - cron: '0 10 * * 3'  # every wednesday at 10 am
+
+jobs:
+  create-release-tag:
+    name: create release tag
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Fetch Sources
+        uses: actions/checkout@v2
+
+      - name: Setup Java
+        uses: actions/setup-java@v1
+        with:
+          java-version: 1.8
+
+      - name: Setup Gradle Cache
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.gradle/caches
+            ~/.gradle/wrapper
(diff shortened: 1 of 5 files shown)


ArtsiomCh:
Looks like magic for me... Let's @pavel-snyk have a look. I'm not expert in that yaml things...
PS also with that changes to publish plugin we'll run `./gradlew release` but not creating tag in git, right?

bastiandoetsch:
Yeah, exactly.

michelkaporin:
😍 love it. What's the plan with version increment? Will we have separate GH actions to release patch, minor and major?

bastiandoetsch:
> 😍 love it. What's the plan with version increment? Will we have separate GH actions to release patch, minor and major?

I didn't think of it. I would think of:
- CI/CD with patch increments
- minor, major releases would be done manually.

General question would be anyway, if we need to differentiate, as we don't provide a library - so doing patch releases would theoretically be fine for everything, as the plugin should not be sourced as a dependency.

bastiandoetsch:
We could also move from semver releases to a date/time-based versioning scheme, which I personally prefer. Like `2022.0114080823`.

bastiandoetsch:
I've now added the workflow for a weekly automatic patch release. Please review.

.github/workflows/weekly-release.yml
```diff
@@ -0,0 +1,41 @@
+name: Weekly Release
+
+on:
+  schedule:
+    - cron: '0 10 * * 3'  # every wednesday at 10 am
```
michelkaporin, line 5: I'd prefer to release earlier. Given the max review time, a release might get published only on 10am Fridays with another delay until user base will install it. I think Monday would be the best time to release, imo.

Also what'll be the timezone here?

bastiandoetsch: It's in UTC, I'll change to Mondays :).

michelkaporin (approved):
LGTM 👍

bastiandoetsch:
Merging now, we'll keep an eye out for Monday 10 a.m UTC/11 a.m CET

PR #268, opened 2022-01-13
feat: ✨ enable new welcome workflow for all new users [ROAD-633]

bastiandoetsch:


src/main/kotlin/io/snyk/plugin/ui/toolwindow/SnykAuthPanel.kt
```diff
@@ -29,18 +28,20 @@ import snyk.analytics.AuthenticateButtonIsClicked
...
 import java.awt.event.ActionEvent
 import javax.swing.AbstractAction
 import javax.swing.JButton
 import javax.swing.JLabel
 import javax.swing.JPanel
 
 class SnykAuthPanel(val project: Project) : JPanel(), Disposable {
+    companion object {
+        const val AUTHENTICATE_BUTTON_TEXT = "Test code now"
+    }
```
ArtsiomCh, line 40: minor: Companion object should be in the end by [convention](https://kotlinlang.org/docs/coding-conventions.html#source-file-organization) we currently follow (but [that's discussable](https://discuss.kotlinlang.org/t/companion-objects-on-bottom/17797))

ArtsiomCh:
`SnykToolWindowPanel#enableProductsAccordingToServerSetting` override **incorrectly** the default enablement for products.
Correct enablement (confirmed by Georgi) is presented in SnykApplicationSettingsStateService

ArtsiomCh:
Showing Tree before getting token we have Code Security node in wrong state:
image

src/main/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanel.kt
```diff
@@ -546,15 +545,11 @@ class SnykToolWindowPanel(val project: Project) : JPanel(), Disposable {
...
             settings.token.isNullOrEmpty() -> displayAuthPanel()
             settings.pluginFirstRun -> {
-                if (getAmplitudeExperimentService(project)?.isPartOfExperimentalWelcomeWorkflow() == true) {
-                    pluginSettings().pluginFirstRun = false
-                    enableProductsAccordingToServerSetting()
-                    getSyncPublisher(project, SnykSettingsListener.SNYK_SETTINGS_TOPIC)?.settingsChanged()
-                    displayTreeAndDescriptionPanels()
-                    triggerScan()
-                } else {
-                    displayPluginFirstRunPanel()
```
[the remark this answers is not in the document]
bastiandoetsch: Shouldn't ProductSelectionIsViewed be sent when opening the preference pane? But yeah, for welcome flow it is unused.

bastiandoetsch:
> `SnykToolWindowPanel#enableProductsAccordingToServerSetting` override **incorrectly** the default enablement for products. Correct enablement (confirmed by Georgi) is presented in SnykApplicationSettingsStateService

We should double-check then - it is enabled as it was decided by Georgi as well, but that's easy to change.

bastiandoetsch:
> Showing Tree before getting token we have Code Security node in wrong state: image

Isn't it Code Quality that is in wrong state? We don't know anything about an organization before log in. @deepcodeg

src/main/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanel.kt
```diff
@@ -592,28 +581,13 @@ class SnykToolWindowPanel(val project: Project) : JPanel(), Disposable {
         )
     }
 
-    fun displayPluginFirstRunPanel() {
-        removeAll()
-        val onboardPanel = OnboardPanel(project)
```
ArtsiomCh, line 597: OnboardPanel class seems to be unused anymore and could be deleted

src/main/kotlin/io/snyk/plugin/Utils.kt
```diff
@@ -198,9 +198,10 @@ fun controlExternalProcessWithProgressIndicator(
     checkCancelled.invoke()
 }
 
-fun isIacEnabled(): Boolean = Registry.`is`("snyk.preview.iac.enabled", false)
+fun isIacEnabled(): Boolean = Registry.`is`("snyk.preview.iac.enabled", pluginSettings().iacScanEnabled)
```
ArtsiomCh, line 201: we can't do that.
isIacEnabled and iacScanEnabled serve different things. We should not mix them up.
PS for example isIacEnabled  serve to detect if IaC should be shown and if "snyk.preview.iac.enabled" is absent for any reason - IaC will be shown immediately to the GA (as iacScanEnabled  is `true` by default).

src/main/kotlin/io/snyk/plugin/Utils.kt
```diff
@@ -198,9 +198,10 @@ fun controlExternalProcessWithProgressIndicator(
...
 
-fun isContainerEnabled(): Boolean = Registry.`is`("snyk.preview.container.enabled", false)
+fun isContainerEnabled(): Boolean =
+    Registry.`is`("snyk.preview.container.enabled", pluginSettings().containerScanEnabled)
```
ArtsiomCh, line 204: see comment above

src/main/kotlin/io/snyk/plugin/Utils.kt
bastiandoetsch: Which is the desired behaviour - it should be automatically enabled, if no feature flag is present. We deliver the feature-flag with a default value of false in our plugin.xml, so not having it right now should not be an issue. But if/when it's gone, IaC/Container should be enabled by default.

ArtsiomCh (approved):
LGTM

PR #274, opened 2022-01-21
fix: 🐛 don't delete / replace CLI file before download of new version is successful [ROAD-645]

bastiandoetsch:
A client does not allow installation of executables/download into the plugin directory. Thus, we should not assume that a download of the CLI file can be successful, and only delete the executable after successfully saving the newly downloaded file and after having verified the checksum.


bastiandoetsch:
Oh, I see test-flakiness. I'll investigate.

bastiandoetsch:
@ArtsiomCh it would be interesting to run the tests on windows, as for windows it is using a different `sleep` command, namely `timeout`. See commit  `bcc8745`

michelkaporin (approved):
Reviewed together on a call. LGTM 👍

PR #275, opened 2022-01-24
fix: 🐛 If installation of IntelliJ is on different drive than the temp dir, moving the CLI after download fails

bastiandoetsch:


ArtsiomCh (approved):
Thanks for quick fix!

PR #278, opened 2022-01-24
Allow manual triggering of weekly release workflow

bastiandoetsch:


.github/workflows/weekly-release.yml
```diff
@@ -0,0 +1,42 @@
...
+        if: ${{ !env.ACT }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          git checkout master
+          ./gradlew release
+
+      - name: Release - Dry Run
+        if: ${{ env.ACT }}
+        run: ./gradlew release -Prelease.dryRun
```
michelkaporin: Read this as `-Please.dryRun` 😆

.github/workflows/weekly-release.yml
```diff
@@ -1,6 +1,7 @@
 name: Weekly Release
 
 on:
+  workflow_dispatch:
```
michelkaporin, line 4: Cool! Perhaps we also could have a version number input, in case we decide to release new minor version instead of operating over patch versioning?

michelkaporin (approved):

.github/workflows/weekly-release.yml
bastiandoetsch: That's a good idea, but more complicated... Let's do that, but apart.

bastiandoetsch: I'll create a ticket

.github/workflows/weekly-release.yml
bastiandoetsch: HAHAHA

PR #284, opened 2022-01-25
chore: enable plugin publishing in release workflow

bastiandoetsch:
Enable automatic plugin publishing in release workflow.


michelkaporin:
Let's have tests in there as a gate for a release.

michelkaporin (approved):

PR #285, opened 2022-01-25
chore: add tests as step before releasing

bastiandoetsch:


.github/workflows/release.yml
```diff
@@ -29,6 +29,12 @@ jobs:
           key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle') }}
           restore-keys: ${{ runner.os }}-gradle
 
+      - name: Test
+        run: |
+          ./gradlew test integTest verifyPlugin clean
```
ArtsiomCh, line 34: why only IntegTest and not unit tests also?

ArtsiomCh (approved):
LGTM

PR #292, opened 2022-01-27
fix: caches update and cleaning [ROAD-646]

ArtsiomCh:
For our caches we need following events for file: Create, Change, Delete                               
Also in our caches we need next type of actions: _add_ _update_ _clean_                                
Note: Current implementation meaning:                                                                  
 add - mark results as invalid anymore (i.e. next scan attempt should run scan and not use cache)      
 update - mark item(file) result as obsolete (i.e. next scan attempt should run scan and not use cache)
 clean - mark item(file) result as obsolete (i.e. next scan attempt should run scan and not use cache) 
                                                                                                       
BulkFileListener provide next events: Create, ContentChange, Move, Copy, Delete                        
Also in `before` state we have access to old files.                                                    
While in `after` state we have access for new/updated files too (but old might not exist anymore)      
                                                                                                       
Next mapping/interpretation for BulkFileListener type of events should be used:                        
* Create                                                                                                 
   - addressed at `after` state, new file processed to _add_ caches                                      
* ContentChange                                                                                          
   - addressed at `after` state, new file processed to _update_ caches                                   
* Move                                                                                                   
   - addressed at `before` state, old file processed to _clean_ caches                                   
   - addressed at `after` state, new file processed to _add_ caches                                      
* Copy                                                                                                   
   - addressed at `after` state, new file processed to _add_ caches                                      
* Delete                                                                                                 
   - addressed at `before` state, old file processed to _clean_ caches

### DIFF
--- a/src/integTest/kotlin/io/snyk/SnykBulkFileListenerTest.kt
+++ b/src/integTest/kotlin/io/snyk/SnykBulkFileListenerTest.kt
@@ -27,7 +27,6 @@ import snyk.oss.OssResult
 import java.io.File
 import java.nio.file.Files
 import java.nio.file.Paths
-import java.util.concurrent.Callable
 import java.util.concurrent.TimeUnit
 
 @Suppress("FunctionName")
@@ -101,6 +100,24 @@ class SnykBulkFileListenerTest : BasePlatformTestCase() {
     fun testCurrentIacResults_shouldDropCachedResult_whenIacSupportedFileChanged() {
         val file = "k8s-deployment.yaml"
         val filePath = "/src/$file"
+        createFakeIacResultInCache(file, filePath)
+
+        myFixture.configureByText(file, "some text")
+
+        await().atMost(2, TimeUnit.SECONDS).until { cacheUpdated(filePath) }
+    }
+
+    private fun cacheUpdated(filePath: String?): Boolean {
+        val iacCachedIssues = project.service<SnykToolWindowPanel>().currentIacResult!!.allCliIssues!!
+        return iacCachedIssues.any { iacFile ->
+            (filePath == null || iacFile.targetFilePath == filePath) && iacFile.obsolete
+        }
+    }
+
+    private fun isIacUpdateNeeded(): Boolean =
+        project.service<SnykToolWindowPanel>().currentIacResult?.iacScanNeeded ?: true
+
+    private fun createFakeIacResultInCache(file: String, filePath: String) {
         val iacIssuesForFile = IacIssuesForFile(emptyList(), file, filePath, "npm")
         val iacVulnerabilities = listOf(iacIssuesForFile)
         val fakeIacResult = IacResult(iacVulnerabilities, null)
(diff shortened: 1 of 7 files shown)


ArtsiomCh:
@pavel-snyk  @bastiandoetsch please have a look for any conceptual objections/corrections
PS tests will be added

src/main/kotlin/io/snyk/plugin/SnykBulkFileListener.kt
```diff
@@ -123,43 +145,53 @@ class SnykBulkFileListener : BulkFileListener {
...
             }
+
+        if (changed) {
+            val newIacCache = IacResult(newIacFileList.toImmutableList(), null)
+            toolWindowPanel.currentIacResult = newIacCache
+            toolWindowPanel.iacScanNeeded = true
+            ApplicationManager.getApplication().invokeLater {
+                toolWindowPanel.displayIacResults(newIacCache)
+            }
+            DaemonCodeAnalyzer.getInstance(toolWindowPanel.project).restart()
```
bastiandoetsch, line 194: Why is this outside the invokeLater?

bastiandoetsch (commented):
How can we test the changed functionality? It feels to me, this logic should have additional tests in `SnykBulkFileListenerIntegTest`

src/main/kotlin/io/snyk/plugin/SnykBulkFileListener.kt
ArtsiomCh: displayIacResults require EDT while inside SnykBulkFileListener we seems to be in some background thread.

src/integTest/kotlin/io/snyk/SnykBulkFileListenerTest.kt
```diff
@@ -101,27 +100,95 @@ class SnykBulkFileListenerTest : BasePlatformTestCase() {
...
         val file = "k8s-deployment.yaml"
         val filePath = "/src/$file"
+        createFakeIacResultInCache(file, filePath)
+
+        myFixture.configureByText(file, "some text")
+
+        await().atMost(2, TimeUnit.SECONDS).until { cacheUpdated(filePath) }
+    }
+
+    private fun cacheUpdated(filePath: String?): Boolean {
```
bastiandoetsch, line 110: minor: I would name this cacheInvalidatedForFilePath

src/integTest/kotlin/io/snyk/SnykBulkFileListenerTest.kt
```diff
@@ -101,27 +100,95 @@ class SnykBulkFileListenerTest : BasePlatformTestCase() {
...
+        val iacCachedIssues = project.service<SnykToolWindowPanel>().currentIacResult!!.allCliIssues!!
+        return iacCachedIssues.any { iacFile ->
+            (filePath == null || iacFile.targetFilePath == filePath) && iacFile.obsolete
```
bastiandoetsch, line 113: What is the purpose to check here for filePath == null? Should we ever have entries in iacCachedIssues when the filePath == null? In which scenario would that be a valid cache entry?

src/integTest/kotlin/io/snyk/SnykBulkFileListenerTest.kt
```diff
@@ -101,27 +100,95 @@ class SnykBulkFileListenerTest : BasePlatformTestCase() {
...
+        val existingFile = "existing.yaml"
+        createFakeIacResultInCache(existingFile, "/src/$existingFile")
+
+        assertFalse(isIacUpdateNeeded())
+
+        myFixture.addFileToProject("new.yaml", "some text")
 
-        await().atMost(2, TimeUnit.SECONDS).until(cacheUpdated(toolWindowPanel, filePath))
+        assertTrue(isIacUpdateNeeded())
+        assertFalse(cacheUpdated(null))
```
bastiandoetsch, line 140: I dont' understand what is asserted here :(

bastiandoetsch (approved):
Looks good to me, just some questions. But approved.

src/integTest/kotlin/io/snyk/SnykBulkFileListenerTest.kt
ArtsiomCh: filePath is method param we check against. 
`filePath == null` is the case when we want to check if _any_ IaC file with issues been marked as obsolete
adding kdoc to make it clear for the future (me) reader.

src/integTest/kotlin/io/snyk/SnykBulkFileListenerTest.kt
ArtsiomCh: will add message to clarify
otherwise I probably will not understand it either a month later

PR #294, opened 2022-02-02
fix some annoying exceptions

ArtsiomCh:
fix: [InvocationTargetException ](https://sentry.io/organizations/snyk/issues/2980659540/?project=5956531&referrer=slack)on project-less authentication;
fix: memory leak in Authentication panel

### DIFF
--- a/src/main/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanel.kt
+++ b/src/main/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanel.kt
@@ -11,6 +11,7 @@ import com.intellij.openapi.fileEditor.FileDocumentManager
 import com.intellij.openapi.fileEditor.FileEditorManager
 import com.intellij.openapi.project.Project
 import com.intellij.openapi.ui.SimpleToolWindowPanel
+import com.intellij.openapi.util.Disposer
 import com.intellij.openapi.vfs.VirtualFile
 import com.intellij.openapi.vfs.VirtualFileManager
 import com.intellij.psi.PsiDocumentManager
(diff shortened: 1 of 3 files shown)


src/main/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanel.kt
```diff
@@ -594,9 +595,9 @@ class SnykToolWindowPanel(val project: Project) : JPanel(), Disposable {
             settings.pluginFirstRun -> {
                 pluginSettings().pluginFirstRun = false
                 enableProductsAccordingToServerSetting()
-                getSyncPublisher(project, SnykSettingsListener.SNYK_SETTINGS_TOPIC)?.settingsChanged()
                 displayTreeAndDescriptionPanels()
-                triggerScan()
+                // don't trigger scan for Default project i.e. no project opened state
+                if (project.basePath != null) triggerScan()
```
michelkaporin, line 600: It would be great to have this tested, if it's possible.

michelkaporin (approved):

src/main/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanel.kt
ArtsiomCh: I'm afraid writing tests for that will take 10 times longer then actual fix :(
That might be luck of my testing experience... but here particularly I would either drop the PR then wasting time for writing tests...

bastiandoetsch: Agree, a test would be nice. Would probably involve a few mocks, so yeah, I understand the reluctance. Still, would be great :)

michelkaporin: @ArtsiomCh I feel you here. 

@bastiandoetsch I think we should invest more time into our architecture to enable unit testing of such small bits.

bastiandoetsch: For sure. SnykToolWindowPanel needs a thorough decomposition, it is way too complex and needs to be split-up.

PR #297, opened 2022-02-04
chore: 🔧 update CI to test on macOS, Linux and Windows

bastiandoetsch:


michelkaporin (approved):
💪💪 LGTM. I am not sure why do you have another "Test" check on top of the matrix ones but perhaps it's not much relevant and just visually confusing 😅

PR #298, opened 2022-02-04
feat: ✨ use http proxy host and proxy port settings from IntelliJ proxy configuration

bastiandoetsch:
Takes configuration from IntelliJ http proxy settings and uses it for CLI download, API calls, scans and authentication.

image

Can be verified using the VM:

https://github.com/bastiandoetsch/jvm-development-environment

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,19 +17,19 @@ fix advice:
 
 **Snyk detects the critical vulnerability Log4Shell, which was found in the open source Java library log4j-core - a component of one of the most popular Java logging frameworks, Log4J. The vulnerability was categorized as Critical with a CVSS score of 10, and with a mature exploit level.**
 
-### Proxy Setup
-If you are a behind a proxy, please configure the proxy in the IDE. Currently, http and https proxies are supported by the plugin.
-
-### Environment setup
-The plugin uses the Snyk CLI to perform vulnerability scans. In order for this to function correctly, 
-certain environment variables need to be set.
-
-1. `JAVA_HOME` to analyse Java JVM-based projects via Snyk CLI
-2. `PATH` to find maven when analysing Maven projects, to find python for python projects, etc
-
 ### Useful links
 
 - This plugin works with projects written in Java, JavaScript, .NET and many more languages. See the [full list of languages and package managers Snyk supports](https://snyk.co/ucWSd)
 - [Bug tracker](https://github.com/snyk/snyk-intellij-plugin/issues)
 
 <!-- Plugin description end -->
+
+### Proxy Setup
+If you are a behind a proxy, please configure the proxy in the IDE. Currently, http and https proxies are supported by the plugin.
+
+### Environment setup
+The plugin uses the Snyk CLI to perform vulnerability scans. In order for this to function correctly,
+certain environment variables need to be set.
+
+1. `JAVA_HOME` to analyse Java JVM-based projects via Snyk CLI
+2. `PATH` to find maven when analysing Maven projects, to find python for python projects, etc
--- a/README.md
+++ b/README.md
@@ -17,6 +17,16 @@ fix advice:
 
 **Snyk detects the critical vulnerability Log4Shell, which was found in the open source Java library log4j-core - a component of one of the most popular Java logging frameworks, Log4J. The vulnerability was categorized as Critical with a CVSS score of 10, and with a mature exploit level.**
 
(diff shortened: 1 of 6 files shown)


src/main/kotlin/io/snyk/plugin/cli/ConsoleCommandRunner.kt
```diff
@@ -98,6 +99,10 @@ open class ConsoleCommandRunner {
         commandLine.environment["SNYK_INTEGRATION_VERSION"] = pluginInfo.integrationVersion
         commandLine.environment["SNYK_INTEGRATION_ENVIRONMENT"] = pluginInfo.integrationEnvironment
         commandLine.environment["SNYK_INTEGRATION_ENVIRONMENT_VERSION"] = pluginInfo.integrationEnvironmentVersion
+        val proxySettings = HttpConfigurable.getInstance()
+        if (proxySettings != null && proxySettings.USE_HTTP_PROXY && proxySettings.PROXY_HOST.isNotEmpty()) {
+            commandLine.environment["https_proxy"] = "http://${proxySettings.PROXY_HOST}:${proxySettings.PROXY_PORT}"
```
michelkaporin, line 104: Why `http` hardcoded? What about respecting auth parameters?

michelkaporin:
Will querying for `cli-config/settings/sast`, as well as Advisor pick this up as well?

bastiandoetsch:
> Will querying for `cli-config/settings/sast`, as well as Advisor pick this up as well?

Surprisingly, it does @michelkaporin . I was expecting having to change something, but it wasn't necessary. Tested with the VM.

src/main/kotlin/io/snyk/plugin/cli/ConsoleCommandRunner.kt
bastiandoetsch: `http` is somehow needed by the CLI - I don't know why they expect a hard-coded http even for https. Try it out, you'll see an exception. About authentication: yes, it should be part of the hostname, I agree.

michelkaporin:
>Surprisingly, it does @michelkaporin . I was expecting having to change something, but it wasn't necessary. Tested with the VM.

@bastiandoetsch It would be great to make it explainable to understand what's the resolution route that it takes there 🤔

bastiandoetsch:
@michelkaporin the only reason I can think of is that IntelliJ automatically sets the configured proxy as the system-wide proxy in the JVM Security Manager. OKHTTP, which is used under the hood of Advisor & the Sast call, automatically reads it.

README.md
```diff
@@ -17,6 +17,16 @@ fix advice:
 
 **Snyk detects the critical vulnerability Log4Shell, which was found in the open source Java library log4j-core - a component of one of the most popular Java logging frameworks, Log4J. The vulnerability was categorized as Critical with a CVSS score of 10, and with a mature exploit level.**
 
+### Proxy Setup
```
pavel-snyk, line 20: @bastiandoetsch , let's move proxy and environment setup block it behind of `<!-- Plugin description end -->` tag, so it will not be generated to marketplace description.

bastiandoetsch:
Yep, it seems like IDEA is using `ProxySelector.getDefault()` . See `com.intellij.util.proxy.CommonProxy`.

src/main/kotlin/io/snyk/plugin/services/download/CliDownloader.kt
```diff
@@ -80,7 +80,7 @@ class CliDownloader {
     }
 
     fun expectedSha(): String {
-        val shaUrl = URL(String.format(SHA256_DOWNLOAD_URL, Platform.current().snykWrapperFileName))
-        return shaUrl.readText(Charsets.UTF_8).split(" ")[0]
+        val url = String.format(SHA256_DOWNLOAD_URL, Platform.current().snykWrapperFileName)
+        return HttpRequests.request(url).forceHttps(true).readString().split(" ")[0]
```
pavel-snyk, line 84: @bastiandoetsch , why `forceHttps(true)` now?

bastiandoetsch: Because we don't want any possibility to have the SHA-256 integrity compromised. SSL ensures that.

pavel-snyk (approved):
LGTM.

PR #310, opened 2022-02-16
fix: use environment variable when disabling usage analytics [ROAD-741]

pavel-snyk:
This PR use CLI environment variable (`SNYK_CFG_DISABLE_ANALYTICS`) instead of CLI flag. The flag doesn't work for IaC and Container.


bastiandoetsch (approved):
LGTM

PR #311, opened 2022-02-16
feat: ✨track quickfixes [ROAD-688]

bastiandoetsch:


src/main/kotlin/io/snyk/plugin/services/SnykAnalyticsService.kt
```diff
@@ -145,6 +147,22 @@ class SnykAnalyticsService : Disposable {
         }
     }
 
+    fun logQuickFixIsDisplayed(event: QuickFixIsDisplayed) {
+        if (!settings.usageAnalyticsEnabled) return
```
michelkaporin, line 151: Is `userId.isBlank()` check needed here or userId is always initialised to that point?

src/main/kotlin/snyk/common/intentionactions/AlwaysAvailableReplacementIntentionAction.kt
```diff
@@ -7,20 +7,31 @@ import com.intellij.openapi.project.Project
 import com.intellij.openapi.util.TextRange
 import com.intellij.psi.PsiFile
 import com.intellij.util.DocumentUtil
+import io.snyk.plugin.getSnykAnalyticsService
```
michelkaporin, line 10: Let's fix warnings

michelkaporin (approved):
Great, left few small comments 👍

src/main/kotlin/io/snyk/plugin/services/SnykAnalyticsService.kt
bastiandoetsch: It is already initialized - you can't get a quickfix without results, and no results without token. And once we get the token, the userID is updated.

src/main/kotlin/snyk/common/intentionactions/AlwaysAvailableReplacementIntentionAction.kt
bastiandoetsch: I wonder how that snyked in.

bastiandoetsch:
Rebased against master, added changelog, and force-pushed.

src/main/kotlin/io/snyk/plugin/Utils.kt
```diff
@@ -67,6 +68,8 @@ fun getSnykCliDownloaderService(project: Project): SnykCliDownloaderService? = p
 
 fun getSnykProjectSettingsService(project: Project): SnykProjectSettingsStateService? = project.serviceIfNotDisposed()
 
+fun getSnykAnalyticsService(project: Project): SnykAnalyticsService? = project.serviceIfNotDisposed()
```
ArtsiomCh, line 71: don't needed
SnykAnalyticsService is not project service

src/main/kotlin/snyk/common/intentionactions/AlwaysAvailableReplacementIntentionAction.kt
```diff
@@ -8,19 +8,28 @@ import com.intellij.openapi.util.TextRange
...
+import snyk.analytics.QuickFixIsTriggered
 
 class AlwaysAvailableReplacementIntentionAction(
     val range: TextRange,
     val replacementText: String,
     private val intentionText: String = intentionDefaultTextPrefix,
     private val familyName: String = intentionDefaultFamilyName,
-    val message: String = ""
+    val message: String = "",
+    val analyticsService: SnykAnalyticsService?
```
ArtsiomCh, line 22: could have default value and reassigned just for tests

ArtsiomCh:
Shell we log quickfixes from Container product?

src/main/kotlin/snyk/common/intentionactions/AlwaysAvailableReplacementIntentionAction.kt
bastiandoetsch: I like that!

bastiandoetsch:
> @bastiandoetsch Please check when it's triggered I can see quickFixIsDisplayed triggered twice even when only 1 popup is shown Also after try to applying it (for package-lock case it doesn't go through), quickFixIsDisplayed triggers multiple times.

Yeah, we cannot control, how often getText() is triggered, or the hover is activated afaik. Do you have any idea?

src/main/kotlin/io/snyk/plugin/services/SnykAnalyticsService.kt
```diff
@@ -145,6 +147,22 @@ class SnykAnalyticsService : Disposable {
...
+
+        catchAll(log, "quickFixIsDisplayed") {
+            itly.logQuickFixIsDisplayed(userId, event)
+        }
+    }
+
+    fun logQuickFixIsTriggered(event: QuickFixIsTriggered) {
+        if (!settings.usageAnalyticsEnabled) return
+
+        catchAll(log, "quickFixIsDisplayed") {
```
ArtsiomCh, line 161: should be `logQuickFixIsTriggered`
PS was confused not to see any logQuickFixIsTriggered in the logs...

ArtsiomCh:
> Yeah, we cannot control, how often getText() is triggered, or the hover is activated afaik. Do you have any idea?
Looks like the only way will be implementing some kind of hacky flag: if it been requested within ~500 ms timeframe then send only one event.
Otherwise we need confirm with Georgy he is ok to have many false events (I don't think so as it ruined the whole idea of such analytic)

src/main/kotlin/io/snyk/plugin/services/SnykAnalyticsService.kt
bastiandoetsch: Damn copy & paste ;)

bastiandoetsch:
> > Yeah, we cannot control, how often getText() is triggered, or the hover is activated afaik. Do you have any idea?
> 
> Looks like the only way will be implementing some kind of hacky flag: if it been requested within ~500 ms timeframe then send only one event. Otherwise we need confirm with Georgy he is ok to have many false events (I don't think so as it ruined the whole idea of such analytic)

I'd rather not log than using such a hack that can never be precise. @deepcodeg, wdyt?

bastiandoetsch:
Force-push to rebase against master

bastiandoetsch:
If we don't want the `QuickFix is Displayed` event, we can always remove it, according to what Georgi wants. I'll merge this branch now in :).

PR #314, opened 2022-02-18
chore: improve on Snyk Code Annotator [ROAD-668] [ROAD-737]

bastiandoetsch:
Fixes from PR comments.

### DIFF
--- a/src/main/kotlin/snyk/code/annotator/SnykCodeAnnotator.kt
+++ b/src/main/kotlin/snyk/code/annotator/SnykCodeAnnotator.kt
@@ -5,15 +5,16 @@ import ai.deepcode.javaclient.core.SuggestionForFile
 import com.intellij.lang.annotation.AnnotationHolder
 import com.intellij.lang.annotation.ExternalAnnotator
 import com.intellij.lang.annotation.HighlightSeverity
+import com.intellij.openapi.diagnostic.logger
 import com.intellij.openapi.progress.ProgressManager
 import com.intellij.openapi.util.TextRange
 import com.intellij.psi.PsiFile
 import io.snyk.plugin.snykcode.core.AnalysisData
 import snyk.common.AnnotatorCommon
-import kotlin.math.max
 import kotlin.math.min
 
 class SnykCodeAnnotator : ExternalAnnotator<PsiFile, Unit>() {
+    val logger = logger<SnykCodeAnnotator>()
     // overrides needed for the Annotator to invoke apply(). We don't do anything here
     override fun collectInformation(file: PsiFile): PsiFile = file
     override fun doAnnotate(psiFile: PsiFile?) {
@@ -28,9 +29,7 @@ class SnykCodeAnnotator : ExternalAnnotator<PsiFile, Unit>() {
             suggestionForFile.ranges.forEach {
                 val textRange = textRange(psiFile, it)
                 if (!textRange.isEmpty) {
-                    val annotationBuilder =
-                        holder.newAnnotation(severity, annotationMessage(suggestionForFile)).range(textRange)
-                    annotationBuilder.create()
+                    holder.newAnnotation(severity, annotationMessage(suggestionForFile)).range(textRange).create()
                 }
             }
         }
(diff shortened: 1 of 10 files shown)


src/main/kotlin/snyk/code/annotator/SnykCodeAnnotator.kt
```diff
@@ -57,17 +56,32 @@
...
 
-        val lineOffSet = min(document.getLineStartOffset(startRow) + startCol, document.textLength - 1)
-        val lineOffSetEnd = max(lineOffSet, min(document.getLineStartOffset(endRow) + endCol, document.textLength - 1))
-        return TextRange.create(lineOffSet, lineOffSetEnd)
+            if (startRow < 0 || startRow > document.lineCount - 1)
+                throw IllegalArgumentException("Invalid range $snykCodeRange")
+            if (endRow < 0 || endRow > document.lineCount - 1 || endRow < startRow)
+                throw IllegalArgumentException("Invalid range $snykCodeRange")
+
+            val lineOffSet = min(document.getLineStartOffset(startRow) + startCol, document.textLength - 1)
```
ArtsiomCh, line 73: let's remove min() in lines above as we have check `lineOffSet > document.textLength - 1` and `lineOffSetEnd > document.textLength - 1` later on
otherwise those checks are useless

src/main/kotlin/snyk/container/YAMLImageExtractor.kt
```diff
@@ -39,7 +39,7 @@ object YAMLImageExtractor {
     fun extractFromFile(file: VirtualFile, project: Project): Set<KubernetesWorkloadImage> {
         val psiFile = findPsiFileIgnoringExceptions(file, project)
         if (psiFile == null || file.isDirectory || !file.isValid) {
-            logger.warn("no psi file found for ${file.path}")
+            logger.debug("no psi file found for ${file.path}")
```
ArtsiomCh, line 42: why do we need that event in log?
again, it will (and is) bloating logs every time when file is directory (happens thousand times on even middle-size project)
the only logging (might) have sense in case of `== nul` or `!isValid`

src/main/kotlin/snyk/container/YAMLImageExtractor.kt
```diff
@@ -39,7 +39,7 @@ object YAMLImageExtractor {
     fun extractFromFile(file: VirtualFile, project: Project): Set<KubernetesWorkloadImage> {
         val psiFile = findPsiFileIgnoringExceptions(file, project)
         if (psiFile == null || file.isDirectory || !file.isValid) {
```
ArtsiomCh, line 41: `file.isDirectory || !file.isValid` should be done before obtaining PsiFile and if true - such obtaining not needed

ArtsiomCh (changes requested):

ArtsiomCh:
Overall looks good. Just some small changes are needed.

ArtsiomCh:
@bastiandoetsch apparently you did fix for ROAD-737 here ;)

src/main/kotlin/snyk/code/annotator/SnykCodeAnnotator.kt
bastiandoetsch: No, if the startCol / endCol is arbitrary high, this could still be out of bounds.

ArtsiomCh: correct, and if we do that `min()` here we'll never reach `lineOffSet > document.textLength - 1` inside `if` below and will show `lineOffSetEnd=document.textLength - 1` instead of properly proceed `throw IllegalArgumentException("Invalid range $snykCodeRange")`

bastiandoetsch: Yeah, it's true, we should throw an exception instead of using the minimum. I'll change that.

bastiandoetsch: Also, I don't know which version/branch I viewed. It makes complete sense and my comment does not make sense.

ArtsiomCh (approved):
LGTM

PR #315, opened 2022-02-21
feat: block LCE upload for Snyk Code [ROAD-687]

pavel-snyk:
This PR adds a new feature for Snyk Code => disable it if local code engine is enabled.

Configuration screen:
Screen Shot 2022-02-22 at 04 00 56

Tree:
Screen Shot 2022-02-22 at 04 01 16

### DIFF
--- a/src/integTest/kotlin/io/snyk/plugin/services/SnykTaskQueueServiceTest.kt
+++ b/src/integTest/kotlin/io/snyk/plugin/services/SnykTaskQueueServiceTest.kt
@@ -24,6 +24,8 @@ import io.snyk.plugin.services.download.SnykCliDownloaderService
 import io.snyk.plugin.setupDummyCliFile
 import io.snyk.plugin.ui.toolwindow.SnykToolWindowPanel
 import org.awaitility.Awaitility.await
+import org.hamcrest.CoreMatchers.equalTo
+import org.hamcrest.MatcherAssert.assertThat
 import org.junit.Test
 import snyk.container.ContainerResult
 import snyk.iac.IacResult
@@ -35,6 +37,7 @@ class SnykTaskQueueServiceTest : LightPlatformTestCase() {
         super.setUp()
         unmockkAll()
         resetSettings(project)
+
         project.replaceService(SnykToolWindowPanel::class.java, SnykToolWindowPanel(project), project)
     }
 
@@ -118,6 +121,20 @@ class SnykTaskQueueServiceTest : LightPlatformTestCase() {
         assertNull(settings.sastOnServerEnabled)
         assertFalse(settings.snykCodeSecurityIssuesScanEnable)
         assertFalse(settings.snykCodeQualityIssuesScanEnable)
+    }
+
+    @Test
+    fun `test should disable Code settings when LCE is enabled`() {
+        val snykTaskQueueService = project.service<SnykTaskQueueService>()
+        val settings = pluginSettings()
+        settings.sastOnServerEnabled = true
+        settings.localCodeEngineEnabled = true
+
+        snykTaskQueueService.scan()
+        Disposer.dispose(service<SnykApiService>())
+
+        assertThat(settings.snykCodeSecurityIssuesScanEnable, equalTo(false))
+        assertThat(settings.snykCodeQualityIssuesScanEnable, equalTo(false))
     }
 
     @Test
--- a/src/main/kotlin/io/snyk/plugin/net/SnykApiClient.kt
+++ b/src/main/kotlin/io/snyk/plugin/net/SnykApiClient.kt
@@ -13,6 +13,9 @@ class SnykApiClient constructor(
     val sastOnServerEnabled: Boolean?
         get() = executeRequest(CliConfigService.apiName, cliConfigService().sast())?.sastEnabled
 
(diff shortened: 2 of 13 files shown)


bastiandoetsch:
Hey @pavel-snyk , had a sneak peek at your PR, and while I know that it's still in draft, I think it could be good to comment now, as I see that you're doing two calls to the API, when one call would be sufficient. I think the "normal" get() should just set the settings for local code, thus avoiding doubling the load on the API.

src/main/kotlin/io/snyk/plugin/ui/UIUtils.kt
```diff
@@ -81,6 +81,8 @@ fun insertTitleAndResizableTextIntoPanelColumns(
 
 fun snykCodeAvailabilityPostfix(): String = when {
     !isSnykCodeAvailable(pluginSettings().customEndpointUrl) -> " (disabled for endpoint)"
+    (pluginSettings().sastOnServerEnabled ?: true && pluginSettings().localCodeEngineEnabled ?: true) ->
+        " (disabled due to local code engine)"
```
bastiandoetsch, line 85: This should be `Local Code Engine` (capitalization)

src/integTest/kotlin/io/snyk/plugin/services/SnykTaskQueueServiceTest.kt
```diff
@@ -120,6 +123,20 @@ class SnykTaskQueueServiceTest : LightPlatformTestCase() {
         assertFalse(settings.snykCodeQualityIssuesScanEnable)
     }
 
+    @Test
+    fun `test should disable Code settings when LCE is enabled`() {
+        val snykTaskQueueService = project.service<SnykTaskQueueService>()
+        val settings = pluginSettings()
+        settings.sastOnServerEnabled = true
```
bastiandoetsch, line 130: This may affect other tests - what do you think? Should we reset the settings to the state before at the end of the test?

src/main/kotlin/io/snyk/plugin/services/SnykTaskQueueService.kt
```diff
@@ -156,11 +156,24 @@ class SnykTaskQueueService(val project: Project) {
...
             override fun run(indicator: ProgressIndicator) {
-                settings.sastOnServerEnabled = service<SnykApiService>().sastOnServerEnabled
+                val sastCliConfigSettings = service<SnykApiService>().sastSettings
+                settings.sastOnServerEnabled = sastCliConfigSettings?.sastEnabled
+                settings.localCodeEngineEnabled = sastCliConfigSettings?.localCodeEngine?.enabled
                 when (settings.sastOnServerEnabled) {
                     true -> {
-                        getSnykCode(project)?.scan()
-                        scanPublisher?.scanningStarted()
+                        if (settings.localCodeEngineEnabled == true) {
```
bastiandoetsch, line 164: Why not an if then else? (minor)

src/main/kotlin/io/snyk/plugin/net/SnykApiClient.kt
```diff
@@ -10,8 +10,8 @@ import retrofit2.Retrofit
 class SnykApiClient constructor(
     private val retrofit: Retrofit
 ) {
-    val sastOnServerEnabled: Boolean?
-        get() = executeRequest(CliConfigService.apiName, cliConfigService().sast())?.sastEnabled
+    val sastSettings: CliConfigSettings?
+        get() = executeRequest(CliConfigService.apiName, cliConfigService().sast())
```
bastiandoetsch, line 14: Another thing that is missing (but we already have a ticket for this), is the transmission of the organisation as a query parameter.

bastiandoetsch (approved):
Only minor stuff. Looks good!

src/integTest/kotlin/io/snyk/plugin/services/SnykTaskQueueServiceTest.kt
pavel-snyk: Clarified, settings will be reset.

src/main/kotlin/io/snyk/plugin/services/SnykTaskQueueService.kt
pavel-snyk: unfortunately `null` has a semantic value -> failed http request.

src/main/kotlin/io/snyk/plugin/net/SnykApiClient.kt
pavel-snyk: agreed, let's do it in another ticket.

src/main/kotlin/io/snyk/plugin/ui/UIUtils.kt
pavel-snyk: 👍, fixed with f81b3396a7de38ace4ab5f5f57f886b714dd0f4b

ArtsiomCh:
After plugin update all our existing users will have Code "disabled due LCE" as `var localCodeEngineEnabled: Boolean? = null` by default
image
So they will have to go to settings to manually re-anable it. 
Is that desired behavior?

michelkaporin:
@pavel-snyk also changelog had to be updated here perhaps?

PR #318, opened 2022-02-22
chore: remove quickfix is displayed event

bastiandoetsch:


ArtsiomCh (approved):
LGTM
Let not merge any not confirmed code into master in the future to avoid such last minute fixes.

PR #328, opened 2022-03-04
feat: sast settings with organization as url query param [ROAD-734]

pavel-snyk:
This PR adds a feature to check SAST API endpoint with organization from settings (if configured).
Default value for org Retrofit query param is `null`. In this case Retrofit will omit URL query param.

### DIFF
--- a/src/main/kotlin/io/snyk/plugin/services/SnykApiService.kt
+++ b/src/main/kotlin/io/snyk/plugin/services/SnykApiService.kt
@@ -23,7 +23,10 @@ import javax.net.ssl.X509TrustManager
 class SnykApiService : Disposable {
 
     val sastSettings: CliConfigSettings?
-        get() = snykApiClient?.sastSettings
+        get() {
+            val pluginSettings = service<SnykApplicationSettingsStateService>()
+            return snykApiClient?.sastSettings(pluginSettings.organization)
+        }
 
     val userId: String?
         get() = snykApiClient?.userId
--- a/src/test/kotlin/io/snyk/plugin/net/CliConfigServiceTest.kt
+++ b/src/test/kotlin/io/snyk/plugin/net/CliConfigServiceTest.kt
@@ -1,0 +1,71 @@
+package io.snyk.plugin.net
+
+import okhttp3.mockwebserver.MockResponse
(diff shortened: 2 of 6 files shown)


src/main/kotlin/io/snyk/plugin/services/SnykApiService.kt
```diff
@@ -23,7 +23,10 @@ import javax.net.ssl.X509TrustManager
 class SnykApiService : Disposable {
 
     val sastSettings: CliConfigSettings?
-        get() = snykApiClient?.sastSettings
+        get() {
+            val pluginSettings = service<SnykApplicationSettingsStateService>()
```
bastiandoetsch, line 27: Any reason to not make that a field / property?

src/test/kotlin/io/snyk/plugin/net/CliConfigServiceTest.kt
```diff
@@ -0,0 +1,71 @@
...
+import org.junit.Test
+import retrofit2.Retrofit
+import retrofit2.converter.gson.GsonConverterFactory
+import snyk.net.HttpClient
+import java.nio.charset.StandardCharsets
+
+class CliConfigServiceTest {
+    @JvmField
+    @Rule
+    val server: MockWebServer = MockWebServer()
```
bastiandoetsch, line 21: Love usage of an http server mock.

bastiandoetsch: We should use that for false-positive and cli download and all the rest, too ;)

bastiandoetsch (approved):
LGTM. Nothing blocking.

src/test/kotlin/io/snyk/plugin/net/CliConfigServiceTest.kt
pavel-snyk: si, senior :).

src/main/kotlin/io/snyk/plugin/services/SnykApiService.kt
pavel-snyk: I prefer functions, Artsiom is for properties. So it's 50/50, but for consistency reasons (userMe), I leave it as property.

PR #333, opened 2022-03-15
fix: extract images from Helm generated k8s yaml with image names inside quotes [ROAD-754]

ArtsiomCh:


src/main/kotlin/snyk/container/YAMLImageExtractor.kt
```diff
@@ -5,10 +5,25 @@ import com.intellij.openapi.project.Project
...
 import com.intellij.psi.PsiFile
 import io.snyk.plugin.findPsiFileIgnoringExceptions
+import org.intellij.lang.annotations.Language
 
 object YAMLImageExtractor {
     private val logger = Logger.getInstance(javaClass)
-    private val imageRegEx = "(?<=([-\\s]image:))(\\s*[a-zA-Z0-9./\\-_]+)(:)?([a-zA-Z0-9./\\-_]+)?".toRegex()
+
+    // see https://kubernetes.io/docs/concepts/containers/images/
+    @Language("regexp")
```
pavel-snyk, line 14: @ArtsiomCh , why do we need this `@Language` annotation?

src/main/kotlin/snyk/container/YAMLImageExtractor.kt
```diff
@@ -5,10 +5,25 @@ import com.intellij.openapi.project.Project
...
+    private const val imageNameSymbol = "[a-zA-Z0-9./\\-_]"
+
+    @Language("regexp")
+    private const val imageName = "($imageNameSymbol+(:)?($imageNameSymbol+)?)"
+
+    @Language("regexp")
+    private const val imageNameDescription = "$imageName|(\"$imageName\")"
+
+    @Language("regexp")
+    private const val imageTag = "[-\\s]image:\\s*"
```
bastiandoetsch, line 24: It would be nice to get rid of the look-back in the regexp, but does this cover leading whitespace, comments and trailing comments in the line, e.g. 
```
#      - image: nxginx/basti:basti-latest-12
      - image: nxginx/basti:basti-latest-12 # trailing comment
```

bastiandoetsch: I'll try some yamls later to check :)

src/main/kotlin/snyk/container/YAMLImageExtractor.kt
ArtsiomCh: to see regex highlighting in Idea (language injection)

src/main/kotlin/snyk/container/YAMLImageExtractor.kt
ArtsiomCh: tested with 
image
only `debian` image used for container scan

ArtsiomCh: in the following PR I'm also going to handle cases for image's digest `@` and both port and tag existence inside the name: `fictional.registry.example:<port>/imagename:<tag>`

src/main/kotlin/snyk/container/YAMLImageExtractor.kt
pavel-snyk: got it, thanks.

pavel-snyk (approved):
LGTM,
nitpick: ideal would be to unit test regex only, but the whole feature is covered by light platform test.

ArtsiomCh:
> LGTM, nitpick: ideal would be to unit test regex only, but the whole feature is covered by light platform test.

Integ tests here is more general and tests the required functionality, while regex is just implementation details (we can switch to manual string parsing and that would require rewrite such details specific tests.)

PR #334, opened 2022-03-16
Container images parsing issues [ROAD-755]

ArtsiomCh:
fix: Container should distinct image names when calling CLI;
fix: Container should add annotations for all duplicated images
fix: Container scan should correctly proceed images with registry hostname in the name
fix: Container scan should extract images with `port` and `tag` as well as `digest` in the path


src/main/kotlin/snyk/container/ContainerService.kt
```diff
@@ -49,16 +48,41 @@ class ContainerService(project: Project) : CliAdapter<ContainerResult>(
...
             )
-            containerIssueImageList.add(enrichedContainerIssuesForImage)
+            patchedIssueImageList.add(enrichedContainerIssuesForImage)
         }
         LOG.debug("container service scan completed")
-        return ContainerResult(containerIssueImageList, null)
+        return ContainerResult(patchedIssueImageList, null)
+    }
+
+    /**
```
bastiandoetsch, line 66: We should ask them again - I had done so a few weeks ago, but didn't get a reply.

bastiandoetsch (approved):
Nice one! LGTM!

src/main/kotlin/snyk/container/ContainerService.kt
ArtsiomCh: it will take time on their side anyway. So we need some workaround for now...

PR #343, opened 2022-03-31
fix: Sanitize TextRange in file before accessing it

ArtsiomCh:
will prevent https://sentry.io/organizations/snyk/issues/3142615100 kind of exception.


bastiandoetsch (approved):
Morning @ArtsiomCh, looks good to me!

PR #348, opened 2022-04-05
refactoring for all description panels for unified and extendable UI [ROAD-541]

ArtsiomCh:
Will unify all description panels.
Also provide ability to add buttons in the bottom right place (according to the latest UI design) and possibility to add late custom panel in the bottom left.


ArtsiomCh:
#349 will substitute it.

PR #355, opened 2022-04-08
chore: enable request logging when SnykCode logger is in debug mode [ROAD-843]

bastiandoetsch:
### Description

This PR enables HTTP request logging for the Snyk Code Client when the `SnykCode` logger is set to debug and enables to monitor / check API communication, thus.

### Checklist

- [ x ] Tests all succeed
- [ x ] Linted
- [ x ] CHANGELOG.md updated
- [ x ] README.md updated, if user-facing

### DIFF
--- a/.snyk
+++ b/.snyk
@@ -1,0 +1,18 @@
+# Snyk (https://snyk.io) policy file, patches or ignores known vulnerabilities.
+version: v1.25.0
+# ignores vulnerabilities until expiry date; change duration by modifying expiry date
+ignore:
+  '*':
+    - build/*:
+        reason: None Given
+        expires: 2032-05-20T07:12:38.106Z
+        created: 2022-04-20T07:12:38.108Z
+    - src/integTest/*:
+        reason: None Given
+        expires: 2022-05-20T07:13:02.819Z
+        created: 2022-04-20T07:13:02.822Z
+    - src/test/*:
+        reason: None Given
+        expires: 2022-05-20T07:13:09.929Z
+        created: 2022-04-20T07:13:09.932Z
+patch: {}
--- a/.snyk
+++ b/.snyk
@@ -7,12 +7,4 @@ ignore:
         reason: None Given
         expires: 2032-05-20T07:12:38.106Z
         created: 2022-04-20T07:12:38.108Z
-    - src/integTest/*:
-        reason: None Given
-        expires: 2022-05-20T07:13:02.819Z
-        created: 2022-04-20T07:13:02.822Z
-    - src/test/*:
-        reason: None Given
-        expires: 2022-05-20T07:13:09.929Z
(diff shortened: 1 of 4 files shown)


ArtsiomCh (changes requested):
Would be happy and not create the ticket if fix could be such an easy thing to do... :)

- We re-create `retrofit` from few other places, so that `requestLogin` param will be overridden to the default `false` there. See `ai.deepcode.javaclient.core.DeepCodeParamsBase#setApiUrl()` usages.
- Not sure where `HttpLoggingInterceptor` will output. Most likely to STDOUT and that probably(?) will go to Idea log as well, but need to be checked. 
- Don't think we want to see such a detailed logs (at least I see in failed tests it's very verbose) every time we have DEBUG level enabled for the snyk classes. Would prefer to have separate feature flag in Idea's Registry for that.
- also probably some tests need to be added...

bastiandoetsch:
moving to slack...

.snyk
```diff
@@ -0,0 +1,18 @@
+# Snyk (https://snyk.io) policy file, patches or ignores known vulnerabilities.
+version: v1.25.0
+# ignores vulnerabilities until expiry date; change duration by modifying expiry date
+ignore:
+  '*':
+    - build/*:
+        reason: None Given
+        expires: 2032-05-20T07:12:38.106Z
+        created: 2022-04-20T07:12:38.108Z
+    - src/integTest/*:
```
ArtsiomCh, line 10: do we really want to ignore any vulns in our tests? may be just in resources...

.snyk
```diff
@@ -0,0 +1,18 @@
...
+        reason: None Given
+        expires: 2022-05-20T07:13:02.819Z
+        created: 2022-04-20T07:13:02.822Z
+    - src/test/*:
```
ArtsiomCh, line 14: see above

src/main/kotlin/io/snyk/plugin/snykcode/core/SnykCodeParams.kt
```diff
@@ -19,10 +20,30 @@ class SnykCodeParams private constructor() : DeepCodeParamsBase(
     { getWaitForResultsTimeout() },
     codeRestApi
 ) {
+    private val requestLogging = Logger.getInstance(SCLogger.presentableName + "RequestLogging").isDebugEnabled
```
ArtsiomCh, line 23: that should be `fun()` (or val with `get()`) cause we need to re-read setting every time it's called out (user can change it after class created).

ArtsiomCh (changes requested):
Some (easy) changes needed
Also please rebase CHANGELOG to resolve merge conflict

ArtsiomCh (approved):
LGTM
Good job!

PR #360, opened 2022-04-13
chore: centralize services invocation [ROAD-862]

ArtsiomCh:
Will help to control/change way we do services invocation/instantiation
Also prepare code base to be buildable with 221


bastiandoetsch (approved):
LGTM

PR #366, opened 2022-04-19
fix: generic .dcignore file creation failure and exceptions [ROAD-531]

### LINKED ISSUE #173: Plugin breaks new templated projects
With the plugin installed, creating a new project using the React template (CRA) will fail.

Snyk immediately adds a `.dcignore` file at the project root, and by the time the `create-react-app` initialization script runs, the directory won't be empty, which is a requirement, causing the script to error and exit.

ArtsiomCh:
### Description

will fix #173 and related `IOException: Permission denied` exceptions

### Checklist

- [x] Linted
- [x] CHANGELOG.md updated
- [x] README.md updated, if user-facing

### DIFF
--- a/src/main/kotlin/io/snyk/plugin/snykcode/core/SnykCodeIgnoreInfoHolder.kt
+++ b/src/main/kotlin/io/snyk/plugin/snykcode/core/SnykCodeIgnoreInfoHolder.kt
@@ -34,14 +34,22 @@ class SnykCodeIgnoreInfoHolder private constructor() : DeepCodeIgnoreInfoHolderB
         val prjBasePath = project.basePath
         val gitignore = File("$prjBasePath/.gitignore")
         val dcignore = File("$prjBasePath/.dcignore")
-        if (!gitignore.exists() && dcignore.createNewFile()) {
-            val fullDcIgnoreText = this.javaClass.classLoader.getResource("full.dcignore")
-                ?.readText()
-                ?: throw RuntimeException("full.dcignore can not be found in plugin's resources")
-            dcignore.writeText(fullDcIgnoreText)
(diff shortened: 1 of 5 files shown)


src/main/kotlin/io/snyk/plugin/snykcode/core/SnykCodeIgnoreInfoHolder.kt
```diff
@@ -34,14 +34,22 @@
...
-        if (!gitignore.exists() && dcignore.createNewFile()) {
-            val fullDcIgnoreText = this.javaClass.classLoader.getResource("full.dcignore")
-                ?.readText()
-                ?: throw RuntimeException("full.dcignore can not be found in plugin's resources")
-            dcignore.writeText(fullDcIgnoreText)
-            SnykBalloonNotificationHelper.showInfo(
-                "We added generic .dcignore file to upload only project's source code.", project
-            )
+        if (!gitignore.exists()) {
+            val genericDcIgnoreFileCreated = try {
```
bastiandoetsch, line 38: Should we just remove this and do it on first scan only? I don't see a down-side to having it only when scanning...

bastiandoetsch (approved):
Nothing hindering approval! Good job.

src/main/kotlin/io/snyk/plugin/snykcode/core/SnykCodeIgnoreInfoHolder.kt
ArtsiomCh: We can... just a bit more code/flags to recognize first scan of project....
Also having it for any Code scan helps to reduce load to our backend even in case user delete (accidently) .dcignore between scans

PR #371, opened 2022-05-01
fix: `InvocationTargetExceptions` by extracting caches to separate class [ROAD-813]

ArtsiomCh:
### Checklist

- [x] Linted
- [x] CHANGELOG.md updated

### DIFF
--- a/src/main/kotlin/snyk/common/SnykCachedResults.kt
+++ b/src/main/kotlin/snyk/common/SnykCachedResults.kt
@@ -1,0 +1,58 @@
+package snyk.common
+
+import com.intellij.openapi.components.Service
+import com.intellij.openapi.project.Project
+import io.snyk.plugin.events.SnykScanListener
+import io.snyk.plugin.snykcode.SnykCodeResults
+import snyk.container.ContainerResult
+import snyk.iac.IacResult
+import snyk.oss.OssResult
+
+@Service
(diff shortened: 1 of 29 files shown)


src/main/kotlin/snyk/common/SnykCachedResults.kt
```diff
@@ -0,0 +1,58 @@
...
+        get() = if (field?.isExpired() == false) field else null
+
+    var currentIacResult: IacResult? = null
+        get() = if (field?.isExpired() == false) field else null
+
+    fun initCacheUpdater() = project.messageBus.connect().subscribe(
+        SnykScanListener.SNYK_SCAN_TOPIC,
+        object : SnykScanListener {
+
+            override fun scanningStarted() {}
```
bastiandoetsch, line 27: These empty code blocks can be removed, I guess?

bastiandoetsch (approved):
Nicely done! Cool fix & refactoring!

src/main/kotlin/snyk/common/SnykCachedResults.kt
ArtsiomCh: Unfortunately, not really :( we have to implement interface methods even if we don't need to do anything there. The replacement for `{}` could be ` = Unit` but it is even longer and imho less readable.

PR #375, opened 2022-05-09
fix: show container specific message on macOS [ROAD-910]

ArtsiomCh:
image
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@
 
 - Empty (or cancelled request for) proxy credentials
 - `IllegalStateException` when no `Document` found for invalid(obsolete) file
-- For Snyk Code multi-file issues code snippet in some data flow steps shown from wrong file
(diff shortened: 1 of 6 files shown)


ArtsiomCh:
@bastiandoetsch please, would you mind to check when time permit if it works on macOS?...

bastiandoetsch:
I'm not a 100% sure what I should see, but I think it's not correct: 
image

If this is not what's expected, let's take some time today to pair on it until it works.

ArtsiomCh:
@bastiandoetsch try my 2nd attempt please when time permit.
Here I was inspired by your idea of extra rows/columns for alignment to put there some spacer. That allowed us to use "pure" panelGridConstraints() without `fill = GridConstraints.FILL_HORIZONTAL` which seems to be the main suspect of 0 height at resulted JEditorPane

aldanchenko (approved):
LGTM 👍

PR #376, opened 2022-05-09
fix: release notes

bastiandoetsch:
### Description

add release notes to github release


ArtsiomCh (approved):
Hope that magic words will work better then the other

PR #382, opened 2022-05-18
feat: Show specific description when no IaC or OSS files found but corresponded scan performed [ROAD-644]

ArtsiomCh:
1. if no file supported by IaC is found, on click on the IaC root node in the tree, the vulnerability description panel should display the following text  (from CLI): `Could not find any valid IaC files`
2. if no file supported by OSS is found, on click on the OSS root node in the tree, the vulnerability description panel should display the following text (from CLI): `Could not detect supported target files in <project-path>.\nPlease see our documentation for supported languages and target files: https://snyk.co/udVgQ and make sure you are in the right directory.`

image

image

### DIFF
--- a/src/main/kotlin/io/snyk/plugin/ui/UIUtils.kt
+++ b/src/main/kotlin/io/snyk/plugin/ui/UIUtils.kt
@@ -19,6 +19,8 @@ import java.awt.Container
 import java.awt.Dimension
 import java.awt.Font
 import java.awt.Insets
+import java.util.regex.Matcher
+import java.util.regex.Pattern
 import javax.swing.ImageIcon
 import javax.swing.JComponent
 import javax.swing.JEditorPane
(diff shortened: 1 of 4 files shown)


src/main/kotlin/io/snyk/plugin/ui/UIUtils.kt
```diff
@@ -342,3 +344,36 @@ fun wrapWithScrollPane(panel: JPanel): JScrollPane {
     )
     return scrollPane
 }
+
+// "stolen" from https://stackoverflow.com/a/12053940/7577274
```
bastiandoetsch, line 348: Please have a look at `StringEscapeUtils.escapeHtml`. That should do the job and is part of our dependencies anyway.

bastiandoetsch: https://howtodoinjava.com/java/string/escape-html-encode-string/

bastiandoetsch (approved):
Looks good besides the unneeded `txt2Html` method. Nice!

src/main/kotlin/io/snyk/plugin/ui/UIUtils.kt
ArtsiomCh: Here we need not only escape html but also replace `\n` and convert html link into real html link (`<href>`)
But good point to delegate that escaping to the library!

PR #389, opened 2022-06-16
feat: Navigation from the Editor to Snyk ToolWindow [ROAD-887]

ArtsiomCh:
feat: Navigation from the Editor (using QuickFix's functionality atnnotations) to Snyk ToolWindow (correspondent node in the Tree opened and Description shown) for:

  - Open Source scan results;
  - SnykCode scan results;
  - Infrastructure as a Code scan results;
  - Container scan results.

fix: Some minor UI renaming and alignments
image
image

PR #390, opened 2022-06-18
Chore: ToolWindowPanel refactoring

ArtsiomCh:
~~!!! Could/Should be reviewed/merged only after #389 !!!~~
- organizing ToolWindow related classes into structured packages;
- extracting RootTreeNodes into separate classes;
- extracting Analytic events reporting into correspondent classes;
- extracting Description panel creation into correspondent classes;
- clean up redundant classes;
- move Errors caching/handling into SnykCachedResults;


src/integTest/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanelIntegTest.kt
```diff
@@ -244,16 +254,16 @@ class SnykToolWindowPanelIntegTest : HeavyPlatformTestCase() {
         val snykError = SnykError(SnykToolWindowPanel.NO_OSS_FILES, project.basePath.toString())
         val snykErrorControl = SnykError("control", project.basePath.toString())
 
-        toolWindowPanel.snykScanListener.scanningOssError(snykErrorControl)
-        toolWindowPanel.snykScanListener.scanningOssError(snykError)
+        scanPublisher.scanningOssError(snykErrorControl)
```
bastiandoetsch, line 257: Why do we react/listen to messages on a publisher? This seems counter-intuitive to me - maybe something like eventHandler as a name would be better, and the event-handler method names would be sth like `onScanningOssError` if you don't want to separate publisher and consumer?

ArtsiomCh: Here we don't react, but send the event to publisher (to imitate end of OSS scan).
Then all subscribers will react accordingly (and we'll check state after).
PS here we now need to send event to the publisher (instead of hacky way to use toolWindowPanel.snykScanListener) cause not only ToolWindow (but also SnykCachedResults too) is processing that events now.

src/main/kotlin/io/snyk/plugin/ui/toolwindow/nodes/DescriptionHolderTreeNode.kt
```diff
@@ -0,0 +1,7 @@
+package io.snyk.plugin.ui.toolwindow.nodes
+
+import io.snyk.plugin.ui.toolwindow.panels.IssueDescriptionPanelBase
+
+interface DescriptionHolderTreeNode {
+    fun getDescriptionPanel(logEventNeeded : Boolean): IssueDescriptionPanelBase
```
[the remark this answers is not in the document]
bastiandoetsch: Yeah, pls format :)

bastiandoetsch (commented):
This is rather impossible to review... The code looks much cleaner, even though the SnykToolWindowPanel still could be smaller - that's the next refactoring then ;).

bastiandoetsch (approved):

PR #411, opened 2022-09-29
fix: snyk code enablement from preference dialog during authentication

bastiandoetsch:
### Description

The checkboxes remained disabled, as we asked the API without a token.

### Checklist

- [x] Linted
- [x] CHANGELOG.md updated

### Screenshots / GIFs

### DIFF
--- a/src/main/kotlin/io/snyk/plugin/services/SnykApiService.kt
+++ b/src/main/kotlin/io/snyk/plugin/services/SnykApiService.kt
@@ -123,8 +123,8 @@ class SnykApiService : Disposable {
                 baseUrl = baseUrl,
                 disableSslVerification = disableSslVerification
             )
-        } catch (t: Throwable) {
-            log.warn("Failed to create Retrofit client for endpoint: $endpoint", t)
+        } catch (ignore: RuntimeException) {
+            log.warn("Failed to create Retrofit client for endpoint: $endpoint", ignore)
             null
         }
 
--- a/src/main/kotlin/io/snyk/plugin/services/SnykApiService.kt
+++ b/src/main/kotlin/io/snyk/plugin/services/SnykApiService.kt
@@ -23,11 +23,10 @@ import javax.net.ssl.X509TrustManager
 
 @Service
 class SnykApiService : Disposable {
-
-    val sastSettings: CliConfigSettings?
-        get() {
-            return getSnykApiClient()?.sastSettings(pluginSettings().organization)
-        }
+    fun getSastSettings(token: String? = pluginSettings().token): CliConfigSettings? {
+        if (token == null) return null
+        return getSnykApiClient(token)?.sastSettings(pluginSettings().organization)
+    }
 
     val userId: String?
         get() = getSnykApiClient()?.getUserId()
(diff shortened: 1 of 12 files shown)


src/main/kotlin/io/snyk/plugin/ui/settings/ScanTypesPanel.kt
```diff
@@ -207,15 +207,20 @@ class ScanTypesPanel(
         }
     }
 
-    private fun checkSastEnable() {
+    fun checkSastEnabled(token: String? = pluginSettings().token) {
         setSnykCodeAvailability(false)
+        if (token.isNullOrBlank()) {
+            showSnykCodeAlert("A Snyk Token is necessary to check for Snyk Code enablement.")
```
michelkaporin, line 213: Where does this get shown? Screenshot would be great 🙏

src/main/kotlin/io/snyk/plugin/settings/SnykProjectSettingsConfigurable.kt
```diff
@@ -88,6 +92,13 @@ class SnykProjectSettingsConfigurable(val project: Project) : SearchableConfigur
             getSyncPublisher(project, SnykResultsFilteringListener.SNYK_FILTERING_TOPIC)?.filtersChanged()
             getSyncPublisher(project, SnykProductsOrSeverityListener.SNYK_ENABLEMENT_TOPIC)?.enablementChanged()
         }
+
+        runBackgroundableTask("Identifying with Analytics service", project, true) {
```
michelkaporin, line 96: Why was it moved here?

michelkaporin:
Let's address new detekt warnings 🙏

src/main/kotlin/io/snyk/plugin/ui/settings/ScanTypesPanel.kt
bastiandoetsch: It's only shown if no token is set, as we can only check Snyk Code enablement with a working API token.

src/main/kotlin/io/snyk/plugin/settings/SnykProjectSettingsConfigurable.kt
bastiandoetsch: It was blocking the UI thread, which is bad - on my computer, Sentry & Amplitude are DNS blocked, so the connection times out only after 60s and would block the UI for that amount of time.

src/main/kotlin/io/snyk/plugin/services/SnykApiService.kt
```diff
@@ -98,33 +97,37 @@
 
     private var currentUniqSnykApiClient: UniqSnykApiClient? = null
 
-    private fun getSnykApiClient(): SnykApiClient? {
+    private fun getSnykApiClient(token: String? = pluginSettings().token): SnykApiClient? {
```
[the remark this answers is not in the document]
bastiandoetsch: Won't fix this. It is just a stupid detekt rule that actively discourages guard expressions. I'll deactivate / adjust it.

michelkaporin (approved):

PR #424, opened 2023-02-15
feat: provided a customer http header with organization name

Arvi3d:
### Description

When user has several organizations with different settings, this might cause unpredictable and conflicting behaviours.
Current implementation does not send a required HTTP header.
This PR solves it.

https://snyksec.atlassian.net/browse/ZEN-357

### Checklist

- [ ] Linted
- [ ] CHANGELOG.md updated

### Screenshots / GIFs


src/main/java/ai/deepcode/javaclient/requests/GetAnalysisRequest.java
```diff
@@ -102,47 +102,63 @@ public int hashCode() {
...
+  }
+
+  private static class AnalysisContext {
+    private final String flow;
+    private final String initiator = "IDE";
+    private AnalysisContextOrg org;
+
+    public AnalysisContext(String flow, String orgName) {
+      this.flow = flow;
+      this.org = new AnalysisContextOrg(orgName);
```
bastiandoetsch, line 140: What happens if the org is empty?

src/main/java/ai/deepcode/javaclient/DeepCodeRestApiImpl.java
```diff
@@ -118,24 +118,28 @@ private interface CreateBundleCall {
     @retrofit2.http.Headers("Content-Type: application/json")
     @POST("bundle")
     Call<CreateBundleResponse> doCreateBundle(
-      @Header("Session-Token") String token, @Body FileContentRequest files);
+      @Header("Session-Token") String token,
+      @Header("snyk-org-name") String orgName,
```
bastiandoetsch, line 122: What's happening if the org is empty? Question applies to all calls, I think.

src/main/java/ai/deepcode/javaclient/core/AnalysisDataBase.java
```diff
@@ -559,7 +559,7 @@ private List<String> checkBundle(
     @NotNull String bundleId
   ) throws TokenInvalid401Exception, BundleIdExpire404Exception, ApiCallNotSucceedException {
     CreateBundleResponse checkBundleResponse =
-      restApi.checkBundle(deepCodeParams.getSessionToken(), bundleId);
+      restApi.checkBundle(deepCodeParams.getSessionToken(), deepCodeParams.getOrgDisplayName(), bundleId);
```
bastiandoetsch, line 562: Repeating, just in case: This can be empty - is that a problem?

src/main/java/ai/deepcode/javaclient/DeepCodeRestApiImpl.java
Arvi3d: it's fine. in this case default org will be taken into account

src/main/java/ai/deepcode/javaclient/requests/GetAnalysisRequest.java
Arvi3d: it's fine. in this case default org will be taken into account

src/main/java/ai/deepcode/javaclient/core/AnalysisDataBase.java
Arvi3d: not a problem. empty string is totally fine

bastiandoetsch (approved):
Looks fine, works for me ;)

Arvi3d:
@bastiandoetsch I don't know about current release approach. Will it be published automatically? Or someone needs to do it manually? If latter, may I ask you please to publish it?

bastiandoetsch:
It will automatically publish on Tuesday

bastiandoetsch:
This is the release scheduling: https://github.com/snyk/snyk-intellij-plugin/blob/cc615a0bb485206f66265a51502c7f6d70bcd590/.github/workflows/release.yml#L16

PR #447, opened 2023-08-28
feat: Enabling SCLE [NEBULA-1371]

metju90:
Enables SCLE


src/main/kotlin/snyk/common/CustomEndpoints.kt
```diff
@@ -1,14 +1,16 @@
 package snyk.common
 
 import io.snyk.plugin.pluginSettings
-import io.snyk.plugin.prefixIfNot
```
metju90, line 4: This will be re-added. Have uncommitted commeted code that made `prefixIfnot` unused

src/main/kotlin/io/snyk/plugin/net/CliConfigService.kt
```diff
@@ -25,13 +25,25 @@ data class CliConfigSettings(
     val localCodeEngine: LocalCodeEngine,
 
     @SerializedName("reportFalsePositivesEnabled")
-    val reportFalsePositivesEnabled: Boolean
+    val reportFalsePositivesEnabled: Boolean,
+
+    @SerializedName("code")
```
bastiandoetsch, line 30: I suppose this is the HTTP Code and HTTP message?

I don't think that technical information should be part of a return data class/object. It's exposing technical details that don't need to be exposed. You're mixing normal flow with error flow - which I would advise against. 

It would in my opinion be better to have only the fields needed in the data class, and have error/exception handling done separately, setting the config accordingly.

src/main/kotlin/io/snyk/plugin/net/SnykApiClient.kt
```diff
@@ -29,13 +31,25 @@
             log.debug("Executing request to $apiName")
             val response = retrofitCall.execute()
             if (!response.isSuccessful) {
+                if (response.code() == 422) {
```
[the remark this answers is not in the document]
bastiandoetsch: There is constants for each error code in the http libraries I think. Would have to look up where, though.

src/main/kotlin/io/snyk/plugin/net/SnykApiClient.kt
```diff
@@ -29,13 +31,25 @@
...
+                    val cliConfigSetting = Gson().fromJson<CliConfigSettings?>(response.errorBody()?.string(), CliConfigSettings::class.java)
```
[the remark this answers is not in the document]
bastiandoetsch: I know this Detekt stuff is annoying, but please fix the issues :) once the PR is ready :)

src/main/kotlin/io/snyk/plugin/net/SnykApiClient.kt
```diff
@@ -29,13 +31,25 @@ class SnykApiClient(
...
+                    return null
+                }
                 log.warn("Failed to execute `$apiName` call: ${response.errorBody()?.string()}")
                 return executeRequest(apiName, retrofitCall.clone(), retryCounter - 1)
             } else {
                 return response.body()
             }
         } catch (t: Throwable) {
             log.warn("Failed to execute '$apiName' network request: ${t.message}", t)
+            if (t is ClientException) {
```
bastiandoetsch, line 48: You can just have a catch block before the block with t: Throwable instead of an if.

src/main/kotlin/io/snyk/plugin/ui/settings/ScanTypesPanel.kt
```diff
@@ -218,26 +219,20 @@ class ScanTypesPanel(
...
             return
         }
-
+        val sastCliConfigSettings: CliConfigSettings? = getSnykApiService().getSastSettings();
+        settings.sastOnServerEnabled = sastCliConfigSettings?.sastEnabled
+        settings.localCodeEngineEnabled = sastCliConfigSettings?.localCodeEngine?.enabled
+        settings.localCodeEngineUrl = sastCliConfigSettings?.localCodeEngine?.url
         val snykCodeAvailable = isSnykCodeAvailable(settings.customEndpointUrl)
         showSnykCodeAlert(
             if (snykCodeAvailable) "" else "Snyk Code only works in SAAS mode (i.e. no Custom Endpoint usage)"
```
bastiandoetsch, line 228: Is this message still correct?

src/main/kotlin/snyk/common/CustomEndpoints.kt
```diff
@@ -1,14 +1,16 @@
...
 import io.snyk.plugin.suffixIfNot
 import java.net.URI
 import java.net.URISyntaxException
 
 fun toSnykCodeApiUrl(endpointUrl: String?): String {
     val endpoint = resolveCustomEndpoint(endpointUrl)
     val uri = URI(endpoint)
+    if (uri.isLocalCodeEngine()) {
```
bastiandoetsch, line 11: As said before, this should be moved into the when in line 16

src/main/kotlin/io/snyk/plugin/net/SnykApiClient.kt
bastiandoetsch, line 1: I haven't seen any tests yet that test the functionality:

- using the right URL
- reacting correctly to an HTTP status

metju90: I didn't run  any tests yet, will do so when we are confident we don't need to shuffle things around

src/main/kotlin/io/snyk/plugin/net/SnykApiClient.kt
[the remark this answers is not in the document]
metju90: > There is constants for each error code

I don't think so, seems like we are checking hardcoded integers for status codes. There are some constants, but they seem to be dedicated to IaC (see https://github.com/snyk/snyk-intellij-plugin/blob/8e30a782d1c944d8529d83752efa15b3611ad9d6/src/main/kotlin/snyk/iac/IacError.kt#L3)

image

src/main/kotlin/io/snyk/plugin/ui/settings/ScanTypesPanel.kt
metju90: Was wondering if we support ST envs. There is a unit test which confirms we do.

I think this `isSnykCodeAvailable` check can be removed altogether. The only purpose I see in this function is to block custom URLs Snyk has no association with. i.e. a user would put `http://google.com` as a custom URL in the plugin settings. If this happens, the plugin will fail the first request to `http://google.com/api/v1/user.me`.

src/main/kotlin/snyk/common/CustomEndpoints.kt
metju90: The only reason I kept it here is because we have no control on how SCLE URLs will look like, its 100% up to the customers. SCLE URLs does not follow our URL conventions.

The return at  line26/28 removes and replaces some text from the URL, this could potentially break a SCLE URL matching any of the rules in the suffix/replacement

src/main/kotlin/snyk/common/CustomEndpoints.kt
```diff
@@ -29,6 +31,9 @@ fun toSnykCodeApiUrl(endpointUrl: String?): String {
 fun toSnykCodeSettingsUrl(endpointUrl: String?): String {
     val endpoint = resolveCustomEndpoint(endpointUrl)
     val uri = URI(endpoint)
+    if (uri.isLocalCodeEngine()) {
```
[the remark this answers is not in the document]
metju90: https://github.com/snyk/snyk-intellij-plugin/pull/447/files#r1308404903

src/main/kotlin/io/snyk/plugin/net/CliConfigService.kt
```diff
@@ -28,10 +28,21 @@ data class CliConfigSettings(
     val reportFalsePositivesEnabled: Boolean
 )
 
+data class SastSettingsError(
+    @SerializedName("userMessage")
+    val userMessage: String
+)
+
```
metju90, line 35: @bastiandoetsch I moved error flow to a different class.

src/main/kotlin/io/snyk/plugin/net/SnykApiClient.kt
```diff
@@ -29,11 +31,24 @@
             log.debug("Executing request to $apiName")
             val response = retrofitCall.execute()
             if (!response.isSuccessful) {
+                if (response.code() == 422) {
```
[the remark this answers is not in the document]
bastiandoetsch: Try `HttpStatusCode.UnprocessableEntity.value` from the okhttp dependency. Would also be easier to understand :)

src/main/kotlin/io/snyk/plugin/ui/settings/ScanTypesPanel.kt
```diff
@@ -219,25 +221,31 @@
...
+            null
+        }
+
+        settings.sastOnServerEnabled = sastCliConfigSettings?.sastEnabled
+        settings.localCodeEngineEnabled = sastCliConfigSettings?.localCodeEngine?.enabled
+        settings.localCodeEngineUrl = sastCliConfigSettings?.localCodeEngine?.url
         val snykCodeAvailable = isSnykCodeAvailable(settings.customEndpointUrl)
         showSnykCodeAlert(
-            if (snykCodeAvailable) "" else "Snyk Code only works in SAAS mode (i.e. no Custom Endpoint usage)"
+            if (sastSettingsError.isNotEmpty()) sastSettingsError else ""
```
bastiandoetsch, line 241: I get a warning in IntelliJ here - please use `ifEmpty`

src/main/kotlin/io/snyk/plugin/Utils.kt
```diff
@@ -193,7 +194,12 @@
     lateinit var checkIfSastEnabled: () -> Unit
     checkIfSastEnabled = {
         if (settings.sastOnServerEnabled != true) {
-            settings.sastOnServerEnabled = getSnykApiService().getSastSettings()?.sastEnabled ?: false
+            settings.sastOnServerEnabled = try {
+                getSnykApiService().getSastSettings()?.sastEnabled ?: false
+            } catch (t: ClientException) {
```
bastiandoetsch, line 199: please rename `t` to `ignored` so the exception is marked as intentionally swallowed for the linter

src/main/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanel.kt
```diff
@@ -488,10 +489,13 @@
 
     private fun enableCodeScanAccordingToServerSetting() {
         pluginSettings().apply {
-            val sastSettings = getSnykApiService().getSastSettings()
+            val sastSettings = try {
+                getSnykApiService().getSastSettings()
+            } catch (t: ClientException) {
```
bastiandoetsch, line 494: Please name the exception `ignored` so that the linter knows what's happening

src/main/kotlin/io/snyk/plugin/ui/UIUtils.kt
```diff
@@ -96,11 +96,9 @@ fun insertTitleAndResizableTextIntoPanelColumns(
 
 fun snykCodeAvailabilityPostfix(): String {
     val sastOnServerEnabled = pluginSettings().sastOnServerEnabled
-    val localCodeEngineEnabled = pluginSettings().localCodeEngineEnabled
     return when {
         !isSnykCodeAvailable(pluginSettings().customEndpointUrl) -> " (disabled for endpoint)"
-        (sastOnServerEnabled == null && localCodeEngineEnabled == null) -> " (unreachable server settings)"
-        (sastOnServerEnabled != false && localCodeEngineEnabled == true) -> " (disabled due to Local Code Engine)"
+        sastOnServerEnabled == null -> " (unreachable server settings)"
```
bastiandoetsch, line 101: Couldn't this also be a misconfigured URL? Is this message still correct?

src/main/kotlin/snyk/common/CustomEndpoints.kt
bastiandoetsch: Why not:
```kotlin
fun toSnykCodeApiUrl(endpointUrl: String?): String {
    val endpoint = resolveCustomEndpoint(endpointUrl)
    val uri = URI(endpoint)

    val codeSubdomain = "deeproxy"
    val snykCodeApiUrl = when {
        uri.isLocalCodeEngine() ->
            return pluginSettings().localCodeEngineUrl!!
        
        uri.isDeeproxy() ->
            endpoint

        uri.isDev() ->
            endpoint.replace("https://dev.", "https://$codeSubdomain.dev.")

        uri.isSnykTenant() ->
            endpoint.replace("https://", "https://$codeSubdomain.")

        else -> "https://$codeSubdomain.snyk.io/"
    }
    return snykCodeApiUrl.removeSuffix("api").replace("app.", "")
}
```

src/main/kotlin/snyk/common/CustomEndpoints.kt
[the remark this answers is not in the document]
bastiandoetsch: see my comment there :)

src/main/kotlin/io/snyk/plugin/ui/UIUtils.kt
metju90: That's correct, this could also be a 422 response from the server. 

We have two options:

1. add `sastResponseError:boolean` to `pluginSettings()`, then use this logic here
2. reword `unreachable server settings` to more generic, like `something went wrong while fetching SAST seetings`.

from UX perspective, option 1 is better. What do you think?

src/test/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanelTest.kt
```diff
@@ -144,7 +144,7 @@ class SnykToolWindowPanelTest : LightPlatform4TestCase() {
     fun `should automatically enable all products on first run after Auth`() {
         every { snykApiServiceMock.getSastSettings() } returns CliConfigSettings(
             true,
-            LocalCodeEngine(false),
+            LocalCodeEngine(true, "http://foo.bar/api", false),
```
bastiandoetsch, line 147: This is now missing a test for the "standard" Snyk Code enablement - would you please have both tested?

bastiandoetsch (commented):
Good job, just some small changes requested :)

src/main/kotlin/io/snyk/plugin/ui/UIUtils.kt
bastiandoetsch: I'd go for option 1.

metju90: Done

src/test/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanelTest.kt
metju90: Done

src/main/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanel.kt
metju90: Done

src/main/kotlin/io/snyk/plugin/ui/settings/ScanTypesPanel.kt
metju90: Done

src/main/kotlin/io/snyk/plugin/net/SnykApiClient.kt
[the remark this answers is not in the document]
metju90: Done

src/main/kotlin/snyk/common/CustomEndpoints.kt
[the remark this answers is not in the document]
metju90: Done

src/main/kotlin/snyk/common/CustomEndpoints.kt
metju90: Done

src/main/kotlin/io/snyk/plugin/net/SnykApiClient.kt
```diff
@@ -29,11 +31,22 @@ class SnykApiClient(
             log.debug("Executing request to $apiName")
             val response = retrofitCall.execute()
             if (!response.isSuccessful) {
+                if (response.code() == HttpStatusCode.UnprocessableEntity.value) {
+                    val responseBodyString = response.errorBody()?.string()
+                    val errorBody= Gson().fromJson<CliConfigSettingsError?>(responseBodyString, CliConfigSettingsError::class.java)
```
bastiandoetsch, line 36: Please fix the linter stuff :)

src/main/kotlin/io/snyk/plugin/services/SnykApplicationSettingsStateService.kt
```diff
@@ -48,7 +48,9 @@ class SnykApplicationSettingsStateService : PersistentStateComponent<SnykApplica
     var containerScanEnabled: Boolean = true
 
     var sastOnServerEnabled: Boolean? = null
+    var sastSettingsError: Boolean? = null;
```
bastiandoetsch, line 51: linter :)

src/main/kotlin/io/snyk/plugin/ui/settings/ScanTypesPanel.kt
```diff
@@ -219,25 +221,34 @@ class ScanTypesPanel(
             return
         }
 
+        var sastSettingsError: String = ""
+        val sastCliConfigSettings: CliConfigSettings? =
+        try {
```
bastiandoetsch, line 226: linter :)

bastiandoetsch (commented):
Nice! Only linting and a changelog entry are missing.

bastiandoetsch (approved):

PR #453, opened 2023-10-04
feat: download arm64 CLI on M1/M2 macs, allow baseURL configuration [HEAD-867]

bastiandoetsch:
### Description

This PR adds the new arm64 binary for the CLI download. Also, it allows to add the baseURL to download the CLI from, to faciliate usage of FIPS enabled binaries, or locally provided CLIs.

### Checklist

- [x] Linted
- [x] CHANGELOG.md updated

### Screenshots / GIFs


src/main/kotlin/io/snyk/plugin/ui/SnykSettingsDialog.kt
```diff
@@ -463,6 +465,15 @@ class SnykSettingsDialog(
             )
         )
 
+        cliBaseDownloadUrlTextField.toolTipText = "The default URL is https://static.snyk.io. " +
```
PeterSchafer, line 468: question: do we really want to mention this here? From a maintenance perspective, I think it would be better to leave this to the user-docs.

PeterSchafer (approved):

PR #467, opened 2024-01-23
feat: snyk controller extension point

cmars:
### Description

Add an extension point to allow third-party plugins to interact with Snyk in the IDE.

### Checklist

- [X] Linted
- [x] CHANGELOG.md updated
- [x] README.md updated, if user-facing

### Screenshots / GIFs

Extension in a demo plugin (built with a local build of this branch)

[snyk-extension-point.webm](https://github.com/snyk/snyk-intellij-plugin/assets/23741/ccaf5a77-6dfc-47f1-af41-abb6cc5bf14e)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 # Snyk Changelog
 
-## [2.6.1]
+## [2.7.0]
+### Added
+- Snyk controller extension point
+
+## [2.6.2]
 ### Fixed
 - remove more intellij-specific code and configuration
 
--- a/EXTENSIONS.md
+++ b/EXTENSIONS.md
@@ -1,0 +1,102 @@
+# Snyk Extensions
+
+This plugin offers an extension point for integrating Snyk with other Jetbrains IDE Plugins.
+
+## What is the Snyk Controller?
+
+The Snyk Controller is, put simply, a way to control Snyk from other plugins.
+
+## Why would I use the Snyk Controller extension?
+
+There are a few of use cases for the Snyk controller:
+
+### Initiating a Snyk scan
+
+There may be situations in which a plugin wants to initiate a security scan, especially at the end of a workflow which
+introduces changes to the project source code, manifest dependencies, OCI container builds, infrastructure files,
+etc. -- anything Snyk can scan.
+
+### Determining whether Snyk is authenticated to a user
+
+It might be useful to know whether the user has authenticated with Snyk; is Snyk ready to run scans?
+
+## How do I use the extension in my plugin?
+
+### Add a dependency on this plugin to your plugin's `plugin.xml` file.
+
+Release >= 2.7.0 provides the extension point.
+
+```xml
+<depends>io.snyk.snyk-intellij-plugin</depends>
+```
+
+### Declare the extension
+
+Add an extension for `io.snyk.snyk-intellij-plugin.controllerManager` to your plugin's `plugin.xml` file.
+
+```xml
+<extensions defaultExtensionNs="io.snyk.snyk-intellij-plugin">
+  <controllerManager implementation="com.example.demointellij.MySnykControllerManager"/>
+</extensions>
+```
+
+### Optional dependency
+
+The dependency on Snyk can also be optional:
+
+```xml
+<depends optional="true" config-file="optional/withSnyk.xml">io.snyk.snyk-intellij-plugin</depends>
+```
+
+Declare the controller extension in the dependency's config file, located relative to the `plugin.xml`:
+
+```xml
+<!-- Contents of optional/withSnyk.xml -->
+<idea-plugin>
+    <extensions defaultExtensionNs="io.snyk.snyk-intellij-plugin">
+        <controllerManager implementation="com.example.demointellij.HelloWorldControllerManager" />
+    </extensions>
+</idea-plugin>
+```
+
+### Implement the controller manager interface
+
+The manager will be instantiated and passed an instance of the controller when the Snyk plugin is initialized.
+
+```kotlin
+package com.example.demointellij
+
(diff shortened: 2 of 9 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 # Snyk Changelog
 
-## [2.6.1]
+## [2.7.0]
```
cmars, line 3: I'm not sure if I should make this edit, or let gradle do it...

bastiandoetsch: Nah, so far we don't write the changelog automatically. Also, you need to release manually, as automatic releases only increase patch versions.

build.gradle.kts
```diff
@@ -133,6 +133,24 @@ tasks {
         }
     }
 
+    val createOpenApiSourceJar by registering(Jar::class) {
```
bastiandoetsch, line 136: What is this needed for?

EXTENSIONS.md
```diff
@@ -0,0 +1,83 @@
...
+It might be useful to know whether the user has authenticated with Snyk; is Snyk ready to run scans?
+
+## How do I use the extension in my plugin?
+
+### Add a dependency on this plugin to your plugin's `plugin.xml` file.
+
+Release >= 2.7.0 provides the extension point.
+
+```xml
+<depends>io.snyk.snyk-intellij-plugin</depends>
```
bastiandoetsch, line 30: Maybe mention [optional dependencies](https://plugins.jetbrains.com/docs/intellij/plugin-dependencies.html#optional-plugin-dependencies).

src/main/kotlin/io/snyk/plugin/SnykPostStartupActivity.kt
```diff
@@ -29,6 +32,13 @@ private val LOG = logger<SnykPostStartupActivity>()
 
 class SnykPostStartupActivity : ProjectActivity {
 
+    private object ExtensionPointsUtil {
+        val controllerLifecycleManager =
+            ExtensionPointName.create<SnykControllerLifecycleManager>(
+                "io.snyk.snyk-intellij-plugin.controllerLifecycleManager"
```
bastiandoetsch, line 38: This should be a constant 🤓

src/main/kotlin/io/snyk/plugin/SnykPostStartupActivity.kt
```diff
@@ -95,6 +105,10 @@ class SnykPostStartupActivity : ProjectActivity {
         if (isContainerEnabled()) {
             getKubernetesImageCache(project)?.scanProjectForKubernetesFiles()
         }
+
+        ExtensionPointsUtil.controllerLifecycleManager.extensionList.forEach {
+            it.register(SnykControllerImpl(project))
```
bastiandoetsch, line 110: This is quite neat. Should we maybe use a singleton implementation?

src/main/kotlin/io/snyk/plugin/extensions/SnykControllerLifecycleManager.kt
```diff
@@ -0,0 +1,11 @@
+package io.snyk.plugin.extensions
+
+// SnykControllerLifecycleManager is the extension point interface
+// which other plugins can implement in order to integrate with Snyk.
+interface SnykControllerLifecycleManager {
+
+    // register is called by Snyk on third-party implementations of this interface
```
bastiandoetsch, line 7: I'd make this more explicit and a JavaDoc comment (`/** */`). Also, while I'm nitpicking ;), I'd remove `Lifecycle` from the name. How about: `register is called by the Snyk IntelliJ plugin to pass the #SnykController to extension point implementers`.

src/main/resources/META-INF/plugin.xml
```diff
@@ -100,4 +100,10 @@
               description="Snyk: Show Low severity issues"/>
     </group>
   </actions>
+
+  <extensionPoints>
+    <extensionPoint
+      name="controllerLifecycleManager"
```
bastiandoetsch, line 106: how about `SnykControllerManager`? I'm not sure that in all searches and portals, the plugin the extension point comes from is highly visible.

bastiandoetsch (approved):
Great job. You rock, Casey! Some non-blocking comments that are opinions rather than needed changes ;).

build.gradle.kts
cmars: Bundling the source into the plugin, so that consumers of the extension point can compile against the interfaces we provide.

AIUI anyone who implements the extension will need this, per this guide: https://plugins.jetbrains.com/docs/intellij/bundling-plugin-openapi-sources.html#bundling-api-sources-in-gradle-build-script

EXTENSIONS.md
cmars: Can do! I need to test that case locally to convince myself it works. You'd still end up with a compile-time dependency, but at runtime, it makes sense that it should be optional (your lifecycle manager would just never get called).

src/main/kotlin/io/snyk/plugin/SnykPostStartupActivity.kt
cmars: I think we might benefit from a singleton here if we found the controller needed to become stateful and we wanted to share that state among all the extensions. Or if we were hyper-optimizing for every byte of heap?

Conversely, choosing a singleton now might conflict with a future need for per-extension state and we'd have to change it.

I'm +/- 0 on this one... we can always change the implementation later if we find we need to?

bastiandoetsch: Yeah, fine with me.

bastiandoetsch (approved):

PR #468, opened 2024-01-30
fix: CLI update & download [IDE-123][IDE-124]

bastiandoetsch:
### Description

We have received several reports that CLI updates under Windows don't work. This is due to a file lock.

This PR ensures two things:

1. Independent of the number of projects, only one instance of the CLI is started in language server mode `snyk language-server` in the background.
2. CLI download, whether initial or an update, comes first, before the CLI is started in the background. This ensures the download does not run into a file lock under Windows.

### Checklist

- [x] Linted
- [x] CHANGELOG.md updated

### Screenshots / GIFs

image

Opened 5 projects, and verified only one snyk-win.exe process:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Snyk Changelog
 
+## [2.7.1]
+### Fixed
+- only start up language server after CLI update (fixes lock error on Windows)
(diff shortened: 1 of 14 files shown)


acke (approved):
Very nice commit.

PR #477, opened 2024-02-29
chore: update snyk job

bastiandoetsch:
### Description

### Checklist

- [ ] Linted
- [ ] CHANGELOG.md updated

### Screenshots / GIFs

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -63,9 +63,9 @@ jobs:
         uses: snyk/actions/setup@0.4.0
 
       - name: Snyk Monitor SCA
-        run: snyk monitor --all-projects --exclude=integTest --org=${{ vars.PLATFORM_HAMMERHEAD_ORG_ID }}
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -63,9 +63,9 @@ jobs:
         uses: snyk/actions/setup@0.4.0
 
       - name: Snyk Monitor SCA
-        run: snyk monitor --all-projects --exclude=integTest --org=${{ vars.PLATFORM_HAMMERHEAD_ORG_ID }}
+        run: snyk monitor --all-projects --exclude=integTest --org=9cff56cd-57d1-49b8-9238-69ebfde7142f
```
cat2608, line 66: Do we want to have this org id hardcoded?

bastiandoetsch: What's the advantage to not having this in code and instead in github settings? It's not a secret either, so I don't see why we need a variable, making the config more complicated.

cat2608 (approved):

PR #484, opened 2024-03-06
refactor: remove deprecations and update them to current API

bastiandoetsch:
### Description

### Checklist

- [x] Linted
- [ ] CHANGELOG.md updated

### Screenshots / GIFs


src/main/kotlin/io/snyk/plugin/services/SnykCliAuthenticationService.kt
```diff
@@ -47,7 +48,7 @@ class SnykCliAuthenticationService(val project: Project) {
         downloadCliIfNeeded()
         if (getCliFile().exists()) executeAuthCommand()
         if (isAuthenticated) executeGetConfigApiCommand()
-
+        LanguageServerWrapper.getInstance().updateConfiguration()
```
teodora-sandu, line 51: Question: Was this something we should have added before and you just noticed?

bastiandoetsch: It was a merge problem. The method was renamed...

bastiandoetsch: Ok, let me rephrase. I noticed, that the token was not updated in Language Server after authentication. So I added it.

teodora-sandu (approved):
I've tested the Settings panel (changing the ignore settings and severities) and also had a look at Code Security and Code Quality issues. I haven't been able to test the other types of issues because of the bug we have with disabling/enabling other types of scanning from the Settings panel but I think it should be okay

PR #487, opened 2024-03-07
fix: check trust for content root before triggering Snyk Code scans

bastiandoetsch:
### Description

Previously, the trust setting was not checked before triggering a scan. This PR fixes this.

### Checklist

- [x] Linted
- [x] CHANGELOG.md updated

### Screenshots / GIFs

### DIFF
--- a/src/main/kotlin/snyk/common/lsp/LanguageServerWrapper.kt
+++ b/src/main/kotlin/snyk/common/lsp/LanguageServerWrapper.kt
@@ -1,6 +1,7 @@
 package snyk.common.lsp
 
 import com.intellij.ide.impl.ProjectUtil
+import com.intellij.openapi.components.service
 import com.intellij.openapi.diagnostic.Logger
 import com.intellij.openapi.project.Project
 import com.intellij.openapi.project.ProjectManager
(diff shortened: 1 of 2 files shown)


src/main/kotlin/snyk/common/lsp/LanguageServerWrapper.kt
```diff
@@ -214,7 +217,12 @@ class LanguageServerWrapper(
             logger.warn("No active project found, not sending scan command.")
             return
         }
-        project.getContentRootVirtualFiles().mapNotNull { it.path }.toSet().forEach(::sendFolderScanCommand)
+        project.getContentRootVirtualFiles().filterNotNull().toSet().forEach {
+            val trustService = service<WorkspaceTrustService>()
+            if (trustService.isPathTrusted(it.toNioPath())) {
+                sendFolderScanCommand(it.path)
+            }
+        }
```
cat2608, line 225: Checking what `toNioPath` I saw that it can `throw` so I wonder if we need to handle the error.
image
It seems that we can `try/catch` it?

cat2608 (approved):

src/main/kotlin/snyk/common/lsp/LanguageServerWrapper.kt
bastiandoetsch: This should not be possible to happen - but yeah, I'll make a note to include a catch with a user notification in the next PR.

PR #491, opened 2024-03-14
fix: workspace folder and content root handling [IDE-204]

bastiandoetsch:
### Description

Previously, content roots were not normalized, so that duplications could occur, e.g. when a child root would be included, even though the parent was already there.

### Checklist

- [x] Linted
- [x] CHANGELOG.md updated

### Screenshots / GIFs

### DIFF
--- a/src/main/kotlin/snyk/common/lsp/LanguageServerWrapper.kt
+++ b/src/main/kotlin/snyk/common/lsp/LanguageServerWrapper.kt
@@ -6,6 +6,7 @@ import com.intellij.openapi.diagnostic.Logger
 import com.intellij.openapi.project.Project
 import com.intellij.openapi.project.ProjectManager
 import com.intellij.openapi.util.io.toNioPathOrNull
+import com.intellij.openapi.vfs.VirtualFile
 import io.snyk.plugin.getCliFile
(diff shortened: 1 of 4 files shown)


cat2608 (approved):
🙏

src/main/kotlin/snyk/common/lsp/LanguageServerWrapper.kt
```diff
@@ -133,8 +134,36 @@ class LanguageServerWrapper(
...
+            for (normalizedRoot in normalizedRoots) {
+                if (!root.path.startsWith(normalizedRoot.path)) continue
+                add = false
+                break
+            }
+            if (add) normalizedRoots.add(root)
+        }
+        logger.debug("Normalized content roots: $normalizedRoots")
+        return normalizedRoots
+    }
```
cat2608, line 166: Does this work fine with super nested folders? Is there a way to check it?

bastiandoetsch: It does, the testing involves a lot of mocking of IntelliJ APIs though. It came to a point, that I deleted the test, because I had to know the implementation in detail. Maybe writing a LightPlatformTest would solve that instead of mocking. I'll look into it.

PR #495, opened 2024-03-25
feat: mark findings as ignored in tree view [IDE-193]

teodora-sandu:
### Description

Adds the `[ Ignored ]` text in front of ignored findings. It needs to be behind a feature flag because the backend will always return the ignore data, so the frontend needs to gate it.

In [this](https://snyk.slack.com/archives/C06LTKLFY2F/p1711376124767019) thread we decided not to show AI fixes if the finding is ignored.

### Checklist

- [x] Linted
- [x] CHANGELOG.md updated

### Screenshots / GIFs

Screenshot 2024-03-25 at 14 11 26

### DIFF
--- a/src/main/kotlin/io/snyk/plugin/ui/toolwindow/SnykTreeCellRenderer.kt
+++ b/src/main/kotlin/io/snyk/plugin/ui/toolwindow/SnykTreeCellRenderer.kt
@@ -111,7 +111,14 @@ class SnykTreeCellRenderer : ColoredTreeCellRenderer() {
                 val issue = value.userObject as ScanIssue
                 nodeIcon = SnykIcons.getSeverityIcon(issue.getSeverityAsEnum())
                 val range = issue.range
-                text = "${if (issue.additionalData.hasAIFix) "⚡️" else ""} ${
+                var showIgnoredLabel = false
+                var showAIFix = false
+                if (pluginSettings().isGlobalIgnoresFeatureEnabled && issue.isIgnored == true) {
+                    showIgnoredLabel = true
+                } else if (issue.additionalData.hasAIFix) {
+                    showAIFix = true
+                }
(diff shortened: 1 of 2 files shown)


src/main/kotlin/io/snyk/plugin/ui/toolwindow/SnykTreeCellRenderer.kt
```diff
@@ -111,7 +111,14 @@ class SnykTreeCellRenderer : ColoredTreeCellRenderer() {
                 val issue = value.userObject as ScanIssue
                 nodeIcon = SnykIcons.getSeverityIcon(issue.getSeverityAsEnum())
                 val range = issue.range
-                text = "${if (issue.additionalData.hasAIFix) "⚡️" else ""} ${
+                var showIgnoredLabel = false
+                var showAIFix = false
+                if (pluginSettings().isGlobalIgnoresFeatureEnabled && issue.isIgnored == true) {
```
acke, line 116: The `== true` is redundant if issue.isIgnored is a boolean?

acke (approved):
Looks good to me. :)

Just a small question.

src/main/kotlin/io/snyk/plugin/ui/toolwindow/SnykTreeCellRenderer.kt
teodora-sandu: `issue.IsIgnored` is an optional boolean, see here: https://github.com/snyk/snyk-intellij-plugin/blob/7a7810841bf5ff1842959f08cd65640c58917cbb/src/main/kotlin/snyk/common/lsp/Types.kt#L31

So basically this check is needed so that if `issue.isIgnored` is null then we treat that as a false. I'm not too good at Kotlin yet so IntelliJ asked me to write the code this way 😂

PR #496, opened 2024-03-26
feat: render html from snyk-ls [IDE-214]

teodora-sandu:
### Description

Uses `JCEF` to render HTML served from `snyk-ls`. This is built using the CLI from https://github.com/snyk/cli/pull/5138 and it looks incomplete because it is: styling and actions will be addressed in future PRs, as will different parts of the UI. This PR just wires up the rendering of the HTML so we can build on top of it.

Tested locally by following the instructions here: https://snyksec.atlassian.net/wiki/spaces/IDE/pages/1966866488/How+to+setup+development+for+Consistent+Ignores

### Checklist

- [x] Linted
- [x] CHANGELOG.md updated

### Screenshots / GIFs
- error is shown in balloon when feature flag is enabled but JCEF is not supported (should never happen but just in case)
Screenshot 2024-03-27 at 15 22 15
- panel gets rendered correctly when feature flag is enabled
Screenshot 2024-03-27 at 15 23 33
- old panel gets rendered when feature flag is disabled
Screenshot 2024-03-27 at 15 25 34

### DIFF
--- a/src/main/kotlin/io/snyk/plugin/ui/toolwindow/panels/SuggestionDescriptionPanelFromLS.kt
+++ b/src/main/kotlin/io/snyk/plugin/ui/toolwindow/panels/SuggestionDescriptionPanelFromLS.kt
@@ -4,8 +4,6 @@ import com.intellij.icons.AllIcons
 import com.intellij.ui.HyperlinkLabel
 import com.intellij.ui.ScrollPaneFactory
 import com.intellij.ui.components.JBTabbedPane
-import com.intellij.ui.jcef.JBCefApp
-import com.intellij.ui.jcef.JBCefBrowserBuilder
 import com.intellij.uiDesigner.core.GridConstraints
 import com.intellij.uiDesigner.core.GridLayoutManager
 import com.intellij.util.ui.JBInsets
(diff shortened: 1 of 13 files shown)


src/main/kotlin/io/snyk/plugin/ui/toolwindow/panels/SuggestionDescriptionPanelFromLS.kt
```diff
@@ -33,11 +40,36 @@ import kotlin.math.max
...
     val project = snykCodeFile.project
+    private val unexpectedErrorMessage = "Snyk encountered an issue while rendering the vulnerability description. Please try again, or contact support if the problem persists. We apologize for any inconvenience caused.";
 
     init {
-        createUI()
+        if (pluginSettings().isGlobalIgnoresFeatureEnabled && issue.additionalData.details != null) {
+            if (!JBCefApp.isSupported()) {
+                val statePanel = StatePanel(SnykToolWindowPanel.SELECT_ISSUE_TEXT)
+                this.add(wrapWithScrollPane(statePanel), BorderLayout.CENTER)
+                SnykBalloonNotificationHelper.showError(unexpectedErrorMessage, null)
```
cat2608, line 55: Is it possible that you wanted to use your new `showErrorBalloonForComponent` function instead of `showError`? It seems that `showErrorBalloonForComponent` doesn't have references 🤔

teodora-sandu: Ah originally I meant to do that while I was figuring out how to create the balloon but what I ended up doing worked better for our case. Good spot!

bastiandoetsch (approved):

cat2608 (approved):

PR #503, opened 2024-04-04
fix: uri mapping between intellij and ls [IDE-239]

bastiandoetsch:
### Description

IntelliJ VirtualFiles do not conform to the URI spec followed by our Language Server under WIndows.

If a drive-letter is part of the URL, an additional slash needs to be added before the drive letter. When receiving URIs, this additional slash needs to be removed, as IntelliJ does not understand it.

### Checklist

- [x] Linted
- [x] CHANGELOG.md updated

### Screenshots / GIFs

### DIFF
--- a/src/main/kotlin/io/snyk/plugin/Utils.kt
+++ b/src/main/kotlin/io/snyk/plugin/Utils.kt
@@ -430,9 +430,29 @@ fun String.toVirtualFile(): VirtualFile {
     return if (!this.startsWith("file://")) {
         StandardFileSystems.local().refreshAndFindFileByPath(this) ?: throw FileNotFoundException(this)
     } else {
-        VirtualFileManager.getInstance().refreshAndFindFileByUrl(this) ?: throw FileNotFoundException(this)
+        VirtualFileManager.getInstance().refreshAndFindFileByUrl(this.toVirtualFileURL()) ?: throw FileNotFoundException(this)
     }
 }
+
+// add a slash when on windows
+fun VirtualFile.toLanguageServerURL(): String {
+    if (this.urlContainsDriveLetter()) {
+        return this.url.replace("file://","file:///")
+    }
+    return this.url
+}
+
+// remove first "/" if on windows
+fun String.toVirtualFileURL(): String {
+    if (this.isWindowsURI() && this.startsWith("file:///") && this.length > 10 && this.substring(9,10) == ":") {
+        return this.replaceFirst("/","")
+    }
+    return this
+}
+
+fun String.isWindowsURI() = SystemUtils.IS_OS_WINDOWS && this.startsWith("file://")
+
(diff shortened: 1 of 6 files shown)


src/test/kotlin/io/snyk/plugin/UtilsKtTest.kt
```diff
@@ -0,0 +1,77 @@
...
+        var virtualFile = mockk<VirtualFile>()
+        every { virtualFile.url } returns uri
+
+        assertEquals("file:///$path", virtualFile.toLanguageServerURL())
+
+        uri = "file://$path"
+        virtualFile = mockk<VirtualFile>()
+        every { virtualFile.url } returns uri
+
+        assertEquals("file:///$path", virtualFile.toLanguageServerURL())
```
teodora-sandu, line 43: I might be missing something but is there a difference between this test and the one before? Did you mean to set `uri` to `file:///$path` instead?

src/main/kotlin/io/snyk/plugin/Utils.kt
```diff
@@ -430,10 +430,30 @@ fun String.toVirtualFile(): VirtualFile {
...
+fun String.toVirtualFileURL(): String {
+    if (this.isWindowsURI() && this.startsWith("file:///") && this.length > 10 && this.substring(9,10) == ":") {
+        return this.replaceFirst("/","")
+    }
+    return this
+}
+
+fun String.isWindowsURI() = SystemUtils.IS_OS_WINDOWS && this.startsWith("file://")
+
+fun VirtualFile.urlContainsDriveLetter() = this.url.isWindowsURI() && this.url.length > 9 && this.url.substring(8,9) == ":"
```
teodora-sandu, line 455: nit: It might be nicer to do this using a regex. I really had to think about what this function was doing, although the tests did help. Just a personal preference though

src/test/kotlin/io/snyk/plugin/UtilsKtTest.kt
bastiandoetsch: Yeah, a slash is missing 🙈

src/main/kotlin/io/snyk/plugin/Utils.kt
bastiandoetsch: Hm... I'm so-so. For the last part it would definitely be easier with a regexp. At least if you understand/know regular expressions ;).

bastiandoetsch: Would be something like this: `file:\/\/(\/)?[a-zA-Z]:.*$`. Do you think that's easier to understand?

bastiandoetsch: image

teodora-sandu: We agreed on a call to keep it as it was! Turns out the regex is no easier 😆

teodora-sandu (approved):

PR #511, opened 2024-04-16
fix: settings for scan types

teodora-sandu:
### Description

At the moment when we change the scan type selection the Tree Nodes associated with them don't get re-enabled. This seems to be because the changes in the `ScanTypesPanel` don't actually get trickled up to `SnykProjectSettingsConfigurable` in order for the `apply()` method to run when `isModified()` returns `true`. The reason for this is that when the selections change the `Dialog.isModified()` function called in `isScanTypeChanged` in `SnykSettingsDialog` hasn't registered that there has been a change. We're now achieving that with the use of `bindSelected`.

We have some validation in place so that pre-apply we don't allow people to deselect all scan types. The new selections only get "persisted" globally once the `apply()` function gets called so the new action listener we added for each checkbox also needs to locally persist the selection changes in order for the validation to work.

Docs about these panels that led to this implementation:
- https://plugins.jetbrains.com/docs/intellij/kotlin-ui-dsl.html#integrating-panels-with-property-bindings
- https://plugins.jetbrains.com/docs/intellij/kotlin-ui-dsl-version-2.html#cellbind

There are two more panels that probably have this issue. The `SeveritiesEnablementPanel` and `IssueViewOptionsPanel`. Before I make those changes there I want to see if this approach is okay.

### Checklist

- [x] Linted
- [ ] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Snyk Security Changelog
 
+## [2.7.15]
+### Fixed
+- Re-enable scan results when re-enabling different scan types
 
 ## [2.7.14]
 ### Added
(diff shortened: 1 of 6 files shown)


bastiandoetsch (approved):
Nice finding, didn't know `bindSelected` before.

PR #514, opened 2024-04-17
feat: implement theme-based dynamic styles for Code Issue Panel [IDE-250]

cat2608:
### Description

- Dynamically sets CSS properties via `browser.executeJavaScript` based on the current IDE theme (dark, darcula or light).
- Visual elements rendered from the Language Server, such as text color, background color, link color, and other UI components, are consistent with the user's theme preference.

❗ Related: https://github.com/snyk/snyk-ls/pull/491 where support for dynamic dark and light themes was added.

### Checklist

- [ ] Linted
- [x] CHANGELOG.md updated

### Screenshots / GIFs

|🌒 Dark|🧛‍♀️ Darcula|☀️ Light|
|-|-|-|
|dark-theme|darcula-theme|light-theme


src/main/kotlin/io/snyk/plugin/ui/jcef/OpenFileLoadHandlerGenerator.kt
```diff
@@ -35,32 +35,76 @@ class OpenFileLoadHandlerGenerator(snykCodeFile: SnykCodeFile) {
...
+                        document.documentElement.style.setProperty('--tab-item-icon-color', style.getPropertyValue('--tab-item-icon-color-${if (isDarkTheme) "dark" else "light"}'));
+                        document.documentElement.style.setProperty('--scrollbar-thumb-color', style.getPropertyValue('--scrollbar-thumb-color-${if (isDarkTheme) "dark" else "light"}'));
+                        document.documentElement.style.setProperty('--scrollbar-color', style.getPropertyValue('--scrollbar-color-${if (isDarkTheme) "dark" else "light"}'));
+
+                        if (${isDarculaTheme}) {
+                            document.documentElement.style.setProperty('--data-flow-body-color', style.getPropertyValue('--data-flow-body-color-darcula'));
+                            document.documentElement.style.setProperty('--example-line-added-color', style.getPropertyValue('--example-line-added-color-darcula'));
+                            document.documentElement.style.setProperty('--example-line-removed-color', style.getPropertyValue('--example-line-removed-color-darcula'));
+                        }
+                        })();
```
cat2608, line 105: I'd like to find a better way to handle this maybe by creating a mapping of theme-related properties that can be iterated over 🤔

src/main/kotlin/io/snyk/plugin/ui/jcef/OpenFileLoadHandlerGenerator.kt
```diff
@@ -35,32 +35,76 @@ class OpenFileLoadHandlerGenerator(snykCodeFile: SnykCodeFile) {
...
+                                e.preventDefault();
+                                window.openFileQuery(target.getAttribute("file-path") + ":" +
+                                                     target.getAttribute("start-line") + ":" +
+                                                     target.getAttribute("end-line") + ":" +
+                                                     target.getAttribute("start-character") + ":" +
+                                                     target.getAttribute("end-character"));
+                            }
+                        });
+                    })();
+                """
```
cat2608, line 70: - Use event delegation for the `.data-flow-clickable-row` elements. This change reduces the
  number of event listeners attached to the DOM by binding a single listener on
  the document rather than individual listeners on each element.
- Prevent redefinition of the `openFileQuery` function initialising just once per page load.

src/main/kotlin/io/snyk/plugin/ui/jcef/OpenFileLoadHandlerGenerator.kt
cat2608: Also, I'm still investigating how to move this logic to the Language Server template where we'd be identifying the IDE via a variable sent by the IDE. For IntelliJ, it would be something like:

```kotlin
browser.executeJavaScript("window.hostIDE = 'jetbrains';", browser.url, 0)
```

So we will be handling theme specific for each IDE since VSCode needs custom properties like `--vscode-editor-foreground`

Still WIP 🏗️

bastiandoetsch: If it's IDE specific - does it need to be moved to LS? If it's easier to code & maintain, sure. But IDE-specific stuff could easily stay in the IDE.

bastiandoetsch: Regarding the `if (isDarktheme)...` I'd just define a variable `theme` that you can add, so you don't clutter the styles with inline `if` clauses.

bastiandoetsch: Hm... for custom style properties (e.g. --vscode-editor-foreground), could they just be template variables to be substituted in the IDE?

CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 # Snyk Security Changelog
 
 
+## [2.7.15]
```
bastiandoetsch, line 4: 2.7.16 is the new 2.7.15 ;)

src/main/kotlin/io/snyk/plugin/ui/jcef/OpenFileLoadHandlerGenerator.kt
```diff
@@ -35,32 +35,76 @@ class OpenFileLoadHandlerGenerator(snykCodeFile: SnykCodeFile) {
         return JBCefJSQuery.Response("success")
     }
 
+    fun isDarcula(): Boolean {
+        val scheme = EditorColorsManager.getInstance().globalScheme
+        return scheme.name.contains("Darcula")
+    }
+
```
bastiandoetsch, line 42: Why do we need to specify specific themes? There are thousands on the marketplace, is there a chance to have generic handling?

src/main/kotlin/io/snyk/plugin/ui/jcef/OpenFileLoadHandlerGenerator.kt
```diff
@@ -35,32 +39,78 @@ class OpenFileLoadHandlerGenerator(snykCodeFile: SnykCodeFile) {
...
+                          '--example-line-added-color',
+                          '--example-line-removed-color',
+                          '--tabs-bottom-color',
+                          '--tab-item-color',
+                          '--tab-item-hover-color',
+                          '--tab-item-icon-color',
+                          '--scrollbar-thumb-color',
+                          '--scrollbar-color',
+                        ]
+                        properties.forEach(p => document.documentElement.style.setProperty(p, style.getPropertyValue(p + "-" + "${colorTheme}")))
```
cat2608, line 104: @bastiandoetsch about your comment:

>Regarding the if (isDarktheme)... I'd just define a variable theme that you can add, so you don't clutter the styles with inline if clauses.

Have a look at the following refactor, please and let me know what you thing :)

src/main/kotlin/io/snyk/plugin/ui/jcef/OpenFileLoadHandlerGenerator.kt
cat2608: I'll get back to you in a moment 🙏

src/main/kotlin/io/snyk/plugin/ui/jcef/OpenFileLoadHandlerGenerator.kt
bastiandoetsch: Yeah, much better. Still hoping for a more generic solution though. I feel uneasy having specific themes in our extension design - would expect that IntelliJ provides all styles from the theme, and we can just reference them (color, font, etc).

cat2608: Sorry @bastiandoetsch , maybe I'm not understanding what you mean. The development so far when moving the HTML and CSS to LSP is like "server-side rendering":

- The initial template and styles are prepared before they reach the client (the IDE in this case)
- Once delivered, any dynamic adjustments based on the IDE settings or themes are handled client-side

Then, once the page is loaded in the IDE, IntelliJ in this case, we still need to apply the [JetBrains typography](https://jetbrains.design/intellij/principles/typography/#colors) from the Kotlin code with `EditorColorsManager.getInstance().globalScheme`. We dynamically update the CSS variables in the HTML to match these styles.

Do you have any other suggestion? Maybe I missed exploring other alternatives 🤔

src/main/kotlin/io/snyk/plugin/ui/jcef/OpenFileLoadHandlerGenerator.kt
cat2608: @neilm-snyk , there is a follow-up ticket to address the styles from JetBrains APIs, so I believe it will automatically pick up the colours, fonts, and so on. Thank you 🙏

PR #523, opened 2024-04-29
fix: remove incorrect /v1 path [IDE-126]

teodora-sandu:
### Description

In https://github.com/snyk/snyk-intellij-plugin/pull/515/ the `/v1` was appended in IntelliJ for all calls, even though there are some that don't need it. Only old v1 APIs need this path, otherwise we should leave it out. We noticed this problem by having the `snykCodeConsistentIgnores` feature flag enabled, which enables a flow that calls some other APIs.

In order for this to not introduce a new bug, we need https://github.com/snyk/snyk-ls/pull/498 to get merged first.

### Checklist

- [x] Linted
- [x] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [2.7.17]
 ### Fixed
 - Fixed problem in re-enablement of scan types when only one scan type was selected
+- Don't add /v1 to all API calls through the Language Server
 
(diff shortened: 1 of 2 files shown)


cat2608 (approved):
Tested it locally and 🚀 
Thank you!

PR #524, opened 2024-04-29
fix: custom endpoint validation and default [IDE-131]

teodora-sandu:
### Description

Improves the validation so that only `http(s)://api.` APIs are configured in the custom endpoint field.

Tested manually by running the extension with the latest CLI.

### Checklist

- [x] Linted
- [x] CHANGELOG.md updated

### Screenshots / GIFs
Screenshot 2024-05-01 at 09 42 50
Screenshot 2024-05-01 at 09 42 42
Screenshot 2024-05-01 at 09 42 38
Screenshot 2024-05-01 at 16 58 50

### DIFF
--- a/src/main/kotlin/snyk/common/CustomEndpoints.kt
+++ b/src/main/kotlin/snyk/common/CustomEndpoints.kt
@@ -100,8 +100,7 @@ internal fun resolveCustomEndpoint(endpointUrl: String?): String {
 
 fun URI.isSnykTenant() =
     isSnykDomain()
-        && ((host.lowercase().startsWith("app.") && path.lowercase().endsWith("/api"))
-        || (host.lowercase() == "snyk.io" && path.lowercase().endsWith("/api"))
+        && ((host.lowercase() == "snyk.io" && path.lowercase().endsWith("/api"))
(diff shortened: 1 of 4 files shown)


src/main/kotlin/snyk/common/CustomEndpoints.kt
```diff
@@ -100,8 +100,7 @@ internal fun resolveCustomEndpoint(endpointUrl: String?): String {
 
 fun URI.isSnykTenant() =
     isSnykDomain()
-        && ((host.lowercase().startsWith("app.") && path.lowercase().endsWith("/api"))
-        || (host.lowercase() == "snyk.io" && path.lowercase().endsWith("/api"))
+        && ((host.lowercase() == "snyk.io" && path.lowercase().endsWith("/api"))
```
teodora-sandu, line 103: The ticket mentions that we should not use `app.snyk.io/api` but nothing about `snyk.io/api`. This is a legacy way of calling APIs, as far as I know, which works today with `api.snyk.io` too. I think we could improve the validation some more so `snyk.io/api` is not allowed but it's not something we were asked to do and I don't know for sure that this wouldn't break something else so I'd rather leave it out

cat2608 (approved):
Tested it locally and the field is working fine :)
Thank you!

PR #525, opened 2024-05-08
feat: load Snyk OSS issues in the Tree view via LS [IDE-238]

teodora-sandu:
### Description

This PR loads the Tree Nodes for OSS from LS. It also renders a simple Suggestion Panel with a title and subtitle, but the follow-up UI elements will be added in a separate PR.

It has been written so that all the product lines loaded via the LS are rendered the same way. So the `ScanIssue` type now exposes functions that are meant to be used in the UI elements to render information about the issue, depending on what product-line it is.

It requires the changes in https://github.com/snyk/snyk-ls/releases/tag/v20240513.094925 to work, so update the CLI locally for now and run `make build`, then change the path to the CLI in the Snyk Settings. Also make sure to enable the `snyk.preview.snyk.oss.ls.enabled` registry flag in IntelliJ.

### Checklist

- [x] Linted
- [ ] CHANGELOG.md updated

### Screenshots / GIFs

Without feature flag:
Screenshot 2024-05-09 at 15 56 25
Screenshot 2024-05-09 at 15 56 19
Screenshot 2024-05-10 at 15 50 38

With feature flag:
Screenshot 2024-05-10 at 16 08 18
Screenshot 2024-05-10 at 16 08 07
Screenshot 2024-05-10 at 16 08 11

### DIFF
--- a/src/main/kotlin/io/snyk/plugin/SnykFile.kt
+++ b/src/main/kotlin/io/snyk/plugin/SnykFile.kt
@@ -4,7 +4,6 @@ import com.intellij.openapi.application.ApplicationManager
 import com.intellij.openapi.project.Project
 import com.intellij.openapi.util.Iconable
 import com.intellij.openapi.vfs.VirtualFile
-import io.snyk.plugin.getPsiFile
 import snyk.common.RelativePathHelper
 import javax.swing.Icon
 
@@ -17,3 +16,6 @@ data class SnykFile(val project: Project, val virtualFile: VirtualFile) {
         }
     }
 }
+
+fun toSnykFileSet(project: Project, virtualFiles: Set<VirtualFile>) =
+    virtualFiles.map { SnykFile(project, it) }.toSet()
--- a/src/main/kotlin/io/snyk/plugin/Utils.kt
+++ b/src/main/kotlin/io/snyk/plugin/Utils.kt
@@ -56,6 +56,7 @@ import snyk.common.SnykCachedResults
 import snyk.common.UIComponentFinder
 import snyk.common.isSnykTenant
 import snyk.common.lsp.ScanInProgressKey
+import snyk.common.lsp.ScanIssue
 import snyk.common.lsp.ScanState
 import snyk.container.ContainerService
 import snyk.container.KubernetesImageCache
@@ -94,6 +95,17 @@ fun getSnykTaskQueueService(project: Project): SnykTaskQueueService? = project.s
 fun getSnykToolWindowPanel(project: Project): SnykToolWindowPanel? = project.serviceIfNotDisposed()
 
 fun getSnykCachedResults(project: Project): SnykCachedResults? = project.serviceIfNotDisposed()
+
+fun getSnykCachedResultsForProduct(project: Project, product: ProductType): MutableMap<SnykFile, List<ScanIssue>>? {
+    return when (product) {
+        ProductType.OSS -> getSnykCachedResults(project)?.currentOSSResultsLS
+        ProductType.IAC -> getSnykCachedResults(project)?.currentIacResultsLS
+        ProductType.CONTAINER -> getSnykCachedResults(project)?.currentContainerResultsLS
+        ProductType.CODE_SECURITY -> getSnykCachedResults(project)?.currentSnykCodeResultsLS
+        ProductType.CODE_QUALITY -> getSnykCachedResults(project)?.currentSnykCodeResultsLS
+        ProductType.ADVISOR -> mutableMapOf()
+    }
+}
 
 fun getAnalyticsScanListener(project: Project): AnalyticsScanListener? = project.serviceIfNotDisposed()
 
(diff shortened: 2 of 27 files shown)


src/main/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowSnykScanListenerLS.kt
```diff
@@ -47,77 +58,108 @@ class SnykToolWindowSnykScanListenerLS(
...
+            rootNode = this.rootOssIssuesTreeNode,
+            ossResultsCount = snykResults.values.flatten().size,
+        )
     }
 
     private fun displayIssues(
-        snykCodeResults: Map<SnykFile, List<ScanIssue>>,
-        selectedNodeUserObject: Any?,
-        isSecurity: Boolean
+        enabledInSettings: Boolean,
```
teodora-sandu, line 127: This function got refactored so it can be called for both Code and OSS results.

src/main/kotlin/io/snyk/plugin/ui/toolwindow/SnykTreeCellRenderer.kt
```diff
@@ -111,24 +112,13 @@ class SnykTreeCellRenderer : ColoredTreeCellRenderer() {
                 val issue = value.userObject as ScanIssue
                 nodeIcon = SnykIcons.getSeverityIcon(issue.getSeverityAsEnum())
                 val range = issue.range
-                var showIgnoredLabel = false
```
teodora-sandu, line 114: Moved most of this logic to the issue definition, so we can contain the product-based switching into one place and not clutter this code

src/main/kotlin/io/snyk/plugin/Utils.kt
```diff
@@ -95,6 +96,17 @@ fun getSnykToolWindowPanel(project: Project): SnykToolWindowPanel? = project.ser
 
 fun getSnykCachedResults(project: Project): SnykCachedResults? = project.serviceIfNotDisposed()
 
+fun getSnykCachedResultsForProduct(project: Project, product: ProductType): MutableMap<SnykFile, List<ScanIssue>>? {
+    return when (product) {
```
teodora-sandu, line 100: Although IaC and Container are not retrieved from the LS, it's easier to write the code in a way that is supported for all product lines. So I've added this here

src/main/kotlin/io/snyk/plugin/services/SnykTaskQueueService.kt
```diff
@@ -132,6 +134,12 @@ class SnykTaskQueueService(val project: Project) {
                 if (settings.ossScanEnable) {
                     if (!isSnykOSSLSEnabled()) {
                         scheduleOssScan()
+                    } else {
+                        // the LS deals with triggering scans at startup
+                        // TODO: Refactor when more than Snyk Code is available in LS for IntelliJ
```
teodora-sandu, line 139: This TODO is here for the future, when everything is via the LS

src/main/kotlin/snyk/common/lsp/Types.kt
```diff
@@ -79,6 +85,185 @@ data class ScanIssue(
         endOffset = document?.getLineStartOffset(range.end.line)?.plus(range.end.character)
     }
 
+    fun type(): IssueType {
```
teodora-sandu, line 88: Have a look at the new functions in this file especially to see if anything doesn't make sense. They're used throughout the UI so hopefully they should

src/main/kotlin/snyk/common/lsp/Types.kt
```diff
@@ -201,14 +386,82 @@ data class IssueData(
     @SerializedName("priorityScore") val priorityScore: Int,
     @SerializedName("hasAIFix") val hasAIFix: Boolean,
     @SerializedName("dataFlow") val dataFlow: List<DataFlow>,
+
+    // OSS
```
teodora-sandu, line 390: I don't know if there's a better way to do this. I tried making the `additionalData` an `Any` but that breaks the gson serializer. I don't know if we can define our own adapter for this ... https://stackoverflow.com/questions/28319473/gson-same-field-name-different-types?

At the moment I'm using a field from each product type to tell us which product it is and so which format

src/main/kotlin/io/snyk/plugin/SnykFile.kt
```diff
@@ -17,3 +16,6 @@ data class SnykFile(val project: Project, val virtualFile: VirtualFile) {
         }
     }
 }
+
+fun toSnykFileSet(project: Project, virtualFiles: Set<VirtualFile>) =
```
teodora-sandu, line 20: Just moving a function from one file to another

src/main/kotlin/snyk/common/lsp/Types.kt
```diff
@@ -236,6 +489,31 @@ data class IssueData(
     }
 
     override fun hashCode(): Int {
+        if (this.getProductType() == ProductType.OSS) {
+            var result = ruleId.hashCode()
+            result = 31 + description.hashCode()
```
cat2608, line 494: I'm not sure if this is the reason but I get this error when I run the plugin:

```
java.lang.NullPointerException: Cannot invoke "String.hashCode()" because "this.ruleId" is null
	at snyk.common.lsp.IssueData.hashCode(Types.kt:493)
	at snyk.common.lsp.ScanIssue.hashCode(Types.kt)
	at java.base/java.util.HashMap.hash(HashMap.java:338)
	at java.base/java.util.HashMap.put(HashMap.java:610)
	at java.base/java.util.HashSet.add(HashSet.java:221)
```
and the difference I see it's the calculation for the hash for `description` is using `+` instead of `*` operator .

cat2608: I think we are missing the `* result`

teodora-sandu: Already discussed this in Slack but this runtime error is caused because https://github.com/snyk/snyk-ls/releases/tag/v20240513.094925 is required :)

cat2608 (approved):

PR #526, opened 2024-05-09
feat: add oss ls registry flag [IDE-283]

teodora-sandu:
### Description

Adds the registry flag to be used while integrating OSS into IntelliJ via LS. Also adds a small "fix" which means once we enable the LS we also check that OSS is enabled from the settings, like we do for Code too.

### Checklist

- [x] Linted
- [x] CHANGELOG.md updated


acke (approved):
Looks good to me.

PR #527, opened 2024-05-09
refactor: rename common classes from SnykCode prefix [IDE-283]

teodora-sandu:
### Description

Renames and moves common "LS in IntelliJ" files that were used when integrating Snyk Code via LS, so that we can use them when we integrated Snyk OSS via LS.

There is not actual change but I have tested it in IntelliJ by running the IDE, just to sanity check that Snyk Code scanning still works.

### Checklist

- [x] Linted
- [x] CHANGELOG.md updated

### Screenshots

Screenshot 2024-05-09 at 14 09 49

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Snyk Security Changelog
 
+## [2.7.19]
+### Added
+- Refactors some files so they can be used by more than just Snyk Code
+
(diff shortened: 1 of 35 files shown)


cat2608 (approved):
Tested it locally and it works fine, thank you 🙌 
image

PR #528, opened 2024-05-10
refactor: snyk code ls common functionality [IDE-283]

teodora-sandu:
### Description

Refactors the `SuggestionDescriptionPanelFromLS` by extracting Snyk Code specific panels out of it. This will make it easier to add the OSS panels when integrating OSS via LS into IntelliJ. Eventually this panel will just load data from HTML so as long as we're making it easy to inject panels in it in the meantime, I don't think we need to improve this code even more. These panels that got extracted were also used in `SuggestionDescriptionPanel` but I haven't de-duplicated those because that panel will disappear soon anyway.

Also refactors `SnykCodeAnnotatorLS` so that common functionality can be used when implementing the OSS Annotator when integrating OSS via LS into IntelliJ.

### Checklist

- [x] Linted
- [ ] CHANGELOG.md updated

### Screenshots / GIFs

Before:
Screenshot 2024-05-10 at 09 12 37

After:
Screenshot 2024-05-10 at 09 14 02


teodora-sandu:
The CLI needs to be updated with https://github.com/snyk/snyk-ls/releases/tag/v20240510.143016 but I think the CLI team is already planning on updating the CLI so will wait a bit

src/main/kotlin/io/snyk/plugin/ui/toolwindow/panels/SnykCodeDataflowPanel.kt
```diff
@@ -0,0 +1,153 @@
...
+
+
+    private fun stepsPanel(dataflow: List<DataFlow>): JPanel {
+        val panel = JPanel()
+        panel.layout = GridLayoutManager(dataflow.size, 1, JBUI.emptyInsets(), 0, 0)
+        panel.background = UIUtil.getTextFieldBackground()
+
+        val maxFilenameLength = dataflow.asSequence()
+            .filter { it.filePath.isNotEmpty() }
+            .map { it.filePath.substringAfterLast('/', "").length }
```
acke, line 46: it.filePath.substringAfterLast('/', "").length 
Is this safe for IntelliJ users on Windows machines? 


On Windows, file paths backslash \ is used as the directory separator instead of the forward slash.

acke (approved):
Looks good to me!

src/main/kotlin/io/snyk/plugin/ui/toolwindow/panels/SnykCodeDataflowPanel.kt
teodora-sandu: I think you might be right, although I haven't tested it since this is code that I just moved from one file to another so I'm not sure. 

For context, this panel here is used for Snyk Code within the integration with LS, which is a new feature that not many customers have seen.

We have also recently migrated this to HTML, so the fix should ideally be done there. We plan on testing this in this cycle for HTML so will leave it to then.

PR #529, opened 2024-05-13
feat: load Snyk OSS issues in the suggestion panel via LS [IDE-238]

teodora-sandu:
### Description

This PR is a continuation of https://github.com/snyk/snyk-intellij-plugin/pull/525 and loads the full OSS content inside the Suggestion Panel. The components were taken from https://github.com/snyk/snyk-intellij-plugin/blob/fa6d6e06028ebf45c3b056fb67b3fbf26b630e41/src/main/kotlin/io/snyk/plugin/ui/toolwindow/panels/VulnerabilityDescriptionPanel.kt. I could have de-duplicated them but we want to remove this file anyway soon and it would require retro-fitting the data types into the `IssueData` type for the LS so a bit more effort than worth it.

It requires the changes in https://github.com/snyk/snyk-ls/releases/tag/v20240513.094925 to work, so update the CLI locally for now and run `make build`, then change the path to the CLI in the Snyk Settings. Also make sure to enable the `snyk.preview.snyk.oss.ls.enabled` registry flag in IntelliJ.

### Checklist

- [x] Linted
- [x] CHANGELOG.md updated

### Screenshots / GIFs
Before:
Screenshot 2024-05-13 at 11 08 26
Screenshot 2024-05-13 at 11 08 17

After:
Screenshot 2024-05-13 at 10 27 03
Screenshot 2024-05-13 at 10 26 48
Screenshot 2024-05-10 at 16 08 18
OSS IN LS

### DIFF
--- a/src/main/kotlin/snyk/common/lsp/Types.kt
+++ b/src/main/kotlin/snyk/common/lsp/Types.kt
@@ -389,7 +389,7 @@ data class IssueData(
 
     // OSS
     @SerializedName("license") val license: String?,
-    @SerializedName("identifiers") val identifiers: OssIdentifiers?,
+    @SerializedName("identifiers") val identifiers: OssIdentifiers,
     @SerializedName("description") val description: String,
     @SerializedName("language") val language: String,
     @SerializedName("packageManager") val packageManager: String,
@@ -406,7 +406,7 @@ data class IssueData(
     @SerializedName("isUpgradable") val isUpgradable: Boolean?,
     @SerializedName("projectName") val projectName: String,
     @SerializedName("displayTargetFile") val displayTargetFile: String?,
-    @SerializedName("matchingIssues") val matchingIssues: List<IssueData>?,
+    @SerializedName("matchingIssues") val matchingIssues: List<IssueData>,
     @SerializedName("lesson") val lesson: String?,
 
     // Code and OSS
@@ -569,3 +569,4 @@ data class OssIdentifiers(
         return result
     }
 }
+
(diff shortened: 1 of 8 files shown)


src/main/kotlin/snyk/common/lsp/Types.kt
```diff
@@ -389,7 +389,7 @@ data class IssueData(
 
     // OSS
     @SerializedName("license") val license: String?,
-    @SerializedName("identifiers") val identifiers: OssIdentifiers?,
+    @SerializedName("identifiers") val identifiers: OssIdentifiers,
```
acke, line 392: How do we handle the case where OssIdentifiers are null and do we want the program to throw an exception if the value is null? 

"?": Safe call operator. Allows the program to continue without throwing an exception if the value is null.

acke:
What repository should I use to scan and view results? How would I know if the code changes in this ticket are giving me the result I want? What is the Suggestion Panel?

acke (approved):
Looks great!

src/main/kotlin/snyk/common/lsp/Types.kt
teodora-sandu: I don't remember why I changed this but I went back to see what happens if I add it back (since you are correct that it's safer to use in this case). It seems I can add it back. I feel more comfortable doing so because it has `omitempty` set in LS: https://github.com/snyk/snyk-ls/blob/0e5985de1531451cfcdd4788dfe98c8af79fd326/internal/lsp/message_types.go#L1052

PR #531, opened 2024-05-22
fix: append v1 suffix conditionally

teodora-sandu:
### Description

Currently when using the `stable` CLI version in IntelliJ (with or without the "Code in LS" registry flag enabled) we end up failing to run Snyk Code scans because the call to retrieve the SAST settings requires a `/v1` in the path. This has started since https://github.com/snyk/snyk-intellij-plugin/pull/515. 

In https://github.com/snyk/snyk-intellij-plugin/pull/523 we removed this from being added to all API calls because of some bugs we noticed at the time. The bug at the time was that when `snykCodeConsistentIgnores` is enabled, we make some calls to the new Snyk REST API so we should not add this `/v1` to all calls. In https://github.com/snyk/snyk-ls/pull/498 (requirement for PR#523), we made sure to only add `/v1` to the required calls but this missed the cutoff for the first `stable` release and so is not released to `stable` yet.

Instead of making a hotfix in the CLI, which is expensive, we decided to undo the change in https://github.com/snyk/snyk-intellij-plugin/pull/523. Upon testing though, we noticed a new bug similar to the one when `snykCodeConsistentIgnores` is enabled. When the "Code in LS" registry flag was not enabled (which is what all our customers have now and Casey had actually noticed [here](https://snyk.slack.com/archives/C06B1U3RKG8/p1714665917838529?thread_ts=1714646918.499379&cid=C06B1U3RKG8)) the calls to `deeproxy` were getting the `/v1` appended to them and failing. So this PR also adds a fix for that.

### Checklist

- [ ] Linted
- [x] CHANGELOG.md updated

### Screenshots / GIFs

In `master` with `stable`
- with "Code in LS" and without "Code in LS"
Screenshot 2024-05-22 at 15 10 09

In `master` with `preview`
- with "Code in LS" and without "Code in LS"
Screenshot 2024-05-22 at 15 11 14

In PR with `stable`
- with "Code in LS"
Screenshot 2024-05-22 at 16 00 58

- without "Code in LS"

In PR with `preview`
- with "Code in LS"
Screenshot 2024-05-22 at 16 02 51

- without "Code in LS"
Screenshot 2024-05-22 at 16 05 35

### DIFF
--- a/src/main/kotlin/snyk/common/CustomEndpoints.kt
+++ b/src/main/kotlin/snyk/common/CustomEndpoints.kt
@@ -73,7 +73,8 @@ fun getEndpointUrl(): String {
         ""
     }
     val customEndpointUrl = resolveCustomEndpoint(endpointUrl)
-    return customEndpointUrl.removeTrailingSlashesIfPresent()
+    // we need to set v1 here, to make the sast-enabled calls work in LS
+    return customEndpointUrl.removeTrailingSlashesIfPresent().suffixIfNot("/v1")
 }
 
 fun isSnykCodeAvailable(endpointUrl: String?): Boolean {
--- a/src/main/kotlin/snyk/common/CustomEndpoints.kt
+++ b/src/main/kotlin/snyk/common/CustomEndpoints.kt
@@ -21,12 +21,14 @@ fun toSnykCodeApiUrl(endpointUrl: String?): String {
         uri.isDev() ->
             endpoint
                 .replace("api.", "")
+                .replace("/v1", "")
                 .replace("https://dev.", "https://$codeSubdomain.dev.")
                 .suffixIfNot("/")
 
         uri.isSnykTenant() ->
             endpoint
                 .replace("https://api.", "https://")
+                .replace("/v1", "")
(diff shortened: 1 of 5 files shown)


teodora-sandu:
IaC and Container failing but it's failing with the CLI too for me so I don't think it's related, must be something in my setup since I've seen these fail for weeks:
Screenshot 2024-05-22 at 16 47 48
Screenshot 2024-05-22 at 16 48 16
Screenshot 2024-05-22 at 16 48 09

src/main/kotlin/snyk/common/CustomEndpoints.kt
```diff
@@ -21,12 +21,14 @@ fun toSnykCodeApiUrl(endpointUrl: String?): String {
         uri.isDev() ->
             endpoint
                 .replace("api.", "")
+                .replace("/v1", "")
```
bastiandoetsch, line 24: These changes always make me anxious without unit tests. Could you please add some?

bastiandoetsch (approved):

PR #535, opened 2024-05-29
feat: inject custom styling [IDE-240]

teodora-sandu:
### Description

Extracts custom IntelliJ styling from `snyk-ls` and injects it into the HTML after compiling `.scss` file to `.css` using https://github.com/EtienneMiret/sass-gradle-plugin. This requires https://github.com/snyk/snyk-ls/pull/522 so that `${ideStyle}` is available.

To test:
- Run `make build-debug` in `snyk-ls` in the branch from https://github.com/snyk/snyk-ls/pull/522
- Run `make build` in the CLI after adding a replace directive for `snyk-ls`
- Run the sandbox using `runIde`
- Replace the CLI path in the sandbox to the path for the local CLI and disable automatically managing the binaries
- Make sure to use an org that has the `snykCodeConsistentIgnores` FF enabled (e.g. `ide-consistent-ignores-test`)
- Navigate to a Snyk Code vulnerability

### Checklist

- [x] Linted
- [x] CHANGELOG.md updated

### Screenshots / GIFs

Without the LS change:
Screenshot 2024-05-29 at 15 08 42

With the LS change:

Screenshot 2024-05-29 at 15 01 02

### DIFF
--- a/src/main/kotlin/snyk/common/lsp/LanguageServerWrapper.kt
+++ b/src/main/kotlin/snyk/common/lsp/LanguageServerWrapper.kt
@@ -107,7 +107,7 @@ class LanguageServerWrapper(
         try {
             val snykLanguageClient = SnykLanguageClient()
             languageClient = snykLanguageClient
-            val logLevel = if (snykLanguageClient.logger.isDebugEnabled) "trace" else "info"
+            val logLevel = if (snykLanguageClient.logger.isDebugEnabled) "debug" else "info"
             val cmd = listOf(lsPath, "language-server", "-l", logLevel)
 
             val processBuilder = ProcessBuilder(cmd)
--- a/src/main/kotlin/snyk/common/lsp/LanguageServerWrapper.kt
+++ b/src/main/kotlin/snyk/common/lsp/LanguageServerWrapper.kt
@@ -107,7 +107,7 @@ class LanguageServerWrapper(
         try {
             val snykLanguageClient = SnykLanguageClient()
             languageClient = snykLanguageClient
-            val logLevel = if (snykLanguageClient.logger.isDebugEnabled) "debug" else "info"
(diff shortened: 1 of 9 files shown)


src/main/resources/stylesheets/snyk_code_suggestion.scss
```diff
@@ -0,0 +1,80 @@
...
+  font-family: "Courier New", Courier, monospace;
+}
+
+.example-line.added {
+  background-color: var(--example-line-added-color);
+}
+
+.example-line.removed {
+  background-color: var(--example-line-removed-color);
+}
```
cat2608, line 80: Why did we need to use a SASS file? I thought SASS is needed when we need things like nesting or mixins or other SASS [magic](https://sass-lang.com/guide/) ✨

teodora-sandu: Mostly because we use SASS in VSCode so I wanted to be as consistent as possible across the two, but also because there are more features available in SASS (like you said) and in the future we will have more than Snyk Code styling but still want to share stylesheets or colours across Snyk OSS, Snyk IaC, and Snyk Container

src/main/kotlin/snyk/common/lsp/LanguageServerWrapper.kt
```diff
@@ -108,7 +108,7 @@ class LanguageServerWrapper(
         try {
             val snykLanguageClient = SnykLanguageClient()
             languageClient = snykLanguageClient
-            val logLevel = if (snykLanguageClient.logger.isDebugEnabled) "debug" else "info"
+            val logLevel = if (snykLanguageClient.logger.isDebugEnabled) "trace" else "info"
```
teodora-sandu, line 111: Remove this !!

bastiandoetsch (approved):

PR #542, opened 2024-06-11
fix: make plugin compatible to 2024.2, reduce time in UI Thread [IDE-395]

bastiandoetsch:
### Description

- enable 2024.2 in properties
- reduce time spent in UI thread by optimizing file listener

### Checklist

- [x] Linted
- [x] CHANGELOG.md updated

### Screenshots / GIFs

### DIFF
--- a/src/main/kotlin/io/snyk/plugin/snykcode/SnykCodeBulkFileListener.kt
+++ b/src/main/kotlin/io/snyk/plugin/snykcode/SnykCodeBulkFileListener.kt
@@ -1,7 +1,9 @@
 package io.snyk.plugin.snykcode
 
+import com.google.common.cache.CacheBuilder
 import com.intellij.codeInsight.daemon.DaemonCodeAnalyzer
 import com.intellij.openapi.application.runInEdt
+import com.intellij.openapi.diagnostic.logger
 import com.intellij.openapi.progress.ProgressIndicator
 import com.intellij.openapi.progress.ProgressManager
 import com.intellij.openapi.progress.Task.Backgroundable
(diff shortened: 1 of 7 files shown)


src/main/kotlin/io/snyk/plugin/snykcode/SnykCodeBulkFileListener.kt
```diff
@@ -11,51 +13,66 @@ import com.intellij.openapi.vfs.VirtualFileManager
...
 class SnykCodeBulkFileListener : SnykBulkFileListener() {
+    // Cache for debouncing file updates that come in within one second of the last
+    // Key = path, Value is irrelevant
+    private val debounceFileCache =
+        CacheBuilder.newBuilder()
+            .expireAfterWrite(Duration.ofMillis(1000)).build<String, Boolean>()
+
     override fun before(project: Project, virtualFilesAffected: Set<VirtualFile>) = Unit
 
     override fun after(project: Project, virtualFilesAffected: Set<VirtualFile>) {
```
teodora-sandu, line 33: Do we need similar changes in https://github.com/snyk/snyk-intellij-plugin/blob/1b78014fbf3f8810e35ef7ce7d48e5a63ffbbf02/src/main/kotlin/snyk/oss/OssBulkFileListener.kt?

teodora-sandu: I think it would be good to have some re-usable code for this but I think we can leave that for when we finish the OSS in LS pitch next cycle right?

bastiandoetsch: Yeah, I usually only do this kind of thing, the second time we need it :)

teodora-sandu (approved):

PR #549, opened 2024-06-14
fix: token validation & endpoint config for oauth2 [IDE-458]

bastiandoetsch:
### Description

### Checklist

- [ ] Linted
- [ ] CHANGELOG.md updated

### Screenshots / GIFs

image


teodora-sandu (commented):
The only thing I'd say is maybe it'd be good if we could render the text on the next line rather than have to scroll right. This works too but it's not the nicest user experience
Screenshot 2024-07-11 at 14 45 34

PR #550, opened 2024-06-14
fix: dataflow links work across files [IDE-391]

teodora-sandu:
### Description

The change in this PR is to load the virtual file from the `filePath` in the component instead of using the provided virtual file. This way we can navigate to lines in other files that are not the originating file of the issue.

There are some linting changes too, after configuring the default linter in IntelliJ to format on save.

### Checklist

- [x] Linted
- [x] CHANGELOG.md updated

### Screenshots / GIFs

https://github.com/snyk/snyk-intellij-plugin/assets/81559517/042ca443-238c-4be2-871b-bf9d64896f2c

### DIFF
--- a/src/main/kotlin/io/snyk/plugin/ui/jcef/OpenFileLoadHandlerGenerator.kt
+++ b/src/main/kotlin/io/snyk/plugin/ui/jcef/OpenFileLoadHandlerGenerator.kt
@@ -3,28 +3,32 @@ package io.snyk.plugin.ui.jcef
 import com.intellij.openapi.application.ApplicationManager
 import com.intellij.openapi.editor.colors.ColorKey
 import com.intellij.openapi.editor.colors.EditorColorsManager
+import com.intellij.openapi.project.Project
+import com.intellij.openapi.vfs.VirtualFileManager
 import com.intellij.ui.JBColor
 import com.intellij.ui.jcef.JBCefBrowserBase
 import com.intellij.ui.jcef.JBCefJSQuery
 import com.intellij.util.ui.UIUtil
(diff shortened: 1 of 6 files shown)


src/main/kotlin/io/snyk/plugin/ui/jcef/OpenFileLoadHandlerGenerator.kt
```diff
@@ -3,29 +3,33 @@ package io.snyk.plugin.ui.jcef
...
+        val filePath = values[0]
         val startLine = values[1].toInt()
         val endLine = values[2].toInt()
         val startCharacter = values[3].toInt()
         val endCharacter = values[4].toInt()
 
+        val virtualFile =
+            VirtualFileManager.getInstance().findFileByNioPath(Paths.get(filePath))
+                ?: return JBCefJSQuery.Response("error")
+
```
bastiandoetsch, line 32: Is there a chance to pre-calculate the VirtualFile? IntelliJ doesn't like to do this kind of thing in the UI thread and will log a warning/error if it takes too long.

bastiandoetsch (approved):

src/main/kotlin/io/snyk/plugin/ui/jcef/OpenFileLoadHandlerGenerator.kt
teodora-sandu: Good suggestion, I think I did that in https://github.com/snyk/snyk-intellij-plugin/pull/550/commits/cbe985923cc1ac75f2ab25b0f2673bf275428436. Basically just pre-computing the virtual files before building the query used by the UI

PR #552, opened 2024-06-19
fix: colours, font family, and font size for consistent ignores [IDE-386]

teodora-sandu:
### Description

Changes some colours in order to match the [designs](https://www.figma.com/design/hcI2QHUtHfcIjgrpYlMqff/Holistic-Ignores?node-id=2292-61688&t=larzTCeHFfRtz6av-0)

Make sure to run https://github.com/snyk/snyk-ls/pull/552 first if you want to call out any inconsistent UI elements, in case they're already addressed there.

### Checklist

- [x] Linted
- [x] CHANGELOG.md updated

### DIFF
--- a/src/main/resources/stylesheets/snyk_code_suggestion.scss
+++ b/src/main/resources/stylesheets/snyk_code_suggestion.scss
@@ -26,6 +26,8 @@ a,
   color: var(--link-color);
 }
 
+// TODO: remove ignore styling at the end of cycle 5 2025
+// TODO: keep the font-size
 .ignore-warning {
   background: #FFF4ED;
   color: #B6540B;
--- a/src/main/resources/stylesheets/snyk_code_suggestion.scss
+++ b/src/main/resources/stylesheets/snyk_code_suggestion.scss
@@ -44,13 +44,6 @@ a,
   border-top: 1px solid var(--border-color)
 }
 
-/* Need to override h2 settings for the markdown*/
-h2 {
-  line-height: 1.25rem;
-  font-size: 1.4rem;
-  margin-bottom: 10px;
-}
-
 .issue-overview-header,
 .ignore-details-header,
 .data-flow-header,
@@ -66,7 +59,7 @@ h2 {
 
 .data-flow-text {
   font-family: var(--editor-font);
-  font-size: 1.1rem;
+  font-size: 1.05rem;
 }
 
 .data-flow-clickable-row {
@@ -99,7 +92,7 @@ h2 {
 
 .example-line > code {
   font-family: var(--editor-font);
-  font-size: 1.1rem;
+  font-size: 1.05rem;
 }
 
 .example-line.added {
--- a/src/main/resources/stylesheets/snyk_code_suggestion.scss
+++ b/src/main/resources/stylesheets/snyk_code_suggestion.scss
@@ -55,7 +55,8 @@ h2 {
 .ignore-details-header,
 .data-flow-header,
 .ai-fix-header,
-.example-fixes-header {
+.example-fixes-header,
+.overview-text > h2 {
   font-size: 0.85rem;
 }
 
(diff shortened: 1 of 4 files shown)


src/main/resources/stylesheets/snyk_code_suggestion.scss
```diff
@@ -26,24 +26,25 @@ a,
   color: var(--link-color);
 }
 
-.delimiter {
-  border: 1px solid #505254;
-}
-
 .ignore-warning {
-  background: #FFF4ED;
```
teodora-sandu, line 34: The colours are the same across IDEs so I just moved it to `snyk-ls`.

src/main/resources/stylesheets/snyk_code_suggestion.scss
```diff
@@ -84,6 +81,14 @@ a,
   border-bottom: 3px solid #3474f0;
 }
 
+.example-line {
+  background-color: var(--editor-color);
+}
+
+.example-line > code {
+  font-family: "Courier New", Courier, monospace;
```
teodora-sandu, line 89: I tried to get this from the libraries in `Utils.kt` but we don't get the CSS name and it seems like monospace is what Jetbrains uses by default so I didn't bother writing a conversion function from the JB name for the font to the CSS one

bastiandoetsch: According to Perplexity AI:

To get the configured editor font in the IntelliJ SDK, you can use the `EditorColorsManager` and `EditorColorsScheme` classes. Here's an example:

```java
import com.intellij.openapi.editor.colors.EditorColorsManager;
import com.intellij.openapi.editor.colors.EditorColorsScheme;
import com.intellij.openapi.editor.colors.FontPreferences;

// Get the editor colors manager
EditorColorsManager editorColorsManager = EditorColorsManager.getInstance();

// Get the current editor color scheme
EditorColorsScheme editorColorsScheme = editorColorsManager.getGlobalScheme();

// Get the font preferences from the color scheme
FontPreferences fontPreferences = editorColorsScheme.getFontPreferences();

// Get the configured editor font
String fontFamily = fontPreferences.getFontFamily();
int fontSize = fontPreferences.getSize(null); // Pass null for the default font type
```

The `fontFamily` string contains the name of the configured font family, and `fontSize` contains the configured font size in points[1][2].

If you need to get the font for a specific language or file type, you can use the `EditorColorsScheme.getFont()` method and pass the appropriate language or file type as an argument[2].

Additionally, the `EditorColorsManager` class provides methods to get the global color scheme, as well as to add and remove color schemes[2].

Citations:
[1] https://stackoverflow.com/questions/8691088/change-font-into-intellij-10-11
[2] https://www.jetbrains.com/help/idea/configuring-colors-and-fonts.html
[3] https://www.jetbrains.com/help/idea/settings-editor-font.html
[4] https://askubuntu.com/questions/454361/font-in-intellij-idea-on-ubuntu-14-04
[5] https://plugins.jetbrains.com/docs/intellij/working-with-text.html

bastiandoetsch (approved):

src/main/resources/stylesheets/snyk_code_suggestion.scss
```diff
@@ -84,6 +80,15 @@ a,
...
 }
 
+.example-line {
+  background-color: var(--editor-color);
+}
+
+.example-line > code {
+  font-family: var(--editor-font);
+  font-size: 1.05rem;
+}
```
cat2608, line 90: I think this needs to be `monospace`. It's picking up `"JetBrains Mono"`

teodora-sandu: I meant for it to be `JetBrains Mono`, why do you think it should be `monospace`? 
- https://www.jetbrains.com/help/idea/settings-editor-font.html
- https://www.w3schools.com/whatis/tryit.asp?font=JetBrains%20Mono

PR #570, opened 2024-07-10
feat: load html for oss in ls [IDE-285]

teodora-sandu:
### Description

Adds some styling so that the new HTML panel and the native panel look more or less the same. The only noticeable differences are:
- The text in the `Detailed paths` changed from `Fix` to `Remediation`
- The font for the code, which now matches the one chosen by the customer
- The labels don't have a colon anymore, which matches what we see in VSCode too (could add the colons back if needed via CSS)

Refactors the code a bit too so that we can re-use more of it for both Snyk Code and Snyk OSS.

Needs https://github.com/snyk/cli/pull/5356 for testing.

### Checklist

- [x] Linted
- [ ] CHANGELOG.md updated

### Screenshots / GIFs
Native:
goal 1
goal 2

HTML:
Screenshot 2024-07-10 at 13 24 10
Screenshot 2024-07-10 at 13 24 21

### DIFF
--- a/src/main/kotlin/io/snyk/plugin/ui/jcef/ThemeBasedStylingGenerator.kt
+++ b/src/main/kotlin/io/snyk/plugin/ui/jcef/ThemeBasedStylingGenerator.kt
@@ -1,0 +1,119 @@
+package io.snyk.plugin.ui.jcef
+
+import com.intellij.openapi.editor.colors.ColorKey
+import com.intellij.openapi.editor.colors.EditorColorsManager
+import com.intellij.openapi.editor.colors.EditorColorsScheme
+import com.intellij.openapi.editor.colors.FontPreferences
+import com.intellij.ui.jcef.JBCefBrowserBase
+import com.intellij.util.ui.JBUI
+import com.intellij.util.ui.UIUtil
+import org.cef.browser.CefBrowser
+import org.cef.browser.CefFrame
+import org.cef.handler.CefLoadHandlerAdapter
+import java.awt.Color
+
+class ThemeBasedStylingGenerator {
+    fun toCssHex(color: Color): String {
+        return "#%02x%02x%02x".format(color.red, color.green, color.blue)
+    }
+
+    fun shift(
+        colorComponent: Int,
+        d: Double,
+    ): Int {
+        val n = (colorComponent * d).toInt()
+        return n.coerceIn(0, 255)
+    }
+
+    fun getCodeDiffColors(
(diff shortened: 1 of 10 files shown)


src/main/kotlin/io/snyk/plugin/ui/jcef/ThemeBasedStylingGenerator.kt
```diff
@@ -0,0 +1,119 @@
...
+            } else {
+                Color(shift(baseColor.red, 1.25), shift(baseColor.green, 0.85), shift(baseColor.blue, 0.85))
+            }
+        return Pair(addedColor, removedColor)
+    }
+
+    fun generate(jbCefBrowser: JBCefBrowserBase): CefLoadHandlerAdapter {
+        val isDarkTheme = EditorColorsManager.getInstance().isDarkEditor
+        val isHighContrast =
+            EditorColorsManager.getInstance().globalScheme.name.contains("High contrast", ignoreCase = true)
```
bastiandoetsch, line 51: Wondering: should we also set the other font? It looks like the font is different in "native" and "html", and that maybe is due to not taking the theme font?

src/main/kotlin/io/snyk/plugin/ui/jcef/Utils.kt
```diff
@@ -6,24 +6,27 @@ import com.intellij.ui.jcef.JBCefBrowserBuilder
...
+        loadHandlerGenerators: List<LoadHandlerGenerator>,
     ): Component? {
         if (!JBCefApp.isSupported()) {
             return null
         }
         val cefClient = JBCefApp.getInstance().createClient()
         cefClient.setProperty("JS_QUERY_POOL_SIZE", 1)
         val jbCefBrowser =
-            JBCefBrowserBuilder().setClient(cefClient).setEnableOpenDevToolsMenuItem(false)
+            JBCefBrowserBuilder().setClient(cefClient).setEnableOpenDevToolsMenuItem(true)
```
bastiandoetsch, line 22: Guess, you'll want to switch this back?

src/main/kotlin/io/snyk/plugin/ui/jcef/ThemeBasedStylingGenerator.kt
teodora-sandu: We could, but what would be the point of it? Is it to make the change more gradual once we enable this feature?

PR #578, opened 2024-08-18
fix: `font-family` and `font-size` are set from Language Server

cat2608:
### Description

This PR fixes the `font-size`and `font-family` accordantly to Snyk Design System to keep the look and feel of products and experiences consistent.

> [!NOTE]  
> This change needs to be tested together with the Language Server https://github.com/snyk/snyk-ls/pull/627

### Checklist

- [ ] Linted
- [ ] CHANGELOG.md updated

### Screenshots / GIFs

|Before|After|
|-|-|
before-oss|after-oss|
|before-code|after-code|
before-oss-dark-theme|after-oss-dark-theme|

### DIFF
--- a/src/main/kotlin/io/snyk/plugin/ui/jcef/ThemeBasedStylingGenerator.kt
+++ b/src/main/kotlin/io/snyk/plugin/ui/jcef/ThemeBasedStylingGenerator.kt
@@ -74,10 +74,7 @@ class ThemeBasedStylingGenerator {
                     val tabItemHoverColor =
                         globalScheme.getColor(ColorKey.find("INDENT_GUIDE")) // The closest color to target_rgb = RGB (235, 236, 240)
 
-                    val editorColorsManager = EditorColorsManager.getInstance()
-                    val editorColorsScheme: EditorColorsScheme = editorColorsManager.globalScheme
-                    val fontPreferences: FontPreferences = editorColorsScheme.fontPreferences
-                    val editorFont = fontPreferences.fontFamily
+                    val preBackgroundColor = if (isDarkTheme) "#000000" else "#ffffff"
 
                     val themeScript = """
                         (function(){
@@ -98,6 +95,7 @@ class ThemeBasedStylingGenerator {
                                 '--tabs-bottom-color': "${tearLineColor?.let { toCssHex(it) }}",
                                 '--border-color': "$borderColor",
                                 '--editor-color': "$editorColor",
+                                '--vulnerability-overview-pre-background-color': "$preBackgroundColor",
                                 '--label-color': "'$labelColor'",
                             };
                             for (let [property, value] of Object.entries(properties)) {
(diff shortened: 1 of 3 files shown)


src/main/kotlin/io/snyk/plugin/ui/jcef/ThemeBasedStylingGenerator.kt
```diff
@@ -74,10 +74,7 @@ class ThemeBasedStylingGenerator {
                     val tabItemHoverColor =
                         globalScheme.getColor(ColorKey.find("INDENT_GUIDE")) // The closest color to target_rgb = RGB (235, 236, 240)
 
-                    val editorColorsManager = EditorColorsManager.getInstance()
-                    val editorColorsScheme: EditorColorsScheme = editorColorsManager.globalScheme
-                    val fontPreferences: FontPreferences = editorColorsScheme.fontPreferences
-                    val editorFont = fontPreferences.fontFamily
+                    val preBackgroundColor = if (isDarkTheme) "#000000" else "#ffffff"
```
bastiandoetsch, line 77: Why are the values hard coded? Can't we use values from the color scheme?

bastiandoetsch: https://plugins.jetbrains.com/docs/intellij/themes-customize.html#finding-attribute-keys-for-ui-controls should help to find the right entry.

src/main/resources/stylesheets/snyk_code_suggestion.scss
```diff
@@ -27,7 +27,7 @@ a,
 }
 
 // TODO: remove ignore styling at the end of cycle 5 2025
-// TODO: keep the font-size
+// TODO: keep the font-size - it should come from Language Server
```
bastiandoetsch, line 30: Actually, shouldn't it come from the IDE settings? https://www.perplexity.ai/search/intellij-sdk-editorcolorsschem-z5REg62VRIOT53bY4zTslg#1

cat2608: I'm not sure why we need to "_remove ignore styling at the end of cycle 5 2025. Keep the font-size_", but I just wanted to note that in my opinion, if possible, move as much as we can to Language Server so we have less repeated rules spread across IDEs. 

The class `ignore-warning` is defined in Language Server and we are overriding the same rules with the same values form IntelliJ but indeed using a custom `font-size`. With this PR, properties related to `font` are relative to the `root` definition so it could be that this `ignore-warning` component does not need to have its custom font-size specified from IntelliJ if this component stays.

I'll remove my comment to not cause confusion :)

ide-vs-ls

ignores-warning

src/main/kotlin/io/snyk/plugin/ui/jcef/ThemeBasedStylingGenerator.kt
cat2608: It works better now! Thank you 🙏 

theme

bastiandoetsch (approved):

PR #579, opened 2024-08-26
chore: refactor and unit test

acke:
### Description

### Checklist

- [ ] Linted
- [ ] CHANGELOG.md updated

### Screenshots / GIFs


bastiandoetsch:
Let's wait for https://github.com/snyk/snyk-intellij-plugin/pull/549 - there's some fixes & conflicts would occur.

src/main/kotlin/snyk/common/lsp/SnykLanguageClient.kt
```diff
@@ -120,33 +120,37 @@ class SnykLanguageClient :
...
-            }
+    fun clearCache(scanPublisher: SnykScanListenerLS, snykFile: SnykFile) {
+        updateCacheForProduct(scanPublisher, "code", snykFile, emptyList())
+        updateCacheForProduct(scanPublisher, "oss", snykFile, emptyList())
+    }
 
-            LsProductConstants.Container.value -> {
-                // TODO implement
-            }
+    private fun updateCacheForProduct(
```
bastiandoetsch, line 145: I wouldn't extract this to a method for effectively 3 lines of code. The effect may be updating the cache, but the event bus usage makes it flexible - there could be other listeners. I would call it something like, well, publishDiagnostics - and then the question is, if extracting it really gives us benefits in maintainability and understandability.

src/main/kotlin/snyk/common/lsp/SnykLanguageClient.kt
```diff
@@ -120,33 +120,37 @@ class SnykLanguageClient :
...
-                scanPublisher.onPublishDiagnostics(product, snykFile, issueList)
-            }
-
-            LsProductConstants.InfrastructureAsCode.value -> {
-                // TODO implement
-            }
+    fun clearCache(scanPublisher: SnykScanListenerLS, snykFile: SnykFile) {
+        updateCacheForProduct(scanPublisher, "code", snykFile, emptyList())
+        updateCacheForProduct(scanPublisher, "oss", snykFile, emptyList())
+    }
```
bastiandoetsch, line 143: I would not call this clearCache, it should be sth like `publishEmptyDiagnostics`. Using the event bus is meant to decouple event producers from event consumers. We should not assume that a cache is consuming the event.

bastiandoetsch (approved):

PR #604, opened 2024-09-07
fix: only create one annotation per finding, not one per fix

bastiandoetsch:
### Description

Previously, each fix was created as an annotation. This cluttered the hover/tooltip. Now only the issue is created, and fixes are added to that single line of text.

### Checklist

- [x] Linted
- [ ] CHANGELOG.md updated

### Screenshots / GIFs
Before:
image
image

After:
image
image


acke (approved):
LGTM!

PR #634, opened 2024-11-12
fix: add iac ignore button [IDE-683]

DariusZdroba:
### Description

This provides the logic to ignore an IaC finding for a specific file in the IntelliJ plugin, the ignore button is implicitly hidden and set to show in IntelliJ only.

The **HTML** panel that comes from the **Language Server** receives dynamic attributes with injected values such as the `path` and `Issue.ID`, these are extracted in a custom JS script and called through a callback when clicking **ignore button** , this triggers the `IgnoreHandler` .
In the `IgnoreHandler` the `IgnoreService` is called which calls calls the **CLI** through the **LS Wrapper** to execute the `snyk ignore` command on the specified params.

This path and issue id comes injected in the HTML template which comes from Snyk LS.
### Checklist

- [x] Linted
- [x] CHANGELOG.md updated

### Screenshots / GIFs

### DIFF
--- a/src/main/kotlin/io/snyk/plugin/ui/jcef/IgnoreInFileHandler.kt
+++ b/src/main/kotlin/io/snyk/plugin/ui/jcef/IgnoreInFileHandler.kt
@@ -5,6 +5,7 @@ import com.intellij.openapi.diagnostic.Logger
 import com.intellij.openapi.project.Project
 import com.intellij.ui.jcef.JBCefBrowserBase
 import com.intellij.ui.jcef.JBCefJSQuery
+import io.snyk.plugin.getContentRootPaths
 import io.snyk.plugin.runInBackground
 import io.snyk.plugin.ui.SnykBalloonNotificationHelper
 import org.cef.browser.CefBrowser
@@ -33,7 +34,7 @@ class IgnoreInFileHandler(
             val issueId = params[0] // ID of issue that needs to be ignored
             val filePath = params[1]
             // Computed path that will be used in the snyk ignore command for the --path arg
-            val computedPath = filePath.removePrefix("${project.basePath}${File.separator}")
+            val computedPath = filePath.removePrefix("${project.getContentRootPaths().firstOrNull()}${File.separator}")
             // Avoid blocking the UI thread
             runInBackground("Snyk: applying ignore...") {
                 val result = try {
--- a/src/main/kotlin/io/snyk/plugin/ui/jcef/IgnoreInFileHandler.kt
+++ b/src/main/kotlin/io/snyk/plugin/ui/jcef/IgnoreInFileHandler.kt
@@ -1,0 +1,96 @@
+package io.snyk.plugin.ui.jcef
+
+import com.intellij.openapi.diagnostic.LogLevel
+import com.intellij.openapi.diagnostic.Logger
+import com.intellij.openapi.project.Project
+import com.intellij.ui.jcef.JBCefBrowserBase
+import com.intellij.ui.jcef.JBCefJSQuery
+import io.snyk.plugin.runInBackground
+import io.snyk.plugin.ui.SnykBalloonNotificationHelper
+import org.cef.browser.CefBrowser
+import org.cef.browser.CefFrame
+import org.cef.handler.CefLoadHandlerAdapter
+import snyk.common.IgnoreService
+import snyk.common.lsp.LanguageServerWrapper
+import java.io.File
+import java.io.IOException
+
+class IgnoreInFileHandler(
+    private val project: Project,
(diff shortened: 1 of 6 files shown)


src/main/kotlin/io/snyk/plugin/ui/jcef/IgnoreInFileHandler.kt
```diff
@@ -0,0 +1,96 @@
...
+
+    fun generateIgnoreInFileCommand(jbCefBrowser: JBCefBrowserBase): CefLoadHandlerAdapter {
+        val applyIgnoreInFileQuery = JBCefJSQuery.create(jbCefBrowser)
+
+        applyIgnoreInFileQuery.addHandler { value ->
+            val params = value.split("|@", limit = 2)
+            val issueId = params[0] // ID of issue that needs to be ignored
+            val filePath = params[1]
+            // Computed path that will be used in the snyk ignore command for the --path arg
+            val computedPath = filePath.removePrefix("${project.basePath}${File.separator}")
```
bastiandoetsch, line 36: is it necessary to make the file relative?

bastiandoetsch: `project.basePath` should not be used in new code - instead the content root path should be used.

src/main/kotlin/io/snyk/plugin/ui/toolwindow/panels/JCEFDescriptionPanel.kt
```diff
@@ -341,9 +365,20 @@ class SuggestionDescriptionPanelFromLS(
...
               }
 
+              function applyIgnoreInFile() {
+                console.log('Applying ignore', issue);
+                if (!issue) return;
+
+                window.applyIgnoreInFileQuery(issue + '|@' + filePath + ' > ' + resourcePath);
+                toggleElement(ignoreInFileBtn, "hide");
+                console.log('Applying ignore');
+              }
```
bastiandoetsch, line 375: Can't this script snippet be provided by LS?

src/main/kotlin/io/snyk/plugin/ui/toolwindow/panels/JCEFDescriptionPanel.kt
```diff
@@ -382,6 +421,17 @@ class SuggestionDescriptionPanelFromLS(
...
+                console.log('[[receiveIgnoreInFileResponse]]', success);
+                if(success){
+                    ignoreInFileBtn.disabled = true;
+                    console.log('Ignored in file', success);
+                    document.getElementById('ignore-file-issue').disabled = true;
+                }else{
+                    toggleElement(ignoreInFileBtn, "show");
+                    console.error('Failed to apply fix', success);
+                }
+              }
```
bastiandoetsch, line 434: Can't the whole added part be pushed down to LS?

src/main/kotlin/io/snyk/plugin/ui/jcef/IgnoreInFileHandler.kt
DariusZdroba: Yes, it doesn't work with the full path, it seems that the `snyk ignore` expects the path to be relative to the project root folder.

DariusZdroba: I'll use the `getContentRootPaths()` instead of the `basePath`

src/main/kotlin/io/snyk/plugin/ui/toolwindow/panels/JCEFDescriptionPanel.kt
DariusZdroba: It can, it just depends how we want to use this logic, my question would be : **Should** we push it down to LS? do we plan on using it from LS? eventually with other IDE's than IntelliJ

bastiandoetsch (approved):

src/main/kotlin/io/snyk/plugin/ui/toolwindow/panels/JCEFDescriptionPanel.kt
```diff
@@ -209,7 +233,6 @@ class SuggestionDescriptionPanelFromLS(
 
     private fun getCustomScript(): String {
         return """
-            (function () {
```
DariusZdroba, line 212: When extracting the script, we need acces to the variables which are locked in the scope of this self invoked function and can't be accessed outside, we can have the logic encapsulated in a self invoked function in the LS, that would require some minimal changes to all IDE's that inject scripts.

PR #641, opened 2024-12-17
fix: update root node branch name on changing base branch [IDE-841]

DariusZdroba:
### Description

Root Node base branch text was only initialized never updated, this PR updates it whenever the other RootTreeNodes update.
Note: The state is and was already saved as intended in the folder config.
### Checklist

- [ ] Linted
- [x] CHANGELOG.md updated

### Screenshots / GIFs

### DIFF
--- a/src/main/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanel.kt
+++ b/src/main/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanel.kt
@@ -552,8 +552,21 @@ class SnykToolWindowPanel(
 
         val newContainerTreeNodeText = getNewContainerTreeNodeText(settings, containerResultsCount, addHMLPostfix)
         newContainerTreeNodeText?.let { rootContainerIssuesTreeNode.userObject = it }
+
+        val newRootTreeNodeText = getNewRootTreeNodeText()
+        newRootTreeNodeText.let { rootTreeNode.info = it }
     }
 
+    private fun getNewRootTreeNodeText() : String {
+        val folderConfig = service<FolderConfigSettings>().getFolderConfig(project.basePath.toString())
+            if (folderConfig?.let {
+                getRootNodeText(
+                    it.folderPath,
+                    it.baseBranch
+                )
+            } != null)  return folderConfig.let { getRootNodeText(it.folderPath, it.baseBranch) }
+            return "Choose branch on ${project.basePath}"
+    }
     private fun getNewContainerTreeNodeText(
         settings: SnykApplicationSettingsStateService,
         containerResultsCount: Int?,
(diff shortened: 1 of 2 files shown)


src/main/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanel.kt
```diff
@@ -552,8 +552,21 @@ class SnykToolWindowPanel(
 
         val newContainerTreeNodeText = getNewContainerTreeNodeText(settings, containerResultsCount, addHMLPostfix)
         newContainerTreeNodeText?.let { rootContainerIssuesTreeNode.userObject = it }
+
+        val newRootTreeNodeText = getNewRootTreeNodeText()
+        newRootTreeNodeText.let { rootTreeNode.info = it }
```
DariusZdroba, line 557: Can this be part of the `updateTreeRootNodesPresentation` or is the context for this only supposed to extend the product issue nodes? 

My intended logic was to have all the RootNodes update at the same time when this is triggered.

bastiandoetsch:
Would you please update the changelog?

src/main/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanel.kt
bastiandoetsch: This is fine, I think.

src/main/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanel.kt
```diff
@@ -552,8 +552,21 @@ class SnykToolWindowPanel(
...
     }
 
+    private fun getNewRootTreeNodeText() : String {
+        val folderConfig = service<FolderConfigSettings>().getFolderConfig(project.basePath.toString())
```
bastiandoetsch, line 561: You shouldn't use the `project.basePath` as stated in the IntelliJ SDK docs. In the future we'll most likely use content root nodes like in Eclipse in the future. We can pair on this to explore how to solve this issue, but I feel like using `project.basePath` with folderConfigs creates inconsistencies.

bastiandoetsch: That said, it's like that in the other code, too. Just wondering if we can improve it.

DariusZdroba: Ok, as discussed this is part of a bigger issue, let's pair on this sometime and figure out all the instances of the `project.basePath` that we use where we should use the content root.

bastiandoetsch (approved):

PR #654, opened 2025-02-10
fix: correctly scale css fonts based on ide defaults

andrewrobinsonhodges-snyk:
### Description

### Checklist

- [ ] Linted
- [ ] CHANGELOG.md updated

### Screenshots / GIFs


src/main/kotlin/io/snyk/plugin/ui/jcef/ThemeBasedStylingGenerator.kt
```diff
@@ -32,7 +32,9 @@ class ThemeBasedStylingGenerator {
             val isHighContrast =
                 EditorColorsManager.getInstance().globalScheme.name.contains("High contrast", ignoreCase = true)
             html = html.replace("--default-font: ", "--default-font: \"${JBUI.Fonts.label().asPlain().family}\", ")
-            html = html.replace("var(--main-font-size)", JBFont.small().size.toString() + "px")
+            // CSS from the language server assumes a default size of 10px. JBFont uses a default of 13pt, so we
```
bastiandoetsch, line 35: Wondering. Should we scale it proportionally or just add the +3 everytime. Also, if we don't scale it but use the configured JBFont size - how does it look?

PR #663, opened 2025-03-05
fix: update cache on receive diags

andrewrobinsonhodges-snyk:
### Description

Depending on the LSP method used, LS sends the either the long or the short product name ("Snyk Code" vs "code", for example). This was hard to track, and we had a few bugs where events were getting lost because we were comparing against the wrong strings.

This fix generalises the processing to handle either form, ensuring that all LS callbacks are handled.

### Checklist

- [ ] Linted
- [ ] CHANGELOG.md updated

### Screenshots / GIFs


bastiandoetsch (approved):
Nice one

PR #679, opened 2025-03-20
fix: remove unsupported project specific setting for additional params [IDE-1080]

acke:
### Description

The additional parameters are now correctly read from the folderConfig.

Path and URI Editing Util functions no longer rely on Java's libraries. These change behaviour depending on the OS on which they're running. We've added our own util functions which are OS independent, and should mean that UTs and general file handling are more stable and predictable.

### Checklist

- [ ] Linted
- [ ] CHANGELOG.md updated

### Screenshots / GIFs


src/main/kotlin/io/snyk/plugin/ui/SnykSettingsDialog.kt
```diff
@@ -178,7 +178,13 @@ class SnykSettingsDialog(
 
             cliPathTextBoxWithFileBrowser.text = applicationSettings.cliPath
             cliBaseDownloadUrlTextField.text = applicationSettings.cliBaseDownloadURL
-            additionalParametersTextField.text = applicationSettings.getAdditionalParameters(project)
+            service<FolderConfigSettings>().getAllForProject(project)
```
bastiandoetsch, line 181: I may be blind, but why is this called here?

src/main/kotlin/io/snyk/plugin/ui/SnykSettingsDialog.kt
```diff
@@ -178,7 +178,13 @@ class SnykSettingsDialog(
...
+            additionalParametersTextField.text = project.basePath?.let {
+                service<FolderConfigSettings>().getAdditionalParams(
+                    it
+                )
+            }
```
bastiandoetsch, line 186: I'm not sure I understand the logic here. `project.basePath` should not be used if it can be avoided. Why not iterate over the return values of `getAllForProject` and apply the first we find? In the future, we'll need to work on content roots, not projects, as configuration can differ by content root.

src/main/kotlin/io/snyk/plugin/ui/SnykSettingsDialog.kt
```diff
@@ -178,7 +178,13 @@ class SnykSettingsDialog(
...
+                //applicationSettings.getAdditionalParameters(project)
```
bastiandoetsch, line 187: This comment can go :)

src/main/kotlin/snyk/common/lsp/settings/FolderConfigSettings.kt
```diff
@@ -61,4 +61,15 @@ class FolderConfigSettings : SimplePersistentStateComponent<FolderConfigSettings
...
+    /**
+     * Returns the additional parameters of the folder config at [folderPath] as a string.
+     *
+     * If no config is found for the path, an empty string is returned. If the additional parameters are null,
+     * they will be replaced with an empty list before being joined into a CSV string.
+     */
+    fun getAdditionalParams(folderPath: String): String {
+        val fromJson = gson.fromJson(state.configs[folderPath], FolderConfig::class.java)
+        return (fromJson?.additionalParameters ?: emptyList()).joinToString(" ")
+    }
```
bastiandoetsch, line 74: Why would we have an accessor here? Parsing the JSON for each field access IMHO does not make sense. We should get a folderConfig and use the object instead of custom code in this class. Also, the generated comment (especially the last sentence) is kind of weirdly unnecessarily specific, as we don't need to know how the output is actually achieved.

src/main/kotlin/snyk/common/lsp/settings/FolderConfigSettings.kt
```diff
@@ -1,63 +1,40 @@
...
-                folderPath = fromJson.folderPath,
-                localBranches = fromJson.localBranches ?: emptyList(),
-                additionalParameters = fromJson.additionalParameters ?: emptyList(),
-            )
-            addFolderConfig(copy)
-            return copy
-        }
-        return fromJson
+    internal fun getFolderConfig(folderPath: String): FolderConfig {
+        return configs[folderPath]!!
```
bastiandoetsch, line 23: this may need a fallback to create a new folderconfig, if none is found

bastiandoetsch (approved):

src/test/kotlin/snyk/iac/IacServiceTest.kt
```diff
@@ -41,8 +39,6 @@ class IacServiceTest : LightPlatformTestCase() {
         settingsStateService.lastCheckDate = null
         settingsStateService.organization = ""
 
-        project.service<SnykProjectSettingsStateService>().additionalParameters = ""
-
```
bastiandoetsch, line 45: this was local storage - removed (as the test below)

andrewrobinsonhodges-snyk (approved):

ShawkyZ (approved):

rrama (approved):
Looked at Utils and the test for it, they looked good to me. The rest seems to have had prior approvals, so I am happy.

PR #793, opened 2026-02-27
fix: allow navigation to ranges in files which are not IDE documents [IDE-1781]

andrewrobinsonhodges-snyk:
### Description

The navigateToSource function used when a user clicks on an issue in the tree view previously validated the range by comparing to the length of the target file as an IDE document.

This mean that we were silently failing to navigate to files that the IDE could not process as a document (e.g., .gitleaksignore). The fix is to log the discrepancy and let the PsiNavigationSupport library handle it. The file will now open, even if the correct range cannot be determined.

Note that we expect that the ranges returned by the Secrets scanner should always be correct, so this was an overly defensive practice.

### Checklist

- [x] Linted
- [ ] CHANGELOG.md updated

### Screenshots / GIFs

### DIFF
--- a/src/test/kotlin/io/snyk/plugin/UtilsKtTest.kt
+++ b/src/test/kotlin/io/snyk/plugin/UtilsKtTest.kt
@@ -325,12 +325,10 @@ class UtilsKtTest {
     every { appMock.isDisposed } returns false
     // Mock invokeLater to execute runnables immediately
     // The top-level invokeLater delegates to Application.invokeLater with ModalityState
-    every { appMock.invokeLater(any<Runnable>()) } answers {
-      firstArg<Runnable>().run()
-    }
-    every { appMock.invokeLater(any<Runnable>(), any<com.intellij.openapi.application.ModalityState>()) } answers {
-      firstArg<Runnable>().run()
-    }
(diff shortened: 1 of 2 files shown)


src/test/kotlin/io/snyk/plugin/UtilsKtTest.kt
```diff
@@ -311,6 +313,89 @@ class UtilsKtTest {
...
+    mockkStatic(ApplicationManager::class)
+    mockkStatic(PsiNavigationSupport::class)
+
+    val appMock = mockk<com.intellij.openapi.application.Application>(relaxed = true)
+    every { ApplicationManager.getApplication() } returns appMock
+    every { appMock.isDisposed } returns false
+    every { appMock.invokeLater(any<Runnable>()) } answers { firstArg<Runnable>().run() }
+    every {
+      appMock.invokeLater(any<Runnable>(), any<com.intellij.openapi.application.ModalityState>())
+    } answers { firstArg<Runnable>().run() }
```
rrama, line 370: Might want to pull this out as a common test util.

rrama (approved):
LGTM.

PR #795, opened 2026-03-05
feat(IDE-1701): settings page auth flow — bridge persist and forward apiUrl

nick-y-snyk:
## What & Why

The settings page drives authentication via `snyk.login` with `[authMethod, endpoint, insecure]` args. Before forwarding to the LS, the IDE must persist these values locally so they survive a page close. Additionally, `$/snyk.hasAuthenticated` now carries `apiUrl` which is forwarded to the settings panel webview so the settings page can update both fields after auth.

Bridge persist written directly to `pluginSettings()` properties — no `updateConfiguration()` call means no `didChangeConfiguration` loop.

## Changes

**`ExecuteCommandBridge.kt`** (new shared class)
- `LOGIN_TIMEOUT_MS = 120_000L`; `LONG_RUNNING_COMMANDS = setOf("snyk.login")` — login waits up to 2 min for browser OAuth
- Replaces `__ideLogin__`/`__ideLogout__` with `window.__ideExecuteCommand__(cmd, args, callback)` bridge
- Before `ls.executeCommandWithArgs`, when `command == "snyk.login"` and `args.size >= 3`, calls `saveLoginArgs(args)`:
  - Maps `args[0]` → `AuthenticationType` (`"oauth"` → `OAUTH2`, `"pat"` → `PAT`, `"token"` → `API_TOKEN`)
  - Sets `pluginSettings().authenticationType`, `.customEndpointUrl`, `.ignoreUnknownCA`
  - No `updateConfiguration()` call → no `didChangeConfiguration` sent to LS
- `LanguageServerWrapper.executeCommandWithArgs` gains `timeoutMillis: Long = 5_000` param

**`HTMLSettingsPanel.kt`**
- `companion object { @Volatile var instance: HTMLSettingsPanel? = null }` — set in `init`, cleared in `dispose()`
- `setAuthToken(token: String, apiUrl: String?)` — escapes both, executes `window.setAuthToken(safeToken, safeApiUrl)` via JCEF on `invokeLater` with `ModalityState.any()`

**`SnykLanguageClient.kt`**
Calls `HTMLSettingsPanel.instance?.setAuthToken(param.token ?: "", param.apiUrl)` before saving to `pluginSettings()` — settings page webview updates immediately after auth.

**`SaveConfigHandler.kt`**
Removes the `messageBus.subscribe(SnykSettingsListener.SNYK_SETTINGS_TOPIC, ...)` block that pushed token updates to the browser — replaced by the direct `HTMLSettingsPanel.instance?.setAuthToken(...)` call in `SnykLanguageClient`.

## Tests

- `SaveConfigHandlerExecuteCommandTest.kt`: bridge persist test added (sends `snyk.login` with `["pat", "https://api.eu.snyk.io", true]`, asserts `authenticationType == PAT`, `customEndpointUrl`, `ignoreUnknownCA`); all `executeCommandWithArgs` verify calls include timeout arg
- `SnykLanguageClientTest.kt`: `setAuthToken` called on `HTMLSettingsPanel.instance` when panel is open (new test)

## Test plan

- [ ] Settings page: change endpoint + auth method → click Authenticate → IDE saves values → auth succeeds → token and apiUrl appear in settings page
- [ ] Close/reopen settings page → values persist
- [ ] Panel login → auth succeeds → token and apiUrl shown immediately
- [ ] No `didChangeConfiguration` sent when bridge saves auth params
- [ ] Run test suite: `./gradlew test`

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -47,6 +47,10 @@ repositories {
   maven { url = uri("https://cache-redirector.jetbrains.com/intellij-dependencies") }
 }
 
+configurations.all {
+  resolutionStrategy.force("com.fasterxml.jackson.core:jackson-core:2.18.6")
+}
+
 dependencies {
   intellijPlatform {
     intellijIdeaCommunity(properties("platformVersion"))
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -47,9 +47,7 @@ repositories {
   maven { url = uri("https://cache-redirector.jetbrains.com/intellij-dependencies") }
 }
 
-configurations.all {
-  resolutionStrategy.force("com.fasterxml.jackson.core:jackson-core:2.18.6")
-}
+configurations.all { resolutionStrategy.force("com.fasterxml.jackson.core:jackson-core:2.18.6") }
 
 dependencies {
   intellijPlatform {
--- a/src/main/kotlin/io/snyk/plugin/ui/jcef/ExecuteCommandBridge.kt
+++ b/src/main/kotlin/io/snyk/plugin/ui/jcef/ExecuteCommandBridge.kt
@@ -19,6 +19,12 @@ import snyk.common.lsp.LanguageServerWrapper
  * - Dispatch incoming command requests to the Language Server.
  * - Return results to the JS callback via `window.__ideCallbacks__`.
  */
+data class ExecuteCommandRequest(
+  val command: String = "",
+  val args: List<Any> = emptyList(),
+  val callbackId: String? = null,
+)
+
 class ExecuteCommandBridge(private val project: Project) {
   private val log = logger<ExecuteCommandBridge>()
   private val gson = Gson()
@@ -82,7 +88,7 @@ class ExecuteCommandBridge(private val project: Project) {
    */
   internal fun dispatch(value: String, callbackExecutor: ((String, String) -> Unit)? = null) {
     try {
-      val request = gson.fromJson(value, TreeViewCommandRequest::class.java)
+      val request = gson.fromJson(value, ExecuteCommandRequest::class.java)
       log.debug("ExecuteCommandBridge: received command=${request.command}, args=${request.args}")
       if (request.command.isBlank()) {
         log.warn("ExecuteCommandBridge: received empty command, ignoring")
--- a/src/main/kotlin/io/snyk/plugin/ui/jcef/ExecuteCommandBridge.kt
+++ b/src/main/kotlin/io/snyk/plugin/ui/jcef/ExecuteCommandBridge.kt
@@ -5,6 +5,8 @@ import com.intellij.openapi.diagnostic.logger
 import com.intellij.openapi.project.Project
 import com.intellij.ui.jcef.JBCefBrowserBase
 import com.intellij.ui.jcef.JBCefJSQuery
+import io.snyk.plugin.pluginSettings
+import io.snyk.plugin.services.AuthenticationType
 import org.jetbrains.concurrency.runAsync
 import snyk.common.lsp.LanguageServerWrapper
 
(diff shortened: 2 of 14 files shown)


build.gradle.kts
```diff
@@ -47,6 +47,8 @@ repositories {
   maven { url = uri("https://cache-redirector.jetbrains.com/intellij-dependencies") }
 }
 
+configurations.all { resolutionStrategy.force("com.fasterxml.jackson.core:jackson-core:2.18.6") }
```
ShawkyZ, line 50: let's see if we can bump the direct dep

nick-y-snyk: The only way to get a newer jackson version here would be to bump the IntelliJ platform version (from 2024.2 to a newer one that ships jackson 2.18.x+). Otherwise, the
  resolutionStrategy.force is the correct approach for this situation.

  Want to check what jackson version newer IntelliJ platform versions ship with, or keep the force override as-is?

src/main/kotlin/io/snyk/plugin/ui/jcef/ExecuteCommandBridge.kt
```diff
@@ -0,0 +1,146 @@
...
+ * - Inject the JS-side `window.__ideExecuteCommand__` bridge into a JCEF panel.
+ * - Dispatch incoming command requests to the Language Server.
+ * - Return results to the JS callback via `window.__ideCallbacks__`.
+ */
+class ExecuteCommandBridge(private val project: Project) {
+  private val log = logger<ExecuteCommandBridge>()
+  private val gson = Gson()
+
+  companion object {
+    internal val SAFE_CALLBACK_ID = Regex("^[a-zA-Z0-9_]+$")
```
andrewrobinsonhodges-snyk, line 33: Maybe rename this to SAFE_CALLBACK_ID_REGEX or similar to clarify that this isn't actually the ID.

src/main/kotlin/io/snyk/plugin/ui/jcef/ExecuteCommandBridge.kt
```diff
@@ -0,0 +1,146 @@
...
+      val ignoreUnknownCA = args[2] as? Boolean ?: false
+      val authType =
+        when (authMethodStr) {
+          "oauth" -> AuthenticationType.OAUTH2
+          "pat" -> AuthenticationType.PAT
+          "token" -> AuthenticationType.API_TOKEN
+          else -> AuthenticationType.OAUTH2
+        }
+      pluginSettings().authenticationType = authType
+      pluginSettings().customEndpointUrl = endpoint
```
andrewrobinsonhodges-snyk, line 146: Is it correct that we remove the endpoint if it is not passed in as an argument?

andrewrobinsonhodges-snyk: Similar for setting ignoreUnknownCA to false below

src/test/kotlin/io/snyk/plugin/ui/jcef/SaveConfigHandlerExecuteCommandTest.kt
```diff
@@ -0,0 +1,154 @@
...
+  fun tearDown() {
+    unmockkAll()
+  }
+
+  @Test
+  fun `dispatchSettingsCommand routes snyk login to language server with auth args`() {
+    val request =
+      TreeViewCommandRequest(
+        command = "snyk.login",
+        args = listOf("oauth2", "https://api.snyk.io", false),
```
andrewrobinsonhodges-snyk: The actual argument is "oauth". This test only works as oauth is the default. It's worth adding explicit tests for the four cases.

andrewrobinsonhodges-snyk (approved):

src/main/kotlin/io/snyk/plugin/ui/jcef/ExecuteCommandBridge.kt
nick-y-snyk: these values are fetched from html form, so we save old value or a new one if user changed it

src/test/kotlin/io/snyk/plugin/ui/jcef/SaveConfigHandlerExecuteCommandTest.kt
nick-y-snyk: fixed + added

PR #815, opened 2026-04-28
fix: all services should be accessed at the app level so config changes stick when made in other projects

andrewrobinsonhodges-snyk:
### Description

Two issues fixed:

Severity filters were reading from the global tree view state, so per-folder settings were not getting reflected in the Snyk tool window.

When setting global settings with multiple projects open, only the IntelliJ instance where the project was open was getting updated. We need to share all services at the app level to avoid this.

### Checklist

- [x] Linted
- [ ] CHANGELOG.md updated

### Screenshots / GIFs


src/main/kotlin/io/snyk/plugin/ui/jcef/SaveConfigHandler.kt
```diff
@@ -169,7 +169,13 @@ class SaveConfigHandler(
       config.folderConfigs?.let { applyFolderConfigs(it) }
     }
 
-    LanguageServerWrapper.getInstance(project).updateConfiguration()
+    // Notify all open projects' language servers so global settings propagate everywhere.
```
bastiandoetsch, line 174: good catch.

src/main/kotlin/io/snyk/plugin/ui/jcef/SaveConfigHandler.kt
```diff
@@ -169,7 +169,13 @@ class SaveConfigHandler(
...
+    // Without this, only the current project's LS/HTML page would reflect global changes.
+    for (openProject in com.intellij.ide.impl.ProjectUtil.getOpenProjects()) {
```
bastiandoetsch: nit: this should be imported instead of using the full package signature

src/main/kotlin/io/snyk/plugin/ui/jcef/SaveConfigHandler.kt
```diff
@@ -169,7 +169,13 @@ class SaveConfigHandler(
...
+      if (!openProject.isDisposed) {
```
bastiandoetsch: We should check our own Disposable in addition: SnykPluginDisposable

src/main/kotlin/snyk/common/lsp/SnykLanguageClient.kt
```diff
@@ -325,6 +326,17 @@ class SnykLanguageClient(private val project: Project, val progressManager: Prog
             logger.error("Error processing snyk folder configs", e)
           }
         }
+
```
bastiandoetsch, line 329: why not just move them into the async block and do a syncPublish? this would prevent additional thread switching.

bastiandoetsch (dismissed):

bastiandoetsch:
SnykToolWindowSnykScanListenerTest > test filtered issue count matches displayed tree nodes FAILED

I'm fixing this now.

src/main/kotlin/snyk/common/lsp/SnykLanguageClient.kt
andrewrobinsonhodges-snyk: It would, but the downside is that any locks on the EDT in invokeLater might block on the context of the async block, causing deadlocks. In practice, we only have one of these (updating the tree view), but it does make it more fragile.

I don't mind either way: The threads are lightweight, but on the other hand the risk of deadlocking is low.

The tradeoff is maintainability; any future changes to the listener would need to be super careful about their EDT handling.

bastiandoetsch (approved):

PR #838, opened 2026-06-03
fix: don't duplicate timeouts for LS loading - fixes fallback to local HTML settings page [IDE-2218]

andrewrobinsonhodges-snyk:
If the LS is down, we should show the fallback HTML settings page after the timer pops. Because of the way the LS check was implemented, we actually stacked timers, which caused the loading screen to show for over 30s in some cases.

### Description

### Checklist

- [ ] Linted
- [ ] CHANGELOG.md updated

### Screenshots / GIFs


src/main/kotlin/snyk/common/lsp/LanguageServerWrapper.kt
```diff
@@ -275,6 +293,43 @@ class LanguageServerWrapper(private val project: Project) : Disposable {
...
+      ) {
+        // Re-apply the cached version so callers/settings observe the same value as a fresh check.
+        cached.parsedVersion?.let { ps.currentLSProtocolVersion = it }
+        return cached.compatible
+      }
+    }
+
+    val (compatible, parsedVersion) = runCliProtocolVersionCheck(required)
+    parsedVersion?.let { ps.currentLSProtocolVersion = it }
+    protocolVersionCheckCache =
```
bastiandoetsch: [Should Fix] This caches the protocol-check result unconditionally — including the transient failure cases (`runCliProtocolVersionCheck` returns `false to null` on timeout, non-zero exit, and exception at lines 344/349/354). The cache key is the CLI file's path/mtime/size and is only cleared on explicit `restart()`, so a one-off timeout on a loaded machine makes the LS report incompatible for the rest of the session. Before this change each `ensureLanguageServerInitialized()` re-ran the subprocess and self-healed. Since the PR's own intent is to handle LS timeouts gracefully, caching a timeout as a permanent verdict is the opposite of the goal.

Fix: only cache definitive results — cache on success and on a *parsed* version mismatch (`parsedVersion != null`), but skip caching when `parsedVersion == null` (timeout/exit≠0/exception). Add a test where the CLI times out on the first call and succeeds on the second, asserting the second call re-runs and returns true. (Flagged by 2 reviewers as the most severe finding.)

— AI review

src/main/kotlin/io/snyk/plugin/ui/settings/HTMLSettingsPanel.kt
```diff
@@ -118,20 +118,34 @@ class HTMLSettingsPanel(private val project: Project) : JPanel(BorderLayout()),
...
+    // CLI protocol-version check (a subprocess that blocks up to 10s) on every
+    // call while the LS is not up, so forcing it here would stall the settings
+    // dialog for ~30s before the fallback renders when the CLI is missing or
+    // incompatible. The fallback exists precisely for that case, so fall back
+    // quickly instead. A short poll still picks up an init that another thread
+    // (e.g. project startup) is completing in the background.
     repeat(3) { attempt ->
-      try {
-        val lsHtml = lsWrapper.getConfigHtml()
+      if (lsWrapper.isInitialized) {
```
bastiandoetsch: [Should Fix] Two things on this gate:

1. **Test gap (the headline claim is unproven).** The two new tests exercise `verifyCliProtocolVersion()` caching directly, but the actual duplicated-timeout fix is *this* `isInitialized` guard — and it has no test. Add a test on `getHtmlContent` (or a thin extracted helper) asserting that when `isInitialized == false`, init/`verifyCliProtocolVersion` is never invoked and the fallback HTML is returned. That's the regression that matters.

2. **Narrow TOCTOU.** `getConfigHtml()` (line 133) internally calls `ensureLanguageServerInitialized()`; `isInitialized` can flip true→false between this check and that call (the LS-termination listener sets it false), so the dialog thread can still hit the 10s blocking init the gate is meant to avoid. Narrow window, but worth closing — e.g. a wrapper method that owns a bounded 'try-get config HTML within a deadline, else not-ready' policy rather than the panel polling around `isInitialized`.

— AI review

bastiandoetsch:
The caching approach is the right idea and the fast-fallback achieves the goal (settings dialog no longer stalls ~30s); `restart()` correctly nulls the cache, and `getHtmlContent` runs off-EDT so the poll loop doesn't freeze the UI. Security scan: clean, no PR-attributable issues. A few more non-blocking notes:

[Suggestion] Cache key is `(cliPath, lastModified, size, requiredVersion)`; a same-size in-place binary swap within coarse mtime granularity could serve a stale verdict. The auto-download path changes size and `restart()` clears the cache, so it's narrow — please just confirm every binary-replacement path either changes size or routes through `restart()` (a content hash would be fully robust but likely overkill).
[Suggestion] `verifyCliProtocolVersion` computes the cache key from one `getCliFile()` read while `runCliProtocolVersionCheck` re-resolves `cliPath` for the subprocess — capture the path/metadata once and pass them in so the cached result can't be attributed to a different file.
[Suggestion] The cache-hit path re-writes `ps.currentLSProtocolVersion` on a read path; confirm nothing else legitimately writes that field between calls (or document the cache as its single source of truth).
[Suggestion] The `@Volatile` comment implies more than it provides — it prevents torn reads but not two concurrent subprocess runs (non-atomic check-then-act). Harmless (idempotent), worth toning down the comment.

— AI review

src/main/kotlin/snyk/common/lsp/LanguageServerWrapper.kt
basti-snyk: **Should-Fix (High): transient CLI-check failures are cached and become sticky for the whole session.**

`runCliProtocolVersionCheck(required)` returns `false to null` for **transient** outcomes too — subprocess timeout, non-zero exit, and any thrown exception (the three `false to null` branches below). This unconditional cache write then stores that `false` under a key of `cliPath / lastModified / size / requiredVersion`, none of which change when the transient condition clears.

Effect: a valid, compatible CLI whose first protocol-version subprocess times out once (cold disk / heavy IO at IDE startup) gets `compatible=false` cached. Every later `ensureLanguageServerInitialized()` then short-circuits on the cached `false` and never re-runs the subprocess, so the Language Server never starts for the rest of the session — until a manual restart (the only in-session cache clear) or the binary's size/mtime changes. Before this PR each call re-ran the check, so a one-off transient failure self-recovered; the caching turns the method's documented fail-closed-**retryable** contract into fail-closed-**sticky**.

Fix: cache only *deterministic* results — a successful parse (compatible, or a parsed-but-mismatched version). Leave timeout / non-zero-exit / exception outcomes uncached so they remain retryable on the next init. This keeps the real win of the change (skipping repeat subprocesses for the two deterministic outcomes) without wedging startup.

— AI review

src/main/kotlin/io/snyk/plugin/ui/settings/HTMLSettingsPanel.kt
```diff
@@ -118,20 +118,34 @@ class HTMLSettingsPanel(private val project: Project) : JPanel(BorderLayout()),
...
+    // Try to get HTML from LS, with retries if LS is still initializing.
+    // Only query the LS once it is already initialized; we must NOT force
+    // initialization from here. ensureLanguageServerInitialized() re-runs the
+    // CLI protocol-version check (a subprocess that blocks up to 10s) on every
+    // call while the LS is not up, so forcing it here would stall the settings
+    // dialog for ~30s before the fallback renders when the CLI is missing or
+    // incompatible. The fallback exists precisely for that case, so fall back
+    // quickly instead. A short poll still picks up an init that another thread
+    // (e.g. project startup) is completing in the background.
     repeat(3) { attempt ->
```
basti-snyk: **Should-Fix (Medium): valid-CLI users who open Settings during initialization can get a transient fallback instead of the real config page.**

This poll waits at most two 1s sleeps (~2s) on `isInitialized`, then returns `loadFallbackHtml()`. The previous path blocked on the `isInitializing` lock in `getConfigHtml()` → `ensureLanguageServerInitialized()` until an in-flight init completed and returned real config. So when the CLI is valid but init is still running (right after project open, or during a CLI download), a normal-but-slow startup that exceeds ~2s now renders the local fallback even though the LS is seconds from ready, and nothing in this diff re-renders once the LS comes up — the user sees the wrong settings page until they reopen the dialog.

This is partly the intended tradeoff (fast fallback vs. the ~30s stall), and it is narrower than the missing/incompatible-CLI case the PR targets. Please confirm an existing refresh path replaces the fallback when the LS finishes initializing; if none exists, either widen the poll modestly or trigger a reload on init-complete. At minimum, document the accepted tradeoff.

— AI review

src/main/kotlin/io/snyk/plugin/ui/settings/HTMLSettingsPanel.kt
basti-snyk: **Suggestion: `isInitialized` is read cross-thread without `@Volatile`.**

`isInitialized` is a plain `var Boolean` in `LanguageServerWrapper`, written by `initialize()` on the init/background thread and now read here directly (outside the `isInitializing` lock) from the settings-panel thread. Under the JMM this read has no visibility guarantee, so the panel may observe a stale `false` and render the fallback even when the LS is already up (or a stale `true`). The field is pre-existing, but this PR adds the new cross-thread reader that decides fallback-vs-LS; marking the field `@Volatile` makes the poll deterministic. Low impact (worst case is one extra fallback/poll cycle), hence a suggestion.

— AI review

src/main/kotlin/snyk/common/lsp/LanguageServerWrapper.kt
andrewrobinsonhodges-snyk: Fixed. `verifyCliProtocolVersion` now only writes `protocolVersionCheckCache` when `parsedVersion != null` — i.e. on a successful parse or a parsed-but-mismatched version. Transient failures (timeout, non-zero exit, exception → `false to null`) are left uncached so the next `ensureLanguageServerInitialized()` re-runs the subprocess and can self-heal. Added a test (`verifyCliProtocolVersion does not cache a transient failure and re-runs on the next call`) asserting the first (transient) call returns false and the second re-runs and returns true. This also resolves the duplicate #3542086373.

andrewrobinsonhodges-snyk: Fixed together with #3367661218: `verifyCliProtocolVersion` now only caches when `parsedVersion != null` (successful parse or parsed-but-mismatched version). Timeout / non-zero-exit / exception outcomes (`false to null`) are left uncached so they stay retryable on the next init. Covered by the new test `verifyCliProtocolVersion does not cache a transient failure and re-runs on the next call`.

src/main/kotlin/io/snyk/plugin/ui/settings/HTMLSettingsPanel.kt
andrewrobinsonhodges-snyk: Both addressed.

**1. Test gap:** extracted the panel's poll/decision logic into a pure, injectable helper `HTMLSettingsPanel.resolveHtmlContent(...)` (the panel can't be constructed headless due to JCEF, hence the companion-helper pattern already used in this file). Added `HTMLSettingsPanelTest` cases asserting: while `isInitialized` is false the LS is **never** queried and the fallback is returned (`usingFallback=true`); when initialized and returning non-blank HTML it returns that HTML (`usingFallback=false`); plus poll-until-ready, blank-response fallback, and exception→lastError cases.

**2. TOCTOU:** closed it. `getConfigHtml` now takes `forceInitialization: Boolean = true`; the settings panel (the only caller) passes `false`, so a not-initialized/flipped-to-false LS returns null immediately instead of entering `ensureLanguageServerInitialized()` → the blocking CLI protocol-version subprocess. The panel thread can no longer trigger LS init. Added a test asserting the non-forcing path returns null without touching `getCliFile()`.

src/main/kotlin/io/snyk/plugin/ui/settings/HTMLSettingsPanel.kt
andrewrobinsonhodges-snyk: Confirmed the refresh path exists, so this is not sticky. When the LS finishes initializing it pushes `$/snyk.configuration`; `SnykLanguageClient.snykConfiguration()` (SnykLanguageClient.kt:356) routes that to `HTMLSettingsPanel.getForProject(project)?.reloadFromLanguageServer()`, which re-fetches config and re-renders over the fallback (`isUsingFallback` flips back to false) — no reopen needed. A valid-but-slow-init user sees the fallback only for the brief poll window, then the real page auto-replaces it. Documented this accepted tradeoff in a comment above the poll in `getHtmlContent`.

src/main/kotlin/io/snyk/plugin/ui/settings/HTMLSettingsPanel.kt
andrewrobinsonhodges-snyk: Done. `isInitialized` is now `@Volatile`. Added a comment clarifying it guarantees visibility of the latest write to cross-thread readers, but only prevents torn/stale reads — it does not make check-then-act atomic (that remains the job of the `isInitializing` lock).

src/main/kotlin/io/snyk/plugin/ui/settings/HTMLSettingsPanel.kt
```diff
@@ -309,7 +342,10 @@ class HTMLSettingsPanel(private val project: Project) : JPanel(BorderLayout()),
     ApplicationManager.getApplication().executeOnPooledThread {
       if (isDisposed) return@executeOnPooledThread
       val lsWrapper = LanguageServerWrapper.getInstance(project)
-      val lsHtml = lsWrapper.getConfigHtml() ?: return@executeOnPooledThread
+      // Non-forcing: this reload is triggered by the LS pushing new config, so it is already up;
+      // never let the settings panel trigger a blocking (re-)initialization from a pooled thread.
+      val lsHtml =
+        lsWrapper.getConfigHtml(forceInitialization = false) ?: return@executeOnPooledThread
```
basti-snyk, line 348: **Should-Fix: the non-forcing recovery here can leave the settings panel stuck on the fallback page.**

`reloadFromLanguageServer()` does a single non-forcing `getConfigHtml(forceInitialization = false)` and returns immediately when the LS isn't up yet. In the fresh-install / CLI-download flow this handler fires from `cliDownloadFinished` while `restartCLI()` → `LanguageServerWrapper.restart()` is still mid-cycle (it `shutdown()`s then `Thread.sleep(1000)` before re-init), so `getConfigHtml` returns `null`, the reload no-ops, and nothing re-triggers once the LS actually comes up — the panel stays on the local fallback HTML until the dialog is closed and reopened. Worse: if `initialize()` bailed at the protocol-version check on the previous CLI, `LanguageServerRestartListener` (registered only after a successful `initialize()`) never subscribed, so `restartCLI()` has no listener and the LS never restarts at all.

Before this change the forcing `getConfigHtml()` on this dedicated pooled thread drove init and recovered the real page. This recovery path is already off the EDT, so blocking/retry is acceptable here — options: retry via the same `resolveHtmlContent` mechanism, force initialization on this recovery path only, or drive `ensureLanguageServerInitialized()`/`restart()` from the download-finished handler.

Suggest an integration test that opens the panel with `isInitialized=false`, then makes the LS become initialized ~1s after `cliDownloadFinished`, and asserts `isUsingFallback` flips back to false.

— AI review

src/main/kotlin/snyk/common/lsp/LanguageServerWrapper.kt
```diff
@@ -113,6 +113,24 @@ class LanguageServerWrapper(private val project: Project) : Disposable {
...
   private var loginFuture: CompletableFuture<Any>? = null
 
+  // Caches the (expensive) CLI protocol-version check. The check spawns a CLI subprocess that
+  // blocks up to PROTOCOL_VERSION_CHECK_TIMEOUT_SECONDS, and it is re-run on every
+  // ensureLanguageServerInitialized() call while the LS is not initialized (e.g. an incompatible
+  // CLI). The result only changes when the CLI binary or the required protocol version changes, so
+  // we key the cache on those and skip the subprocess on a hit. Volatile because tests invoke
+  // verifyCliProtocolVersion() directly, outside the isInitializing lock that otherwise serializes
+  // it via initialize().
+  @Volatile private var protocolVersionCheckCache: ProtocolVersionCheckResult? = null
```
rrama, line 123: Minor AI nitpick:

> **Protocol version cache is per-project-service instance, not shared across projects**
> 
> `LanguageServerWrapper` is a `@Service(Service.Level.PROJECT)` service, so `protocolVersionCheckCache` at line 123 is per-project. If multiple projects are open and all use the same CLI binary, each project independently runs the subprocess on first access. This is acceptable (the check is only expensive on the first call per project lifecycle) but worth noting — a companion-object-level cache keyed on the same fields would avoid redundant subprocess calls across projects.

If it is too difficult to change, then just skip doing it.

src/main/kotlin/io/snyk/plugin/ui/settings/HTMLSettingsPanel.kt
```diff
@@ -73,6 +73,46 @@ class HTMLSettingsPanel(private val project: Project) : JPanel(BorderLayout()),
...
+    internal fun resolveHtmlContent(
+      attempts: Int,
+      isInitialized: () -> Boolean,
+      fetchConfigHtml: () -> String?,
+      loadFallback: () -> String?,
+      onRetryWait: () -> Unit,
+    ): HtmlContentResolution {
+      var lastError: String? = null
+      repeat(attempts) { attempt ->
+        if (isInitialized()) {
```
rrama, line 98: AI nitpick:

> **Redundant isInitialized check between resolveHtmlContent and getConfigHtml**
> 
> The `resolveHtmlContent` lambda checks `isInitialized()` at `HTMLSettingsPanel.kt:98` before calling `fetchConfigHtml()`, which in turn calls `getConfigHtml(forceInitialization = false)` at `LanguageServerWrapper.kt:1182` which checks `isInitialized` again. This double-check is harmless (belt-and-suspenders against the TOCTOU race where `isInitialized` flips between the two reads) but worth being aware of — the outer check in `resolveHtmlContent` is the one that gates the retry/wait logic, while the inner check in `getConfigHtml` is the safety net that prevents sending commands to a dead LS.

rrama (approved):
LGTM! Manually tested with an older version and it really helped 😄

PR #839, opened 2026-06-03
feat: add static resource updater script

andrewrobinsonhodges-snyk:
### Description

### Checklist

- [ ] Linted
- [ ] CHANGELOG.md updated

### Screenshots / GIFs

### DIFF
--- a/.github/workflows/resource-check.yml
+++ b/.github/workflows/resource-check.yml
@@ -12,17 +12,6 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Check Static Resources
-        run: |
-          declare -A resources
-          # Add each resource as a key, value pair, mapping the local resource to the reference file (which should be stored in the language server repository). For example:
-          # resources["<path_to_local_file>"]="<url_of_reference_file>"
-          resources["src/main/resources/html/ScanSummaryInit.html"]="https://raw.githubusercontent.com/snyk/snyk-ls/refs/heads/main/shared_ide_resources/ui/html/ScanSummaryInit.html"
-          resources["src/main/resources/html/settings-fallback.html"]="https://raw.githubusercontent.com/snyk/snyk-ls/refs/heads/main/shared_ide_resources/ui/html/settings-fallback.html"
-          for key in ${!resources[@]}; do
-            candidate=$(sha512sum $key | awk {'print $1'})
-            candidate=${candidate:="null"}
-            reference=$(curl -s ${resources[$key]} | sha512sum | awk {'print $1'})
-            echo "Candidate file $key has sha512sum $candidate"
-            echo "Reference file ${resources[$key]} has sha512sum $reference"
-            [[ $candidate == $reference ]]
-          done
+        # The resource map lives in scripts/update-shared-resources.sh (the single
+        # source of truth). Run that script with --check to update the resources locally.
+        run: scripts/update-shared-resources.sh --check
--- a/scripts/update-shared-resources.sh
+++ b/scripts/update-shared-resources.sh
@@ -1,0 +1,101 @@
+#!/usr/bin/env bash
+#
+# Updates (or verifies) the static HTML resources that are vendored from the
+# Snyk Language Server (snyk-ls) repository.
+#
+# These files are copies of the canonical versions in snyk-ls under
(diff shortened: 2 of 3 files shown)


.github/workflows/resource-check.yml
```diff
@@ -12,17 +12,6 @@ jobs:
...
-          for key in ${!resources[@]}; do
-            candidate=$(sha512sum $key | awk {'print $1'})
-            candidate=${candidate:="null"}
-            reference=$(curl -s ${resources[$key]} | sha512sum | awk {'print $1'})
-            echo "Candidate file $key has sha512sum $candidate"
-            echo "Reference file ${resources[$key]} has sha512sum $reference"
-            [[ $candidate == $reference ]]
-          done
+        # The resource map lives in scripts/update-shared-resources.sh (the single
+        # source of truth). Run that script with --check to update the resources locally.
```
bastiandoetsch, line 16: [Suggestion] This comment is backwards: `--check` only *verifies* (no writes); it's the no-argument invocation that updates resources locally. As written it would send a maintainer to run `--check` and see nothing change. Suggest: "CI verifies with `--check`; run the script with no arguments to update resources locally."

— AI review

bastiandoetsch:
Otherwise this is a clean, net-positive refactor — the new `update-shared-resources.sh` is safe (`set -euo pipefail`, quoted, `curl -fsSL` to a temp file then hash/cp — no `curl|bash`, hardcoded HTTPS GitHub-raw URLs, byte-exact hashing that fixes the old command-substitution newline bug), self-locates via repo root, and is committed `0755`. The `settings-fallback.html` change is vendored byte-for-byte from snyk-ls `main` (verified — `--check` exits 0), so its autosave/visibilitychange behaviour is upstream-owned, not authored here. Security scan clean.

Minor hardening suggestions (non-blocking):
- [Suggestion] Temp files are keyed on `basename` only — a future `RESOURCES` entry with the same filename in a different dir would silently overwrite the first download. Key the temp name on the full relative path (or per-iteration `mktemp`).
- [Suggestion] In `--check` mode a `curl` failure (transient/404/upstream rename) increments `failures` and prints "out of sync — run the updater", which won't fix a network/404. Distinguish download failures from drift in the summary.
- [Suggestion] Add `curl --retry 3 --retry-delay 2 --max-time 30` to match the stated robustness goal.
- [Suggestion · CI hardening] `actions/checkout@v4` is a mutable major tag (pre-existing, repo-consistent); pinning to a 40-char SHA is the stricter posture.

— AI review

bastiandoetsch (approved):
Full verification pass (semantic, adversarial, security, code-review). The script is a genuine robustness improvement over the old inline workflow block, CI wiring is correct, and the vendored HTML matches snyk-ls main byte-for-byte so it won't trip the drift check. No blocking findings — one backwards comment and a few non-blocking hardening suggestions left inline/below. Approving.

— AI review

.github/workflows/resource-check.yml
```diff
@@ -12,17 +12,6 @@ jobs:
...
+        run: scripts/update-shared-resources.sh --check
```
basti-snyk, line 17: [Should Fix] This workflow declares no `permissions:` block, so it inherits the repository/organization default `GITHUB_TOKEN` scopes — which for many repos is still read/write across contents, packages, etc. This job only needs to read the checked-out tree and run a verification script; it never writes back. Add a least-privilege token scope so a compromised action or dependency can't push commits, publish packages, or mutate the repo:

```yaml
jobs:
  static-resource-checks:
    runs-on: ubuntu-latest
    permissions:
      contents: read
```

(Top-level `permissions:` works too and covers any future jobs.)

[Suggestion] Related supply-chain hardening: the resources are fetched from `snyk-ls` `main` (a moving target) and validated only against whatever `main` currently serves — there's no pin to a specific upstream commit/hash, so an upstream change silently becomes the new "correct" bytes. Consider pinning the fetch to a commit SHA (or recording the expected upstream revision) so drift is an explicit, reviewable bump rather than an invisible one.

— AI review

rrama:
@andrewrobinsonhodges-snyk Is this no longer needed given the script in snyk-ls or is there a benefit here?

PR #844, opened 2026-06-17
feat: per-folder reset overrides [IDE-1945]

nick-y-snyk:
### Description

IDE-side support for the per-folder **Reset overrides** button (the button + dialog are served by snyk-ls). See snyk/snyk-ls#1344 for the LS side.

A folder field sent by the dialog as JSON `null` means "reset this override" → the LS `Unset`s the `user:folder:` override (fallback to org/LDX/default). Native nullable types cannot distinguish present-`null` from absent, so the plugin re-parses the raw JSON tree and forwards `{value:null, changed:true}` for it.

The IDE does **not** keep a reset-key whitelist mirroring snyk-ls's `FOLDER_RESET_FIELDS`. It forwards every present-`null` folder field verbatim (only `folderPath` is excluded) and lets snyk-ls ignore any key it doesn't recognize — this avoids a second list that would drift from the LS's. As a result no IDE change is needed when snyk-ls adds resettable fields (e.g. the recently added `additional_parameters`, `additional_environment`, `scan_command_config` are already covered). One side effect: an unknown present-`null` key still mutates IDE-side stored state (`withoutSetting` + `clearExplicitlyChanged`) for a key the LS ignores. See the commit bodies for plugin-specific details (naming fixes, stored-override clearing).

### Checklist

- [ ] README updated, if user-facing


src/main/kotlin/io/snyk/plugin/ui/jcef/SaveConfigHandler.kt
```diff
@@ -539,6 +545,56 @@ class SaveConfigHandler(
...
+        folderObject.get("folderPath")?.takeIf { it.isJsonPrimitive }?.asString ?: continue
+
+      var updated = fcs.getFolderConfig(folderPath)
+      var changed = false
+      for ((key, value) in folderObject.entrySet()) {
+        if (key == "folderPath") continue
+        if (!value.isJsonNull) continue
+        updated = updated.withoutSetting(key)
+        settings.clearExplicitlyChanged(folderPath, key)
+        settings.addPendingFolderReset(folderPath, key)
```
basti-snyk: **Should Fix — two latent ways the folder reset is keyed inconsistently and can silently no-op:**

1. **Path-normalization asymmetry.** Here the queue and explicit-change flag are keyed by the *raw* JSON `folderPath`, but `FolderConfigSettings.getFolderConfig/addFolderConfig` normalize via `Paths.get(...).normalize().toAbsolutePath()`, and `LanguageServerWrapper.applyPendingFolderResets` merges by the *normalized* outbound `folderPath` (`associateBy { it.folderPath }`). If the dialog ever sends a non-normalized path (trailing slash, `file://` URI, relative segment, case-variant on a case-insensitive FS), the merge misses and emits a duplicate unnormalized folder config while the real folder's override re-asserts. Normalizing once at the top of `applyFolderResetsFromRawJson` (or having the store own normalization at `addPendingFolderReset`) gives one canonical key everywhere it's read.

2. **camelCase key mismatch.** The loop forwards `folderObject.entrySet()` keys verbatim to `withoutSetting(key)` / `addPendingFolderReset(key)`, but `FolderConfigData` declares camelCase `@SerializedName(alternate=[...])` (e.g. `snykCodeEnabled`, `additionalParameters`) precisely so a producer may send camelCase, while stored overrides + `LsFolderSettingsKeys` are snake_case. A `{"snykCodeEnabled": null}` reset would be a no-op (stored under `snyk_code_enabled`) and emit a key snyk-ls doesn't recognize. Map raw keys through the same alias resolution before queuing.

Every new test pre-normalizes its path and uses snake_case, so both gaps are uncovered — worth a test driving a trailing-slash/URI path and a camelCase null.

— AI review

nick-y-snyk: Fixed both.

1. **Path:** normalize `folderPath` once at the top of `applyFolderResetsFromRawJson` so the reset queue + explicit-change flags key identically to the store (`FolderConfigSettings`) and the outbound merge (`applyPendingFolderResets`). Added a test driving a trailing-slash/`.`-segment path.

2. **camelCase:** confirmed via the snyk-ls form payload that folder fields are always sent snake_case (vscode relies on the same), so the camelCase `@SerializedName` alternates on `FolderConfigData` had no producer — deleted them. Verbatim-forwarded reset keys are now provably snake_case LS keys, so no alias resolution or whitelist is needed (keeps the documented "snyk-ls is authoritative, no IDE whitelist" design).

src/main/kotlin/io/snyk/plugin/ui/jcef/SaveConfigHandler.kt
```diff
@@ -560,12 +566,77 @@ class SaveConfigHandler(
...
+      // case variant) would otherwise miss the merge and emit a duplicate folder config.
+      val folderPath = fcs.normalizePath(rawFolderPath)
+
+      var updated = fcs.getFolderConfig(folderPath)
+      var changed = false
+      for ((key, value) in folderObject.entrySet()) {
+        if (key == "folderPath") continue
+        if (!value.isJsonNull) continue
+        updated = updated.withoutSetting(key)
+        settings.clearExplicitlyChanged(folderPath, key)
```
basti-snyk: **Should Fix — `folderExplicitChanges` is mutated here on the JCEF save thread but is not synchronized.**

This re-review is against the new commits since my last pass — nice work addressing the path-normalization and multi-project-drain feedback (the `normalizePath` call + `pendingResetsLock` + `ownedFolderPaths` scoping all landed). One concurrency gap remains from the same family:

`clearExplicitlyChanged(folderPath, key)` does a structural `remove` on `folderExplicitChanges` (and removes the outer entry when the inner set empties). It now runs on the JCEF/CEF callback thread during a save, while the multi-project `getSettings()` fan-out reads the same map via `applyPersistedFolderChangedFlags` / `isExplicitlyChanged` on other threads. `folderExplicitChanges` is a plain `mutableMapOf()`, so concurrent structural `remove` against `get`/iteration can throw `ConcurrentModificationException`, corrupt buckets, or drop a folder's flags.

This PR correctly put `pendingResets` / `pendingFolderResets` behind `pendingResetsLock`, but left the adjacent `folderExplicitChanges` (and `explicitChanges`) unguarded — and this reset path adds a new structural-removal writer on the JCEF thread, widening the window. Suggest guarding `folderExplicitChanges`/`explicitChanges` access with the same lock (or a concurrent map, being careful about the empty-set cleanup), consistent with how the pending-reset maps are now protected.

— AI review

src/main/kotlin/io/snyk/plugin/ui/jcef/SaveConfigHandler.kt
```diff
@@ -560,12 +566,77 @@ class SaveConfigHandler(
...
+      val folderObject = element.asJsonObject
+      val rawFolderPath =
+        folderObject.get("folderPath")?.takeIf { it.isJsonPrimitive }?.asString ?: continue
+      // Normalize once so the reset queue and explicit-change flags are keyed identically to the
+      // store (FolderConfigSettings) and the outbound merge in applyPendingFolderResets, which
+      // both key by the normalized path. A raw non-normalized path (trailing slash, file:// URI,
+      // case variant) would otherwise miss the merge and emit a duplicate folder config.
+      val folderPath = fcs.normalizePath(rawFolderPath)
+
+      var updated = fcs.getFolderConfig(folderPath)
```
basti-snyk, line 637: **Should Fix — `getFolderConfig` has a read side-effect that materializes a default config before the reset is even confirmed.**

`getFolderConfig(folderPath)` falls through to `createEmpty(...)` when the folder is not yet in the store, and `createEmpty` *persists* a full default `LspFolderConfig` (`base_branch=main`, empty params, `preferred_org=""`, `scan_command_config={}`, …) into `configs[]` as a side-effect of the read — it is not a pure getter.

Because this call happens unconditionally here, before the per-field `value.isJsonNull` loop and before the `if (changed)` guard, merely iterating a `folderConfigs` element for a never-configured folder synthesizes and stores a default `user:folder` override that did not exist before. That `if (changed) addFolderConfig(...)` guard is then moot — the entry already landed via the read. The synthesized defaults can subsequently be emitted to the LS as if user-chosen, which is the opposite of a reset.

Suggest only resolving the existing config after detecting at least one present-null field for the folder, or adding a non-mutating read variant to `FolderConfigSettings` and using it here.

— AI review

src/main/kotlin/io/snyk/plugin/ui/jcef/SaveConfigHandler.kt
```diff
@@ -560,12 +566,77 @@ class SaveConfigHandler(
...
+   * For each reset field we: remove the stored override (so it can't re-assert on a later sync),
+   * clear its persisted folder explicit-change flag, and queue a pending folder reset that
+   * [LanguageServerWrapper.getSettings] emits as `{value: null, changed: true}`.
+   */
+  private fun applyFolderResetsFromRawJson(jsonString: String) {
+    val root =
+      try {
+        gson.fromJson(jsonString, JsonObject::class.java)
+      } catch (e: JsonSyntaxException) {
+        logger.warn("Could not parse JSON for folder resets: ${jsonString.take(200)}", e)
```
basti-snyk: **Suggestion (non-blocking, log hygiene) — this logs up to 200 chars of the raw save payload, which can contain the auth token.**

The `SaveConfigRequest` payload includes a `token` field (the Snyk auth token). JSON field order is producer-controlled and not guaranteed, so on a malformed payload the first 200 chars logged here could include the token at `warn` level. This mirrors the pre-existing pattern at line ~163, so it is not a new posture regression — but while this path is being added, consider logging only `e.message` + payload length, or redacting `token`/`api_endpoint` before logging (and ideally back-fixing the older call site too).

— AI review

src/main/kotlin/io/snyk/plugin/ui/jcef/SaveConfigHandler.kt
nick-y-snyk: Guarded explicitChanges/folderExplicitChanges with the existing pendingResetsLock across all six accessors (mark/clear/clearAll/isExplicitlyChanged), so the structural map removal on the JCEF save thread can no longer race the multi-project getSettings() reads. Kept the persisted vars (IntelliJ-serialized) and reused the same lock rather than swapping to a concurrent map to keep the empty-set cleanup atomic.

src/main/kotlin/io/snyk/plugin/ui/jcef/SaveConfigHandler.kt
nick-y-snyk: Made getFolderConfig a pure read instead of patching the reset path — root cause was createEmpty persisting its default into the store as a side-effect of a read, which affected all 9 callers, not just this one. createEmpty now returns a transient default without writing to configs; the LS remains the sole owner of folder overrides (populated via addAll). Updated the FolderConfigSettings test that codified the persist-on-read to assert the store stays empty after reading an unknown folder.

src/main/kotlin/io/snyk/plugin/ui/jcef/SaveConfigHandler.kt
nick-y-snyk: Done, and back-fixed the older call site too. Both JSON-parse catches now log only the payload length + e.message instead of the first 200 chars, so the auth token can't leak into warn logs regardless of field order.

src/main/kotlin/io/snyk/plugin/ui/jcef/SaveConfigHandler.kt
```diff
@@ -560,12 +568,82 @@ class SaveConfigHandler(
...
+        // See note at the saveConfig parse above: the payload carries the auth token, so log only
+        // its length, not its contents.
+        logger.warn(
+          "Could not parse JSON for folder resets (${jsonString.length} chars): ${e.message}",
+          e,
+        )
+        return
+      } ?: return
+
+    val folderConfigsJson = root.getAsJsonArray("folderConfigs") ?: return
```
basti-snyk: **Critical — uncaught `ClassCastException` on `"folderConfigs": null` aborts the entire save.**

`root.getAsJsonArray("folderConfigs") ?: return` only guards the *absent* case. Gson's `JsonObject.getAsJsonArray` does `(JsonArray) members.get(name)`, and when the key is *present as JSON `null`* `get(name)` returns `JsonNull.INSTANCE` (not Java `null`), so the `?:` never fires and the cast throws `ClassCastException`. That is not a `JsonSyntaxException`, so the `try/catch` around the parse above (lines 592–603) does not catch it — it escapes `applyFolderResetsFromRawJson`, propagates out of `saveConfig`, and aborts the save *before* the open-projects `updateConfiguration()` loop runs, dropping the global settings in the same payload.

`SaveConfigRequest.folderConfigs` is nullable, so an explicit `"folderConfigs": null` is a legal payload a producer can emit; the typed parse tolerates it but this raw re-parse does not.

Suggested guard:
```kotlin
val folderConfigsJson = root.get("folderConfigs")?.takeIf { it.isJsonArray }?.asJsonArray ?: return
```

Worth a test: a save payload with `"folderConfigs": null` should still apply the global settings rather than throw.

— AI review

src/main/kotlin/io/snyk/plugin/ui/jcef/SaveConfigHandler.kt
```diff
@@ -560,12 +568,82 @@ class SaveConfigHandler(
...
+    for (element in folderConfigsJson) {
+      if (!element.isJsonObject) continue
+      val folderObject = element.asJsonObject
+      val rawFolderPath =
+        folderObject.get("folderPath")?.takeIf { it.isJsonPrimitive }?.asString ?: continue
+      // Normalize once so the reset queue and explicit-change flags are keyed identically to the
+      // store (FolderConfigSettings) and the outbound merge in applyPendingFolderResets, which
+      // both key by the normalized path. A raw non-normalized path (trailing slash, file:// URI,
+      // case variant) would otherwise miss the merge and emit a duplicate folder config.
+      val folderPath = fcs.normalizePath(rawFolderPath)
```
basti-snyk: **Should Fix — a single malformed `folderPath` throws out of the whole save (no per-folder isolation).**

Two related gaps on this reset path:

1. **`normalizePath` can throw `InvalidPathException`.** `fcs.normalizePath` is `Paths.get(folderPath).normalize().toAbsolutePath()`, and `Paths.get` throws `InvalidPathException` (an unchecked `RuntimeException`) for a malformed path (e.g. one containing a NUL byte). `rawFolderPath` comes straight off the wire with no validation, so one bad folder entry aborts `applyFolderResetsFromRawJson` and, since nothing catches it, the entire `saveConfig` — discarding the valid folders and the global settings in the same payload.

2. **`isJsonPrimitive` accepts non-string primitives.** `folderObject.get("folderPath")?.takeIf { it.isJsonPrimitive }?.asString` will happily coerce a number/boolean (`{"folderPath": 123}` → `"123"`). That mis-keys the reset under a bogus normalized path that no owned folder matches, so the reset both fails silently and leaks an entry in `pendingFolderResets` for the session.

Suggested fix: require a string (`it.isJsonPrimitive && it.asJsonPrimitive.isString`) and wrap the per-folder `normalizePath` in a `try/catch (InvalidPathException)` that logs and `continue`s, so one bad entry can't sink the rest of the save. The same `normalizePath` call in `applyFolderConfigs` (line 646) has the identical exposure.

— AI review

src/main/kotlin/io/snyk/plugin/ui/jcef/SaveConfigHandler.kt
nick-y-snyk: Fixed — switched to root.get("folderConfigs")?.takeIf { it.isJsonArray }?.asJsonArray ?: return, so a present-as-null folderConfigs returns JsonNull and is skipped instead of throwing a ClassCastException out of the save. Added a test (folderConfigs present as JSON null still applies global settings) asserting the global cliPath still lands. The typed path (applyFolderConfigs via config.folderConfigs?.let) was already safe since Gson maps JsonNull to a Kotlin null there.

src/main/kotlin/io/snyk/plugin/ui/jcef/SaveConfigHandler.kt
nick-y-snyk: Fixed both gaps. Added FolderConfigSettings.normalizePathOrNull (wraps Paths.get, logs + returns null on InvalidPathException) and use it at both save sites (applyFolderResetsFromRawJson and applyFolderConfigs) with `?: continue`, so one malformed path cannot abort the whole save / drop the global settings. Also tightened the reset-path folderPath check to require a JSON string (isJsonPrimitive && asJsonPrimitive.isString) so a number/boolean cannot be coerced and mis-key the queue. Added a test driving a NUL-byte path + a numeric folderPath alongside a valid one, asserting only the valid folder reset is queued.

src/main/kotlin/io/snyk/plugin/services/SnykApplicationSettingsStateService.kt
```diff
@@ -53,46 +53,107 @@ class SnykApplicationSettingsStateService :
...
+        } else {
+          pendingFolderResets.keys.intersect(ownedFolderPaths)
+        }
+      val snapshot = owned.associateWith { pendingFolderResets.getValue(it).toSet() }
+      owned.forEach { pendingFolderResets.remove(it) }
+      snapshot
+    }
+
+  // [explicitChanges]/[folderExplicitChanges] are APP-scoped and touched from multiple threads:
+  // the JCEF save thread mutates them (mark/clear, including structural map removal on reset) while
```
basti-snyk: 🤖 **Automated review (Claude / IDE-1945 verification)** · **Should Fix**

**`pendingResetsLock` does not protect the state serializer — CME risk.**

`getState()` returns `this` (L304), and `explicitChanges`/`folderExplicitChanges` are plain `mutableSetOf()`/`mutableMapOf()` public `var`s that IntelliJ's `XmlSerializerUtil` reflects over **directly** — it never calls these synchronized methods and never holds `pendingResetsLock`. So a save-triggered serialization iterating these maps, concurrent with `clearExplicitlyChanged`'s structural `folderExplicitChanges.remove(...)` (L138), can still throw `ConcurrentModificationException`. This comment's claim that the lock makes a concurrent structural remove safe is only true mutator-vs-mutator, not vs. the serializer.

**Fix:** back both fields with concurrent collections (`ConcurrentHashMap` + `ConcurrentHashMap.newKeySet()`), or have `getState()` return a deep copy taken under `pendingResetsLock`.

src/main/kotlin/io/snyk/plugin/ui/jcef/SaveConfigHandler.kt
```diff
@@ -560,12 +568,97 @@ class SaveConfigHandler(
...
+      for ((key, value) in folderObject.entrySet()) {
+        if (key == "folderPath") continue
+        if (!value.isJsonNull) continue
+        updated = updated.withoutSetting(key)
+        settings.clearExplicitlyChanged(folderPath, key)
+        settings.addPendingFolderReset(folderPath, key)
+        changed = true
+      }
+      if (changed) {
+        fcs.addFolderConfig(updated)
```
basti-snyk, line 650: 🤖 **Automated review (Claude / IDE-1945 verification)** · **Should Fix**

**Resetting a folder with no prior override persists a full default config.**

For an unknown folder, `getFolderConfig(folderPath)` (L634) returns the `createEmpty` default template (`base_branch="main"`, `preferred_org=""`, …) without persisting. The loop strips only the reset keys, then `addFolderConfig(updated)` here **persists** that default template. This re-materializes exactly the default entry the companion `getFolderConfig` change was made to stop persisting — and the path is realistic precisely because that change made reads non-persisting, so a folder shown in settings isn't in `configs` until this reset writes one.

Verified narrower impact: outbound emission gates by `applyPersistedFolderChangedFlags` (LanguageServerWrapper L834), so the default keys go out as `changed=false` and the LS ignores them — this is **persisted-state pollution / invariant violation**, not bogus active overrides reaching the LS.

**Fix:** capture `val existed = fcs.getAll().containsKey(folderPath)` before the loop; only `addFolderConfig(updated)` when `existed`, otherwise just queue the resets and clear flags.

src/main/kotlin/snyk/common/lsp/LanguageServerWrapper.kt
```diff
@@ -834,7 +835,43 @@ class LanguageServerWrapper(private val project: Project) : Disposable {
...
+  ): List<LspFolderConfig> {
+    // Consume only the resets for folders this project owns. The reset queue is APP-scoped but
+    // SaveConfigHandler fans out the save to every open project's LS, so an unscoped drain would
+    // let the first project clear other windows' resets (they would then re-assert the override).
+    val fcs = service<FolderConfigSettings>()
+    val ownedFolderPaths =
+      configuredWorkspaceFolders
+        .map { fcs.normalizePath(it.uri.fromUriToPath().toString()) }
+        .toSet()
+    val pending = ps.consumePendingFolderResets(ownedFolderPaths)
```
basti-snyk: 🤖 **Automated review (Claude / IDE-1945 verification)** · **Should Fix (borderline)**

**A reset for a folder owned by no open window leaks in the APP-scoped queue forever.**

`consumePendingFolderResets(ownedFolderPaths)` drains only resets for folders this project owns. A reset queued for a path that is not a configured workspace folder of any open project (e.g. the save races a folder-close) is never consumed, never emitted, and can fire stale on a later reopen.

**Fix:** after the fan-out, reap pending folder resets for unowned paths, or only queue a reset when the path is a configured workspace folder.

_Related (Suggestion):_ the **global** reset drain (`ps.consumePendingResets()`, ~L819) is still unscoped — the same first-window-wins race this folder path was designed to avoid. Pre-existing for the global queue and low-impact (global settings are identical per window), but worth unifying if global resets ever need per-window delivery.

src/main/kotlin/io/snyk/plugin/ui/jcef/SaveConfigHandler.kt
```diff
@@ -560,12 +568,97 @@ class SaveConfigHandler(
...
+      // both key by the normalized path. A raw non-normalized path (trailing slash, file:// URI,
+      // case variant) would otherwise miss the merge and emit a duplicate folder config.
+      // normalizePathOrNull isolates one malformed path so it can't abort the whole save.
+      val folderPath = fcs.normalizePathOrNull(rawFolderPath) ?: continue
+
+      var updated = fcs.getFolderConfig(folderPath)
+      var changed = false
+      for ((key, value) in folderObject.entrySet()) {
+        if (key == "folderPath") continue
+        if (!value.isJsonNull) continue
```
basti-snyk, line 641: 🤖 **Automated review (Claude / IDE-1945 verification)** · **Suggestion**

**Code vs. PR description: there is no reset-key whitelist.** This loop forwards *every* present-null folder key (only `folderPath` is excluded), relying on snyk-ls to ignore keys it doesn't recognize. That's architecturally better than mirroring snyk-ls's `FOLDER_RESET_FIELDS` (no two-list drift), but the PR body says it "mirrors `FOLDER_RESET_FIELDS` / extends the whitelist" — code and narrative disagree.

**Fix:** update the PR/commit description to state the IDE forwards all present-nulls; add a small **deny-list** only if specific folder keys must never be IDE-resettable. Note a side effect: an unknown present-null key still mutates IDE-side stored state (`withoutSetting` + `clearExplicitlyChanged`) for a key the LS will ignore.

src/main/kotlin/io/snyk/plugin/ui/jcef/SaveConfigHandler.kt
```diff
@@ -560,12 +568,97 @@ class SaveConfigHandler(
...
+   * was stored under — no camelCase alias resolution is needed (see [FolderConfigData]).
+   *
+   * For each reset field we: remove the stored override (so it can't re-assert on a later sync),
+   * clear its persisted folder explicit-change flag, and queue a pending folder reset that
+   * [LanguageServerWrapper.getSettings] emits as `{value: null, changed: true}`.
+   */
+  private fun applyFolderResetsFromRawJson(jsonString: String) {
+    val root =
+      try {
+        gson.fromJson(jsonString, JsonObject::class.java)
```
basti-snyk, line 591: 🤖 **Automated review (Claude / IDE-1945 verification)** · **Suggestion**

**Payload parsed twice.** On every non-fallback save the payload is deserialized once to `SaveConfigRequest` (~L161) and again to a `JsonObject` here, doubling peak parse memory. The payload is locally produced by the JCEF dialog (bounded, trusted), so there's no DoS path — efficiency only.

**Fix (optional):** parse the raw `JsonObject` once at the top of `saveConfig`, derive `SaveConfigRequest` from it via `gson.fromJson(root, …)`, and pass the same tree here.

src/main/kotlin/snyk/common/lsp/settings/FolderConfigSettings.kt
```diff
@@ -36,39 +37,51 @@ class FolderConfigSettings {
...
 
+  /**
+   * Null-safe [normalizePath]. [Paths.get] throws [InvalidPathException] (unchecked) on a malformed
+   * path (e.g. a NUL byte), so callers iterating folder entries can skip one bad entry instead of
+   * aborting the whole loop.
+   */
+  internal fun normalizePathOrNull(folderPath: String): String? =
+    try {
+      normalizePath(folderPath)
+    } catch (e: InvalidPathException) {
```
basti-snyk: 🤖 **Automated review (Claude / IDE-1945 verification)** · **Suggestion**

**`normalizePathOrNull` catches only `InvalidPathException`.** Its purpose is to isolate one bad folder entry so it can't abort the whole save, but `Paths.get(...).toAbsolutePath()` can also throw `IOError` (wrapping a provider `IOException`) or `SecurityException` (under a security manager). Either escapes this `catch` and propagates up through `applyFolderResetsFromRawJson`/`applyFolderConfigs`, aborting the batch this helper was meant to protect.

**Fix:** also handle `IOError` and `SecurityException` (or catch narrowly with logging).

src/test/kotlin/io/snyk/plugin/ui/jcef/SaveConfigHandlerTest.kt
```diff
@@ -1340,6 +1340,309 @@ class SaveConfigHandlerTest : BasePlatformTestCase() {
...
+            { "folderPath": 123, "snyk_code_enabled": null },
+            { "folderPath": ${gson.toJson(goodPath)}, "snyk_code_enabled": null }
+          ]
+        }
+      """
+        .trimIndent()
+
+    invokeParseAndSaveConfig(json)
+
+    val resets = realSettings.consumePendingFolderResets()
```
basti-snyk: 🤖 **Automated review (Claude / IDE-1945 verification)** · **Suggestion (test coverage)**

The new tests stop at `consumePendingFolderResets()` / `FolderConfigSettings` state; the `LanguageServerWrapper` tests start from a manually-seeded `addPendingFolderReset`. Two gaps in the user-visible contract aren't covered:

1. **End-to-end wire shape:** drive a present-null payload through `saveConfig` and assert the resulting `getSettings()` `folderConfigs` carries `{value:null, changed:true}` for that folder.
2. **Two-wrapper shared-queue drain:** assert window B still emits its reset after window A's `getSettings()` runs (the real-world race the owned-folder scoping defends against).

basti-snyk (commented):
🤖 **Automated review (Claude) — IDE-1945 / PR #844 verification**

**Verdict: CONTESTED** — no hard blocker, but two grounded **Should Fix** correctness defects (verified against source), plus a borderline leak and several suggestions. Reviewed by four agents (semantic, adversarial, security, code-review) + CI oracle; the two contested findings were independently re-verified.

**Should Fix**
1. `pendingResetsLock` doesn't cover the IntelliJ state serializer → residual `ConcurrentModificationException` risk; the in-code "can't CME" comment overstates the guarantee. _(SnykApplicationSettingsStateService.kt)_
2. Resetting a folder with **no prior override** persists a full default config, re-materializing the defaults the sibling `getFolderConfig` change removed (persisted-state pollution; not emitted as active LS overrides). _(SaveConfigHandler.kt:645)_

**Should Fix (borderline)**
3. A reset for a folder owned by no open window leaks forever in the APP-scoped queue; can fire stale on reopen. _(LanguageServerWrapper.kt:863)_

**Suggestions:** PR description says "whitelist" but the code has none (the no-whitelist design is the better choice — fix the description); global reset drain still unscoped (pre-existing); payload parsed twice; `normalizePathOrNull` catches only `InvalidPathException`; missing end-to-end + two-wrapper reset tests. See inline comments.

**Solid / verified positives:** auth-token log-leak fix is genuine and complete; `folderConfigs: null` ClassCastException edge correctly guarded + regression-tested; removal of single-folder→global-state mirroring is correct subtraction removing a real singleton-corruption hazard; path normalization consistent; Snyk Code clean (no new issues).

**Oracle/coverage:** Format&Lint, Plugin Verifier, Snyk Code, macOS unit tests green via CI; ubuntu/windows unit legs were pending at review time; local Snyk SCA not run (unchanged `build.gradle.kts`, CI SCA covers it).

src/main/kotlin/io/snyk/plugin/ui/jcef/SaveConfigHandler.kt
```diff
@@ -560,12 +568,97 @@ class SaveConfigHandler(
...
+          ?.takeIf { it.isJsonPrimitive && it.asJsonPrimitive.isString }
+          ?.asString ?: continue
+      // Normalize once so the reset queue and explicit-change flags are keyed identically to the
+      // store (FolderConfigSettings) and the outbound merge in applyPendingFolderResets, which
+      // both key by the normalized path. A raw non-normalized path (trailing slash, file:// URI,
+      // case variant) would otherwise miss the merge and emit a duplicate folder config.
+      // normalizePathOrNull isolates one malformed path so it can't abort the whole save.
+      val folderPath = fcs.normalizePathOrNull(rawFolderPath) ?: continue
+
+      var updated = fcs.getFolderConfig(folderPath)
```
basti-snyk, line 637: **Should Fix — the reset path re-materializes a full synthetic default override for a folder that had none, re-introducing one layer up the persist-on-read footgun this PR removed from `getFolderConfig`.**

Trace a reset on a folder with **no stored override** (e.g. `{folderPath: <fresh dir>, snyk_code_enabled: null}`):
- `var updated = fcs.getFolderConfig(folderPath)` returns the transient `createEmpty(...)` default (`base_branch="main"`, `additional_parameters=[]`, `preferred_org=""`, `scan_command_config={}`, …) — now a pure read that does *not* persist (good, that was b09b2a16).
- `snyk_code_enabled` is not among `createEmpty`'s keys, so `withoutSetting(key)` early-returns the config unchanged, **but `changed` is still set to `true`**.
- `if (changed) fcs.addFolderConfig(updated)` then persists that entire synthetic default as a real `user:folder` override the user never set.

Net effect: a *reset* on a never-configured folder materializes a stored override (and later emits `base_branch="main"` etc.), the opposite of a reset — re-creating the exact "persist a default on a folder the user never touched" invariant `FolderConfigSettings.getFolderConfig` was just hardened against. The existing `saveConfig folder field sent as null …` test uses a fresh folder but never asserts the store stays empty, so it passes through this.

Fix direction: gate the persist on the folder actually being stored (e.g. only `addFolderConfig` when `fcs.getAll().containsKey(folderPath)`), or build `updated` from the stored config and skip `addFolderConfig` when there was no prior entry; and only set `changed=true` when `withoutSetting` actually removed a key. The pending-reset enqueue + explicit-change clear should still run regardless. Add a test asserting `fcs.getAll()` stays empty after resetting a field on a folder with no prior override.

— AI review

src/main/kotlin/snyk/common/lsp/LanguageServerWrapper.kt
```diff
@@ -834,7 +835,43 @@ class LanguageServerWrapper(private val project: Project) : Disposable {
...
+    folderConfigs: List<LspFolderConfig>,
+    ps: SnykApplicationSettingsStateService,
+  ): List<LspFolderConfig> {
+    // Consume only the resets for folders this project owns. The reset queue is APP-scoped but
+    // SaveConfigHandler fans out the save to every open project's LS, so an unscoped drain would
+    // let the first project clear other windows' resets (they would then re-assert the override).
+    val fcs = service<FolderConfigSettings>()
+    val ownedFolderPaths =
+      configuredWorkspaceFolders
+        .map { fcs.normalizePath(it.uri.fromUriToPath().toString()) }
```
basti-snyk: **Should Fix — this `getSettings()` path uses the throwing `normalizePath` (not `normalizePathOrNull`), so one malformed workspace-folder path aborts the entire settings build.**

`applyPendingFolderResets` builds `ownedFolderPaths` via `fcs.normalizePath(it.uri.fromUriToPath().toString())`. `normalizePath` is `Paths.get(...).normalize().toAbsolutePath()`, which throws unchecked `InvalidPathException` on a malformed path — the exact reason this PR introduced `normalizePathOrNull` and used it on the save side (`SaveConfigHandler.kt:632`). `getSettings()` has no enclosing try/catch, so a single bad workspace-folder path propagates out and drops **all** settings (global + every folder) to the LS, instead of skipping the one bad folder as the rest of the PR carefully does.

Reachability is low (workspace folders are IDE-derived roots), but it is inconsistent with the PR's own defensive pattern. Fix: use `normalizePathOrNull` here and filter nulls, matching the save path. (Minor adjacent: `ownedFolderPaths` is normalized on every `getSettings()` even when the pending-reset queue is empty — the common case — so a short-circuit on an empty queue would avoid the work.)

— AI review

src/main/kotlin/io/snyk/plugin/ui/jcef/SaveConfigHandler.kt
nick-y-snyk: Fixed. Two persist sites, not one: `applyFolderConfigs` (runs first on the typed path — what actually materialized the default for an all-null reset payload) and `applyFolderResetsFromRawJson`. Both now skip persisting a synthetic default for a never-configured folder; the reset still ships via the pending queue. `applyFolderConfigs` gates on `updated !== existing` (a setting was actually applied), `applyFolderResetsFromRawJson` on `existed && changed` (createEmpty pre-populates the reset keys, so an identity check alone isn't safe there). Note this isn't VS Code's map-over-stored shape on purpose: IntelliJ's FolderConfigSettings is seeded asynchronously by the LS (`$/snyk.folderConfigs`) and can lag, so skipping unknown inbound folders would drop a user's first-time config on a not-yet-seeded folder — the `updated !== existing` guard persists real changes regardless of prior existence while still dropping the no-op default. Added an assertion that the fresh folder stays out of the store after a reset.

src/main/kotlin/io/snyk/plugin/ui/jcef/SaveConfigHandler.kt
nick-y-snyk: Fixed. Two persist sites, not one: `applyFolderConfigs` (runs first on the typed path — what actually materialized the default for an all-null reset payload) and `applyFolderResetsFromRawJson`. Both now skip persisting a synthetic default for a never-configured folder; the reset still ships via the pending queue. `applyFolderConfigs` gates on `updated !== existing` (a setting was actually applied), `applyFolderResetsFromRawJson` on `existed && changed` (createEmpty pre-populates the reset keys, so an identity check alone isn't safe there). Note this isn't VS Code's map-over-stored shape on purpose: IntelliJ's FolderConfigSettings is seeded asynchronously by the LS (`$/snyk.folderConfigs`) and can lag, so skipping unknown inbound folders would drop a user's first-time config on a not-yet-seeded folder — the `updated !== existing` guard persists real changes regardless of prior existence while still dropping the no-op default. Added an assertion that the fresh folder stays out of the store after a reset.

src/main/kotlin/io/snyk/plugin/services/SnykApplicationSettingsStateService.kt
nick-y-snyk: Fixed via the concurrent-collections route (not the `getState()` deep-copy, which would break the documented `getState() === this` contract). `explicitChanges`/`folderExplicitChanges` and the inner key sets are now `ConcurrentHashMap.newKeySet()` / `ConcurrentHashMap`-backed — weakly-consistent iteration means `XmlSerializerUtil` reflecting over them off-lock can't CME against a concurrent structural `remove`. Declared types stay `MutableSet`/`MutableMap` so XML serialization is unchanged (verified by `loadState_copiesBeanProperties`). The `pendingResetsLock` blocks stay for the compound check-then-remove invariants, which concurrent collections alone don't make atomic.

src/main/kotlin/snyk/common/lsp/LanguageServerWrapper.kt
nick-y-snyk: Fixed — `applyPendingFolderResets` now builds `ownedFolderPaths` via `normalizePathOrNull` + `mapNotNull`, so a single malformed workspace-folder path is skipped instead of throwing `InvalidPathException` out of `getSettings` and dropping all settings. Matches the save-side pattern. Skipped the empty-queue short-circuit: normalizing a handful of workspace folders per `getSettings` is negligible and a peek method would add API surface for no measurable win — happy to add it if you'd rather.

src/main/kotlin/snyk/common/lsp/LanguageServerWrapper.kt
nick-y-snyk: Won't fix — not reachable. Folder configs reach the IDE only via the `$/snyk.configuration` notification (`LspConfigurationParam.folderConfigs` -> `FolderConfigSettings.addAll`), and a folder reset is only ever produced by the snyk-ls-served HTML settings page, which renders the folders snyk-ls already sent. So any reset's folderPath belongs to a workspace folder the saving project owns by construction and is consumed on that same save's fan-out — nothing queues a reset for a folder no open window owns. (The related unscoped global-reset note is pre-existing and low-impact as you note.)

src/main/kotlin/snyk/common/lsp/settings/FolderConfigSettings.kt
nick-y-snyk: Widened `normalizePathOrNull` to also swallow `SecurityException` (caught via `RuntimeException`, which still covers `InvalidPathException`) and `IOError`, each logged and skipped, so none can escape and abort the batch the helper guards.

src/main/kotlin/io/snyk/plugin/ui/jcef/SaveConfigHandler.kt
nick-y-snyk: Updated the PR description to match the code: the IDE keeps no `FOLDER_RESET_FIELDS` mirror/whitelist — it forwards every present-`null` folder field verbatim (only `folderPath` excluded) and lets snyk-ls ignore unrecognized keys, so no IDE change is needed when the LS adds resettable fields. Noted the side effect (an unknown present-null key still mutates IDE stored state for a key the LS ignores). Skipped the deny-list — no folder key currently needs to be permanently IDE-unresettable; easy to add if one shows up.

src/main/kotlin/io/snyk/plugin/ui/jcef/SaveConfigHandler.kt
nick-y-snyk: Won't do — efficiency only on a cold path. The payload is a settings form produced by the JCEF dialog: bounded, trusted, parsed once per explicit Save click, so the second parse is microseconds and a few KB with no DoS path. Threading one `JsonObject` through `saveConfig` -> `applyFolderConfigs` / `applyFolderResetsFromRawJson` would change two private signatures for no measurable gain. Happy to revisit if the payload ever grows or this becomes hot.

src/test/kotlin/io/snyk/plugin/ui/jcef/SaveConfigHandlerTest.kt
nick-y-snyk: Added the two-window drain test (`getSettings drains only owned folder resets so another window still emits its own`) in `LanguageServerWrapperTest`: window A owns folderA, window B owns folderB, both resets queued on the shared APP queue; A's `getSettings()` drains only folderA and B still emits folderB — the owned-folder scoping the PR added. Also stubbed `normalizePathOrNull` in that test's setUp (the relaxed mock was returning null after the getSettings path moved to `normalizePathOrNull`, which would have silently dropped owned paths).

On the e2e wire-shape: the payload->queue half is covered by `SaveConfigHandlerTest.test saveConfig folder field sent as null queues a pending folder reset` (asserts `consumePendingFolderResets()` carries the keys) and the queue->wire half by `LanguageServerWrapperTest.getSettings emits pending folder reset as value null changed true` (asserts `{value:null, changed:true}`). Those meet at the exact `addPendingFolderReset` seam `saveConfig` uses, so the chain is covered end to end across the two CUTs; a single cross-class test would need a real wrapper wired into the JCEF fan-out for no additional behavior coverage. Can add it if you'd prefer one literal end-to-end test.

src/main/kotlin/io/snyk/plugin/services/SnykApplicationSettingsStateService.kt
```diff
@@ -46,53 +47,121 @@ class SnykApplicationSettingsStateService :
...
   var fileListenerEnabled: Boolean = true
 
-  var explicitChanges: MutableSet<String> = mutableSetOf()
+  // Concurrent-collection backed (not plain mutableSetOf/mutableMapOf): XmlSerializerUtil reflects
+  // over these public vars directly to serialize state and never holds [pendingResetsLock], so a
+  // save-triggered serialization can iterate them concurrently with a structural remove in
+  // clearExplicitlyChanged. ConcurrentHashMap iteration is weakly-consistent → no CME. The
+  // synchronized blocks below remain for the compound check-then-remove invariants (the lock makes
+  // those atomic; the concurrent backing makes the serializer race safe).
+  var explicitChanges: MutableSet<String> = ConcurrentHashMap.newKeySet()
```
basti-snyk, line 60: **Should Fix — `loadState`/`copyBean` discards the `ConcurrentHashMap` backing this change introduces.**

The new concurrent backing here (and at the comment above) exists so `XmlSerializerUtil` can iterate `explicitChanges`/`folderExplicitChanges` off-lock during a save while another thread structurally mutates them in `clearExplicitlyChanged`/`clearAllExplicitlyChanged`. But `loadState(...)` runs `XmlSerializerUtil.copyBean(state, this)`, which assigns the deserialized collection references directly — and `XmlSerializer` instantiates plain `HashSet`/`HashMap` for these field types. After the first `loadState` (every IDE start with persisted settings), these fields are no longer concurrent, so the off-lock serializer iteration the rationale relies on can throw `ConcurrentModificationException`. The `synchronized(pendingResetsLock)` blocks don't help because the serializer never takes that lock.

Fix direction: in `loadState`, after `copyBean`, deep-rewrap into `ConcurrentHashMap.newKeySet()` / `ConcurrentHashMap` (including the inner sets), or serialize from a snapshot taken under `pendingResetsLock`. A test that calls `loadState` with a plain-`HashMap`/`HashSet` state and asserts the live fields are still concurrent would fail today.

— AI review

src/main/kotlin/io/snyk/plugin/ui/jcef/SaveConfigHandler.kt
```diff
@@ -560,12 +568,104 @@ class SaveConfigHandler(
...
+      // normalizePathOrNull isolates one malformed path so it can't abort the whole save.
+      val folderPath = fcs.normalizePathOrNull(rawFolderPath) ?: continue
+
+      // Only an already-stored folder may be rewritten. getFolderConfig synthesizes a default
+      // template for an unknown folder (a pure read since b09b2a16); persisting that here would
+      // materialize a full default user:folder override for a folder the user never configured —
+      // the opposite of a reset. The reset still reaches the LS via addPendingFolderReset below,
+      // so a never-stored folder needs no local write. Matches the VS Code path, which maps over
+      // the already-stored folder configs and never creates an entry for an unknown folder.
+      val existed = fcs.getAll().containsKey(folderPath)
```
basti-snyk: **Suggestion (efficiency) — `fcs.getAll()` copies the whole map per folder element just to do one `containsKey`.**

`getAll()` returns a fresh `HashMap(configs)` (a full copy of the backing `ConcurrentHashMap`), and this runs once per folder element in the loop — O(N·M) for N elements and M stored configs, all to answer a single membership check. Since `configs` is already a `ConcurrentHashMap`, exposing a cheap `contains(path)` on `FolderConfigSettings` (direct `containsKey`, O(1), thread-safe) avoids the repeated full-map copy.

— AI review

src/main/kotlin/io/snyk/plugin/services/SnykApplicationSettingsStateService.kt
```diff
@@ -46,53 +47,59 @@ class SnykApplicationSettingsStateService :
...
+  // APP-scoped and mutated from multiple threads (JCEF query thread enqueues; every
+  // getSettings() across the multi-project fan-out consumes), so guard all access with
+  // [pendingResetsLock] — the consume is a compound snapshot-then-clear that must be atomic.
   @Transient private val pendingResets: MutableSet<String> = mutableSetOf()
 
+  @Transient private val pendingResetsLock = Any()
+
   fun addPendingReset(key: String) {
-    pendingResets.add(key)
+    synchronized(pendingResetsLock) { pendingResets.add(key) }
```
basti-snyk: Suggestion: use a ConcurrentSet or something like that to not having to synchronize all operations.

src/main/kotlin/io/snyk/plugin/services/SnykApplicationSettingsStateService.kt
```diff
@@ -46,53 +47,59 @@ class SnykApplicationSettingsStateService :
...
+  // writer of a stored folder config sets changed=true on the LspFolderConfig itself
+  // ([snyk.common.lsp.settings.withSetting]), which getSettings emits verbatim — there is no
+  // separate folder changed-flag map to keep in sync.
   fun markExplicitlyChanged(settingKey: String) {
-    explicitChanges.add(settingKey)
-  }
-
-  fun markExplicitlyChanged(folderPath: String, settingKey: String) {
-    folderExplicitChanges.getOrPut(folderPath) { mutableSetOf() }.add(settingKey)
+    synchronized(pendingResetsLock) { explicitChanges.add(settingKey) }
```
basti-snyk: Suggestion: `explicitChanges` is a ConcurrentHashMap - so you don't need to synchronize it. Remove the synchronize when modifying `explicitChanges` here and in all other things.

src/main/kotlin/io/snyk/plugin/ui/jcef/ConfigurationDataClasses.kt
```diff
@@ -118,47 +118,33 @@ data class SaveConfigRequest(
   @SerializedName("isFallbackForm") val isFallbackForm: Boolean? = null,
 )
 
+// snyk-ls serves the settings form and sends every folder field as snake_case (the LS key
+// names). The camelCase alternates that used to live here had no producer, so reset keys
+// forwarded verbatim by SaveConfigHandler are guaranteed snake_case and snyk-ls-recognized.
 data class FolderConfigData(
   @SerializedName("folderPath") val folderPath: String,
-  @SerializedName(value = "additional_parameters", alternate = ["additionalParameters"])
+  @SerializedName("additional_parameters")
```
basti-snyk, line 126: removing the alternate names: is that save when migrating from an old folder config?

src/main/kotlin/io/snyk/plugin/ui/jcef/SaveConfigHandler.kt
```diff
@@ -547,50 +554,122 @@ class SaveConfigHandler(
...
+    // a plain `?: return` would miss it and the cast would throw ClassCastException — which is not
+    // a
+    // JsonSyntaxException, so it would escape the catch above and abort the whole save (dropping
+    // the
+    // global settings in the same payload). Take the element only when it is actually an array.
+    val folderConfigsJson =
+      root.get("folderConfigs")?.takeIf { it.isJsonArray }?.asJsonArray ?: return
+    val fcs = service<FolderConfigSettings>()
+
+    for (element in folderConfigsJson) {
```
basti-snyk, line 613: this is very ugly. can we use typed objects here?

src/main/kotlin/snyk/common/lsp/settings/FolderConfigSettings.kt
```diff
@@ -36,39 +37,56 @@ class FolderConfigSettings {
...
+      normalizePath(folderPath)
+    } catch (e: RuntimeException) {
+      logger.warn("Skipping malformed folder path: ${e.message}")
+      null
+    } catch (e: IOError) {
+      logger.warn("Skipping folder path (I/O error): ${e.message}")
+      null
+    }
+
+  // Pure read: returns the stored override or a transient default. The LS owns folder overrides
```
basti-snyk: pure read is a bit misleading, as it's creating an instance, but not blocking.

src/main/kotlin/io/snyk/plugin/ui/jcef/ConfigurationDataClasses.kt
nick-y-snyk: ls handles folder config migration

src/main/kotlin/io/snyk/plugin/services/SnykApplicationSettingsStateService.kt
nick-y-snyk: Done — dropped `pendingResetsLock` entirely. Both `explicitChanges` and `pendingResets` are now `ConcurrentHashMap.newKeySet()`, so all accessors are lock-free. The one compound op (`consumePendingResets` snapshot-then-clear) drains by remove-while-iterate, which is atomic without a lock and can't drop a concurrently-added key. Also re-wrap `explicitChanges` after `copyBean` in `loadState` so the concurrent backing survives IDE restart.

src/main/kotlin/io/snyk/plugin/services/SnykApplicationSettingsStateService.kt
nick-y-snyk: Done — `pendingResets` is now `ConcurrentHashMap.newKeySet()` and the lock is gone; the snapshot-then-clear consume is kept atomic via remove-while-iterate rather than synchronization.

src/main/kotlin/io/snyk/plugin/ui/jcef/SaveConfigHandler.kt
nick-y-snyk: No Gson annotation distinguishes present-null from absent — per the Gson user guide, deserialization maps both a missing entry and an explicit `null` to the field's default (`null` for objects). So a typed `FolderConfigData` parse can't tell `{"snyk_code_enabled": null}` (a reset) from an omitted field; that's the whole reason this path re-walks the raw tree and keys off `value.isJsonNull`.

The only annotation-free typed alternative is declaring each reset-eligible field as `JsonElement?` (absent → `null`, present-null → `JsonNull`), or a custom `TypeAdapter` with a tri-state wrapper — both spread the present-null handling across the entire DTO and every consumer, which is more code than this one localized loop. Happy to file a follow-up if we want the typed tri-state model despite the extra surface.

src/main/kotlin/snyk/common/lsp/settings/FolderConfigSettings.kt
nick-y-snyk: Reworded — `// Pure read` → `// Non-persisting read`, and called out that it does construct a transient default instance, it just never writes to the store.

src/main/kotlin/snyk/common/lsp/LanguageServerWrapper.kt
```diff
@@ -820,40 +818,24 @@ class LanguageServerWrapper(private val project: Project) : Disposable {
...
-        .map {
-          val folderPath = it.uri.fromUriToPath().toString()
-          val config = service<FolderConfigSettings>().getFolderConfig(folderPath)
-          applyPersistedFolderChangedFlags(config, ps)
-        }
+        // Normalize the lookup key the same way updateFolderConfigRefresh stores it, else a folder
+        // whose URI→path isn't already normalized (trailing slash, "." segment) misses the refresh
+        // flag and its settings — and any reset — are silently dropped.
+        .mapNotNull { folderConfigSettings.normalizePathOrNull(it.uri.fromUriToPath().toString()) }
+        .filter { folderConfigsRefreshed[it] == true }
```
basti-snyk: **Should Fix — a folder reset is silently dropped unless that folder's `folderConfigsRefreshed` flag is already `true`, and the reset path never sets it.**

This filter only emits folder configs whose refresh flag is `true`. The flag is set exclusively by `updateFolderConfigRefresh(path, true)` on the LS push-back path, and it is cleared on (re)connect (`folderConfigsRefreshed.clear()` at ~L196). `applyFolderResetsFromRawJson` writes the reset `{value:null, changed:true}` into `FolderConfigSettings` but never touches the refresh flag.

Failure path: user clicks "Reset overrides" for a folder this session before snyk-ls has echoed `$/snyk.configuration` for it (flag still false/absent) — e.g. right after a reconnect that cleared the map. The reset is stored, then this filter drops the folder from `getSettings()`, so `{value:null, changed:true}` never reaches the LS and the `user:folder` override silently persists. The comment just above even notes resets "are silently dropped" for the normalization mismatch case — the same gate drops them whenever the flag isn't already true.

Note the asymmetry with the global reset path (#847), which is not gated by any refresh flag. Either set/bypass the gate for a folder that has a pending reset, or add an acceptance test that proves a reset is delivered for a folder whose refresh flag was not set this session.

— AI review

nick-y-snyk: Good call — deleted the flag entirely instead of patching the gate. The `folderConfigsRefreshed` map was an IntelliJ-only invention doing a job the per-setting `changed` flag already does on the LS side, and it was the root of this whole class (this reset-drop + the earlier normalization-mismatch drop).

Now `getSettings()` emits every configured folder, exactly like VS Code's `resolveFolderConfigs`. An unrefreshed folder ships `createEmpty` defaults with `changed=null`, which snyk-ls ignores (applies folder settings only when `changed=true`), so it's harmless; a reset ships `{value:null, changed:true}` and reaches the LS even before the LS has echoed that folder's config — no gate left to drop it.

Removed: the map, its reconnect `clear()`, the getSettings gate, `getFolderConfigsRefreshed`/`updateFolderConfigRefresh`, and the refresh-update loop in `SnykLanguageClient`. The settings-dialog readiness check now keys off `FolderConfigSettings.getAll().isNotEmpty()`. Rewrote the old `only refreshed folders` test into one asserting every configured folder is emitted including a reset the LS never echoed — the acceptance test you asked for.

src/main/kotlin/io/snyk/plugin/ui/SnykSettingsDialog.kt
```diff
@@ -212,8 +212,7 @@ class SnykSettingsDialog(
       manageBinariesAutomatically.isSelected = applicationSettings.manageBinariesAutomatically
       cliPathTextBoxWithFileBrowser.text = applicationSettings.cliPath
       cliBaseDownloadUrlTextField.text = applicationSettings.cliBaseDownloadURL
-      val haveFolderConfigs =
-        LanguageServerWrapper.getInstance(project).getFolderConfigsRefreshed().isNotEmpty()
+      val haveFolderConfigs = service<FolderConfigSettings>().getAll().isNotEmpty()
```
basti-snyk: **Should Fix — `haveFolderConfigs` switched from a project-scoped check to an app-scoped one, so other open projects' folder configs leak into this project's settings dialog.**

This line (and the identical one at `:277` in `initializeFromSettings()`) changed from the project-scoped
`LanguageServerWrapper.getInstance(project).getFolderConfigsRefreshed().isNotEmpty()` to the app-scoped
`service<FolderConfigSettings>().getAll().isNotEmpty()`.

`FolderConfigSettings` is `@Service(Service.Level.APP)`, and `getAll()` returns the folder configs for *every* open project. `haveFolderConfigs` then gates the per-folder fields (`additionalParametersTextField` / `autoDetectOrgCheckbox`, lines ~232/234) via `isEnabled = haveFolderConfigs`.

Failure: with project A (has a stored folder config) and project B (none) both open, opening Settings for B computes `haveFolderConfigs = true` from the app-wide map, so B's per-folder fields are enabled even though B has no configured folder — while their contents come from the project-scoped `getAdditionalParams(project)` / `isOrgSetByUser(project)` and are empty. The APP map is also never cleared per-project, so a stale entry from a closed/reopened project keeps `haveFolderConfigs` true even in a single-project session.

`FolderConfigSettings` already exposes the project-scoped accessor for exactly this — use `service<FolderConfigSettings>().getFolderConfigs(project).isNotEmpty()` (or `getAllForProject(project)`) in both places.

— AI review

nick-y-snyk: Good catch — that's a regression from my flag deletion. The old `getFolderConfigsRefreshed()` was project-scoped (per-wrapper); I replaced its readiness role with app-scoped `getAll()`, which (since `FolderConfigSettings` is `@Service(APP)`) returns every open project's configs.

Fixed with a project-scoped, stored-only predicate: `FolderConfigSettings.hasStoredConfigForProject(project)` — it intersects the project's configured workspace folders with the actual `configs` keys, so it's true only when *this* project has a real stored config (not a transient `createEmpty` default). Used at both readiness sites (215/277).

Note I didn't use `getFolderConfigs(project).isNotEmpty()` — that resolves each folder via `getFolderConfig`, which returns a `createEmpty` default for an unconfigured folder, so it'd be true whenever the project has any workspace folder at all. The store-membership check is the honest signal.

Also fixed the same app-scope leak one block down that wasn't flagged: `baseBranchInfoLabel` used `getAll().values` (so project B's dialog listed project A's base branches) → now `getAllForProject(project)`. Both call sites, both methods.

src/main/kotlin/io/snyk/plugin/ui/jcef/SaveConfigHandler.kt
```diff
@@ -169,6 +172,10 @@ class SaveConfigHandler(
     val isFallback = config.isFallbackForm == true
     if (!isFallback) {
       config.folderConfigs?.let { applyFolderConfigs(it) }
+      // A folder field present as JSON null is a reset. Gson maps it to a Kotlin null on
+      // FolderConfigData (indistinguishable from absent), so applyFolderConfigs() skips it.
+      // Re-parse the raw JSON tree to detect present-with-null folder fields and queue resets.
+      applyFolderResetsFromRawJson(jsonString)
```
basti-snyk, line 178: **Suggestion (efficiency / drift-risk) — the save payload is parsed twice.** `saveConfig` first does `gson.fromJson(jsonString, SaveConfigRequest::class.java)` (line ~160), then `applyFolderResetsFromRawJson(jsonString)` re-parses the identical token-bearing string into a `JsonObject` (line ~591). That's a second full Gson tokenize + tree allocation of the whole payload for zero new information, and it creates two independent parse sites that must stay in agreement on Gson's lenient quirks (number coercion, duplicate keys). Consider parsing once into a `JsonElement`, deserializing `SaveConfigRequest` from that tree, and passing the same tree to the reset detector. — AI review

src/main/kotlin/io/snyk/plugin/services/SnykApplicationSettingsStateService.kt
```diff
@@ -46,54 +47,55 @@ class SnykApplicationSettingsStateService :
...
   fun addPendingReset(key: String) {
     pendingResets.add(key)
   }
 
   fun consumePendingResets(): Set<String> {
-    val snapshot = pendingResets.toSet()
-    pendingResets.clear()
-    return snapshot
+    val drained = mutableSetOf<String>()
+    val it = pendingResets.iterator()
```
basti-snyk: **Suggestion (simplification) — the hand-rolled `iterator().remove()` drain isn't more correct than `toSet()+clear()`.** With `pendingResets` backed by `ConcurrentHashMap.newKeySet()`, this loop has the identical weakly-consistent window as the simpler idiom: a key added after the drain has passed its bucket is also not observed and simply waits for the next `consumePendingResets()`. The extra machinery (and any comment claiming an atomicity gain) overstates the guarantee. Unless a concrete dropped-key race is demonstrated, prefer `val s = pendingResets.toSet(); pendingResets.clear(); return s` — the idiom used elsewhere in this class. — AI review

src/main/kotlin/io/snyk/plugin/ui/jcef/SaveConfigHandler.kt
```diff
@@ -601,165 +680,66 @@ class SaveConfigHandler(
...
       }
 
-      fcs.addFolderConfig(updated)
+      // Persist only when a setting was actually applied: each applyFolderSetting returns a fresh
+      // copy, so updated differs from `existing` by identity iff at least one `?.let` ran. An
+      // all-null reset payload reaches here too (applyFolderConfigs runs before the raw reset
+      // re-parse) and applies nothing, so without this guard it would persist getFolderConfig's
+      // synthetic default for a never-configured folder — the persist-on-read footgun b09b2a16
+      // removed. Re-persisting an unchanged stored config would be a pointless identical write.
+      if (updated !== existing) {
```
basti-snyk: **Suggestion (latent fragility) — persisting on reference identity (`updated !== existing`) is correct only by coincidence.** This skips the write when `withSetting(...)` returns the same instance. It works today purely because `withSetting` always returns a fresh `copy()` (`LanguageServerSettings.kt`). A natural future optimization — returning `this` when the value is unchanged — would silently stop persisting legitimate changes, and no test pins the referential-distinctness contract. The sibling reset path expresses the same intent with an explicit `changed` boolean; consider using a value-based / explicit-flag check here for the same robustness. — AI review

src/main/kotlin/snyk/common/lsp/settings/FolderConfigSettings.kt
```diff
@@ -9,7 +9,8 @@ import com.intellij.openapi.vfs.VirtualFile
 import com.intellij.util.execution.ParametersListUtil
 import io.snyk.plugin.Severity
 import io.snyk.plugin.fromUriToPath
-import io.snyk.plugin.pluginSettings
+import java.io.IOError
+import java.nio.file.InvalidPathException
```
basti-snyk, line 13: **Suggestion (cleanup) — unused import.** `java.nio.file.InvalidPathException` is referenced only in a KDoc `[InvalidPathException]` link, not in code (the catch clause uses `RuntimeException`). It will trip a no-unused-imports lint if one is active; otherwise it's a minor smell. Safe to remove. — AI review

basti-snyk (approved):

src/main/kotlin/io/snyk/plugin/services/SnykApplicationSettingsStateService.kt
nick-y-snyk: Already in place — `loadState` re-wraps `explicitChanges` into `ConcurrentHashMap.newKeySet()` right after `copyBean` (so the concurrent backing survives every IDE start), and `pendingResets` is a `@Transient` field that's never serialized, so `copyBean` never touches it. `pendingResetsLock` is gone entirely; both sets are lock-free `newKeySet()`. No remaining off-lock CME path.

src/main/kotlin/io/snyk/plugin/services/SnykApplicationSettingsStateService.kt
nick-y-snyk: Done — replaced the hand-rolled drain with `val drained = pendingResets.toSet(); pendingResets.clear(); return drained`. You're right that remove-while-iterate gave no atomicity gain over the simpler idiom (identical weakly-consistent window), so I dropped it and the comment that overstated the guarantee.

src/main/kotlin/io/snyk/plugin/ui/jcef/SaveConfigHandler.kt
nick-y-snyk: Done — added `FolderConfigSettings.contains(folderPath)` (direct `configs.containsKey`, O(1), thread-safe on the backing `ConcurrentHashMap`) and use it here instead of `getAll().containsKey(...)`, so the per-folder-element loop no longer copies the whole map for one membership check. The path is already normalized at this call site, matching the stored keys.

src/main/kotlin/io/snyk/plugin/ui/jcef/SaveConfigHandler.kt
nick-y-snyk: Done — replaced the `updated !== existing` identity check with an explicit `applied` boolean set true inside each field's apply (via a small local `apply(key, value)` helper that wraps `applyFolderSetting` + flips the flag). Persist now gates on `if (applied)`, so a future `withSetting` optimization that returns `this` for an unchanged value can't silently stop persisting real changes — same explicit-intent shape as the reset path's `changed` flag.

src/main/kotlin/io/snyk/plugin/ui/jcef/SaveConfigHandler.kt
nick-y-snyk: Won't do — same call as the earlier thread on this (#3491281963). Efficiency-only on a cold path: the payload is a settings form produced by the JCEF dialog (bounded, trusted), parsed once per explicit Save click, so the second parse is microseconds and a few KB with no DoS path. Threading one `JsonObject` through `saveConfig` -> `applyFolderConfigs` / `applyFolderResetsFromRawJson` would change two private signatures for no measurable gain. Happy to revisit if the payload ever grows or this becomes hot.

src/main/kotlin/snyk/common/lsp/settings/FolderConfigSettings.kt
nick-y-snyk: Won't do — it's not actually unused: the KDoc on `normalizePathOrNull` references it via a `[InvalidPathException]` link, which Dokka/IDE resolve against the import. Removing it would break the doc link (and the IDE would re-flag it as unresolved). The catch deliberately widens to `RuntimeException` to also cover `SecurityException`, but the doc still names `InvalidPathException` as the primary case, so the reference is intentional.

PR #847, opened 2026-06-19
feat: clear global overrides on Project Defaults reset [IDE-2149]

nick-y-snyk:
## What

Implements the JetBrains side of **IDE-2149** — global ("Project Defaults") reset overrides in the HTML settings page (sibling of the per-folder reset, IDE-1945).

When the HTML settings page sends a top-level global field as JSON `null`, IntelliJ treats it as a Project Defaults reset:
- clears the persisted value (restores the plugin default),
- clears explicit-change tracking, and
- queues a pending reset so `getSettings()` emits `{value:null, changed:true}` **once**,

so the override is not re-pushed on the next sync/reconnect.

## Changes
- `SaveConfigHandler` — `applyGlobalResetsFromRawJson` detects top-level null fields among the org-scope resettable keys (`snyk_*_enabled`, `scan_automatic`, `scan_net_new`, `severity_filter_*`, `issue_view_*`, `risk_score_threshold`, `organization`); per key clears explicit-changed, restores the default, and `addPendingReset`. Wired into `saveConfig`.
- Reuses existing global `pendingResets`/`consumePendingResets` and the `getSettings()` emit loop.
- Tests — `SaveConfigHandlerTest` covers org/product/filter/risk-score resets and present/absent/fallback no-reset cases; **re-push guard** in `LanguageServerWrapperTest` (emit once, then `changed:false` on reconnect).

## Notes
- This branch is based on `master` (clean); the per-folder reset (IDE-1945) is a separate stream and is **not** required here — the global path is self-contained.
- Requires the **snyk-ls** IDE-2149 PR for end-to-end behavior; safe to merge independently (null reset is a no-op on the LS side until then).


src/main/kotlin/io/snyk/plugin/ui/jcef/SaveConfigHandler.kt
```diff
@@ -486,6 +491,117 @@ class SaveConfigHandler(
...
+      }
+    if (!root.isJsonObject) return
+    val obj = root.asJsonObject
+
+    for ((fieldNames, key, restoreDefault) in globalResetSpecs(settings)) {
+      val isReset = fieldNames.any { obj.has(it) && obj.get(it).isJsonNull }
+      if (isReset) {
+        settings.clearExplicitlyChanged(key)
+        restoreDefault()
+        settings.addPendingReset(key)
```
basti-snyk, line 569: **Should Fix — the one-shot reset signal this PR queues here has two delivery risks worth verifying end-to-end:**

1. **Multi-project starvation.** `pendingResets` lives on the app-scoped (`Service.Level.APP`) settings singleton, but `saveConfig` loops over *all* open projects, each calling `getSettings()` → `consumePendingResets()`, which snapshots-and-clears the shared set. With 2+ projects open, only the *first* project's language server receives `{value:null, changed:true}`; every other project's LS gets an empty reset set and emits `{value:<default>, changed:false}`, so those LSs keep the stale `user:global` override. The new `LanguageServerWrapperTest` exercises a single wrapper, so it can't catch this. Consider per-project pending-reset tracking (or re-emit to each wrapper before draining).

2. **Null value on the wire.** `ConfigSetting(value=null, changed=true)` is serialized through lsp4j's default Gson (no `serializeNulls`), which omits null fields nested in a non-null params object — so the LS may receive `{changed:true}` with no `value` key at all. The unit tests assert at the Kotlin-object level (pre-serialization), so they'd pass even if the wire signal is degraded. Worth confirming the actual JSON-RPC payload contains `value:null`, and that snyk-ls treats it as "unset override".

(Also, thread-safety: `pendingResets` is a plain `mutableSetOf()` with `add` from the JCEF save thread and `consume` from the getSettings thread — a concurrent set or shared lock would remove the lost-reset / CME risk.)

— AI review

src/main/kotlin/io/snyk/plugin/ui/jcef/SaveConfigHandler.kt
```diff
@@ -486,6 +491,117 @@ class SaveConfigHandler(
...
+   * Org-scope global settings the user can reset to "Project Defaults" from the LS HTML dialog.
+   * Each entry maps the JSON field name(s) the HTML/JS may emit to the canonical LS setting key and
+   * the action that restores the persisted plugin default so the value is not re-pushed after the
+   * one-shot reset signal is consumed.
+   *
+   * Detection of a reset is by explicit JSON null at the top level (a present field with a non-null
+   * value is a normal change, an absent field is "no change"). Gson collapses absent and null into
+   * Kotlin null on [SaveConfigRequest], so reset detection must read the raw JSON.
+   */
+  private fun globalResetSpecs(
```
basti-snyk: **Suggestion — reset correctness hinges on a cross-repo field-name contract that nothing here pins or tests end-to-end.** `globalResetSpecs` keys reset detection on JSON field names (e.g. `snyk_oss_enabled`/`activateSnykOpenSource`) that must match exactly what the snyk-ls–served reset dialog emits. The only HTML in this repo (`settings-fallback.html`) emits none of these fields, so if any name/casing diverges, the reset silently no-ops with no user feedback. Note three keys (`scan_automatic`, `issue_view_open_issues`, `issue_view_ignored_issues`) have snake_case only while peers carry a camelCase alias too — the likeliest to silently miss. Worth pinning the contract (shared constant/fixture from the snyk-ls schema) and adding one test that drives the real reset payload.

Two maintainability notes on this table: the restore-default literals duplicate the `PLUGIN_DEFAULT_*` constants used by the deviation checks in `SnykApplicationSettingsStateService` — if a default changes in one place, reset idempotency silently breaks; referencing the same constant (or a `restoreToDefault(key)` helper) gives one source of truth. And modelling each spec as a `data class GlobalResetSpec` rather than a positional `Triple` would make the contract self-documenting. — AI review

src/main/kotlin/io/snyk/plugin/ui/jcef/SaveConfigHandler.kt
```diff
@@ -507,6 +512,117 @@ class SaveConfigHandler(
...
+    settings: SnykApplicationSettingsStateService
+  ): List<Triple<List<String>, String, () -> Unit>> =
+    listOf(
+      Triple(
+        listOf("snyk_oss_enabled", "activateSnykOpenSource"),
+        LsFolderSettingsKeys.SNYK_OSS_ENABLED,
+      ) {
+        settings.ossScanEnable = true
```
basti-snyk: **Should Fix — `restoreDefault` lambdas hard-code a third copy of the plugin defaults, which can silently break the one-shot reset.**

Each `restoreDefault()` here re-types the default literal (`settings.ossScanEnable = true`, `settings.secretsEnabled = false`, `settings.riskScoreThreshold = null`, etc.). But the *deviation check* that decides whether a value gets re-pushed as `changed:true` reads from the canonical `PLUGIN_DEFAULT_*` constants in `SnykApplicationSettingsStateService` (e.g. `PLUGIN_DEFAULT_SECRETS_SCAN_ENABLE`). These two sources are now independent copies.

If a product decision ever flips a default (say `PLUGIN_DEFAULT_SECRETS_SCAN_ENABLE` → `true`), the deviation comparator tracks the new default but this lambda still writes the stale literal. After a reset the restored value would then *deviate* from the real default, the deviation check returns `true`, and `getSettings` re-asserts the override on the next reconnect — silently defeating the reset, with no compile error to catch the drift. Sourcing each restored value from the `PLUGIN_DEFAULT_*` constant (or a shared `resetGlobalKeyToDefault(key)` helper on the settings service) would keep the restore path and the deviation baseline in lockstep.

— AI review

src/main/kotlin/io/snyk/plugin/ui/jcef/SaveConfigHandler.kt
nick-y-snyk: Fixed in commit 5ce81b73.

**Multi-project starvation (pt 1):** drained  once in  before the project fan-out loop and passed the same snapshot explicitly to each  call.  and  now accept  as a parameter (default still drains inline for all other callers). All open projects' LSs now receive the same  reset set.

**Null value on wire (pt 2) — false positive:** verified against snyk-ls.  is `json:"value"` with no `omitempty` (`internal/types/lsp.go:558`), so a missing `value` key in JSON unmarshals to Go `nil`. The reset detector is `cs.Changed && cs.Value == nil` (`folder_config.go:460`) and the wire contract explicitly documents "Changed: true + Value: nil = reset to default" (`lsp.go:556`). Gson dropping null on our side + Go defaulting absent→nil converge identically. No change needed.

**Thread-safety (pt 3):** already fixed in a prior commit — `pendingResets` is `ConcurrentHashMap.newKeySet()`.

src/main/kotlin/io/snyk/plugin/ui/jcef/SaveConfigHandler.kt
```diff
@@ -180,9 +185,18 @@ class SaveConfigHandler(
...
+    // Only drain if there is at least one live project to send to — if none are ready the
+    // pending resets remain queued and will be picked up on the next LS reconnect/init.
+    val liveProjects =
+      ProjectUtil.getOpenProjects().filter {
+        !it.isDisposed && !SnykPluginDisposable.getInstance(it).isDisposed()
+      }
+    if (liveProjects.isNotEmpty()) {
+      val resets = pluginSettings().consumePendingResets()
+      for (openProject in liveProjects) {
+        LanguageServerWrapper.getInstance(openProject).updateConfiguration(resets = resets)
```
basti-snyk, line 199: why send it per project?

basti-snyk (approved):

PR #856, opened 2026-07-03
feat: remove native folder trust dialog, delegate to HTML tree view [IDE-2166]

nick-y-snyk:
### Description

Removes the native JetBrains trust dialog and delegates trust entirely to the LS-served HTML tree view, matching the VS Code reference implementation (IDE-1882, Done).

**Root cause of double-prompt:** `trust_enabled` was hardcoded to `false`, so the LS never enforced trust regardless of `trusted_folders`. Native dialog was the only gate.

**Changes:**
- `trust_enabled` set to `true` — LS now enforces trust and renders the untrusted-folder CTA in the HTML tree
- All workspace folders sent to LS unconditionally (no IDE-side trust filter)
- Deleted `TrustedProjects.kt` (native `MessageDialogBuilder.yesNo` modal)
- Removed `isPathTrusted` / `isAncestor` from `WorkspaceTrustService` (no callers remain)
- Removed `confirmScanningAndSetWorkspaceTrustedStateIfNeeded` gate from `SnykTaskQueueService.scan()`
- Removed auto-trust from `SnykAuthPanel` on first-run auth (matches VS Code)
- Removed native trust bundle keys
- Cleaned up `promptForTrust` param across all callers and test stubs

Jira: https://snyksec.atlassian.net/browse/IDE-2166

### Checklist

- [x] Linted
- [ ] CHANGELOG.md updated

### Screenshots / GIFs

_Before:_ Native "Trust and Scan Project 'X'?" modal appears on project open.
_After:_ No native modal; untrusted folder CTA rendered in HTML tree view by the LS.


src/main/kotlin/snyk/common/lsp/LanguageServerWrapper.kt
```diff
@@ -673,7 +648,7 @@ class LanguageServerWrapper(private val project: Project) : Disposable {
         changed = ps.lsUserAssertedChangeForLsConfigurationKey(LsSettingsKeys.AUTHENTICATION_METHOD),
       )
 
-    settingsMap[LsSettingsKeys.TRUST_ENABLED] = ConfigSetting(value = false, changed = true)
+    settingsMap[LsSettingsKeys.TRUST_ENABLED] = ConfigSetting(value = true, changed = true)
```
basti-snyk, line 652: **Should-fix (test coverage):** This flip of `TRUST_ENABLED` `false`→`true` is the load-bearing line of this PR — with the native trust dialog removed, it is now the *only* thing keeping folder-trust enforcement alive (delegated to the Language Server). It has no test coverage: `rg TRUST_ENABLED` finds only the definition and this assignment, and the existing `getInitializationOptions` test asserts sibling settings but not this one. A silent revert to `false` would leave neither native nor LS-side trust and would auto-scan/execute package managers on untrusted code — a security regression that passes CI. Please add a one-line assertion to the existing initialization-options test: `assertEquals(true, actual.settings?.get("trust_enabled")?.value)`.

— AI review

src/main/kotlin/snyk/common/lsp/LanguageServerWrapper.kt
```diff
@@ -514,7 +489,7 @@ class LanguageServerWrapper(private val project: Project) : Disposable {
   fun sendScanCommand() {
     if (notAuthenticated()) return
     DumbService.getInstance(project).runWhenSmart {
-      getTrustedContentRoots(project, promptForTrust = true).forEach {
+      getContentRoots(project).forEach {
```
basti-snyk: **Suggestions (non-blocking):** Two low-severity notes on this line, both introduced/exposed by the `getTrustedContentRoots` → `getContentRoots` rename:

1. `getContentRoots` no longer calls `root.toNioPath()`, so the old `UnsupportedOperationException` guard that skipped content roots on non-local/in-memory filesystems is gone. Verified benign for real projects (the throw site was the deleted trust check), but such a root would now be forwarded to the LS as a scannable folder — a minor behavior change worth a glance.
2. `addContentRoots(project)` is invoked once per content root inside this `forEach`, though it is project-scoped and idempotent. Consider hoisting it above the loop so only `sendFolderScanCommand` runs per-root.

— AI review

src/main/kotlin/snyk/common/lsp/LanguageServerWrapper.kt
nick-y-snyk: Done — added `assertEquals(true, actual.settings?.get("trust_enabled")?.value)` to the existing `getInitializationOptions` test.

src/main/kotlin/snyk/common/lsp/LanguageServerWrapper.kt
nick-y-snyk: Both addressed: hoisted `addContentRoots(project)` above the loop, and added `if (!root.isInLocalFileSystem) continue` guard in `getContentRoots` to restore the non-local filesystem filter.

src/main/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanel.kt
```diff
@@ -671,9 +229,15 @@ class SnykToolWindowPanel(val project: Project) : JPanel(), Disposable {
     }
   }
 
+  private fun noIssuesInAnyProductFound(): Boolean {
```
basti-snyk, line 258: This is missing secrets

basti-snyk: pre-existing, but worth fixing.

src/main/kotlin/snyk/common/lsp/LanguageServerWrapper.kt
```diff
@@ -346,41 +344,18 @@ class LanguageServerWrapper(private val project: Project) : Disposable {
...
   private fun processIsAlive() = ::process.isInitialized && process.isAlive
 
-  fun getWorkspaceFoldersFromRoots(
-    project: Project,
-    promptForTrust: Boolean = true,
-  ): Set<WorkspaceFolder> {
+  fun getWorkspaceFoldersFromRoots(project: Project): Set<WorkspaceFolder> {
     if (disposed || project.isDisposed) return emptySet()
-    val normalizedRoots = getTrustedContentRoots(project, promptForTrust)
+    val normalizedRoots = getContentRoots(project)
```
basti-snyk, line 351: This means the contract for `getWorkspaceFoldersFromRoots` changes: it does not return only trusted folders, it now returns all folders. Have you checked all callers, that this is not introducing issues?

src/main/kotlin/io/snyk/plugin/ui/toolwindow/panels/SnykAuthPanel.kt
```diff
@@ -37,26 +37,23 @@ class SnykAuthPanel(val project: Project) : JPanel(), Disposable {
...
                 jButton.setText("Authenticating...")
                 getSnykCliAuthenticationService(project)?.authenticate() ?: ""
 
-                // explicitly add the project to workspace trusted paths, because
-                // scan can be auto-triggered depending on "settings.pluginFirstRun" value
                 jButton.setText("Trusting project paths...")
                 val trustService = service<WorkspaceTrustService>()
-                val paths = project.getContentRootPaths()
-                for (path in paths) {
+                for (path in project.getContentRootPaths()) {
```
basti-snyk: **Should-fix (trust consent) — corrected after verifying snyk-ls.** Correction to my earlier framing on this thread: snyk-ls is **fail-closed and never scans an untrusted folder** (`domain/ide/workspace/folder.go:349` early-returns on `!IsTrusted()`; `IsTrusted` at `:974` requires the folder to be in `TRUSTED_FOLDERS` when `TRUST_ENABLED=true`, and every scan entrypoint funnels through that one gate). So this is **not** "untrusted code gets scanned" — I was wrong about that mechanism.

The accurate issue: this auth handler adds *every* content root to `WorkspaceTrustService` on authenticate, which is forwarded as `TRUSTED_FOLDERS` (`LanguageServerWrapper.kt:683`). So the project open **at authenticate time** is auto-trusted and scanned **without ever hitting the LS consent surface** (the untrusted-folder tree node + `$/snyk.addTrustedFolders` flow) that this PR otherwise relies on. Note the auto-trust loop itself is **pre-existing on master** — what this PR changes around it is what matters:

1. The button was relabeled `"Trust project and scan"` → `"Authenticate"` and the explaining comment removed — so what master *disclosed* as a trust+scan action is now an auth-only label that silently trusts+scans the open repo.
2. The IDE-side modal (`confirmScanningAndSetWorkspaceTrustedStateIfNeeded`, `TrustedProjects.kt`) that used to prompt for this is deleted.

Scope: for the "already-authenticated user opens a new untrusted project" path the posture is **preserved** (LS skips the scan and surfaces the untrusted-folder consent node). The gap is only this auth-time path. If the intent is for snyk-ls to own trust, **drop the IDE-side auto-trust here and let the LS surface consent uniformly**; otherwise restore a trust-labelled/disclosed affordance on the button. (Test ask unchanged: the auth→trust path has no test — drive the action and assert `addTrustedPath` per content root.)

(Minor, unchanged: this loop trusts every content root without the `isInLocalFileSystem` filter `getContentRoots` applies, so non-local/temp roots get persisted to the store even though they are never sent to the LS.)

— AI review

src/main/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanel.kt
```diff
@@ -671,9 +229,15 @@ class SnykToolWindowPanel(val project: Project) : JPanel(), Disposable {
...
+    val cache = getSnykCachedResults(project) ?: return true
+    return cache.currentOSSResultsLS.isEmpty() &&
```
basti-snyk: **Should-fix (correctness) — new at HEAD `830909d`:** `noIssuesInAnyProductFound()` checks map *emptiness*, but the LS cache stores an empty set per product when a scan finishes clean. `SnykLanguageClient.updateCache` (`SnykLanguageClient.kt:138-162`) explicitly does `currentOSSResultsLS[snykFile] = emptySet()` with the comment *"The language server sends empty diagnostics per product when it has finished a scan and did not find anything"*. So after a clean scan the maps are **non-empty** (keys → empty sets) and `.isEmpty()` returns `false`. `displayEmptyDescription()` then routes to `displaySelectVulnerabilityMessage()` ("Select an issue and start improving your project") instead of the "no vulnerabilities found" message. The removed logic counted actual issue nodes (`childCount == 0`) and was correct.

Fix: check that the stored sets are all empty (e.g. `all { it.isEmpty() }` / flatten-and-count) rather than map key presence.

— AI review

src/main/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanel.kt
```diff
@@ -649,13 +211,9 @@ class SnykToolWindowPanel(val project: Project) : JPanel(), Disposable {
...
     treeSplitter.firstComponent = summaryPanel
 
-    if (isHtmlTreeViewEnabled() && JBCefApp.isSupported()) {
-      htmlTreePanel = HtmlTreePanel(project)
-      Disposer.register(this, htmlTreePanel!!)
-      treeSplitter.secondComponent = htmlTreePanel
-    } else {
-      treeSplitter.secondComponent = TreePanel(vulnerabilitiesTree, project, this)
-    }
+    htmlTreePanel = HtmlTreePanel(project)
```
basti-snyk, line 240: **Suggestion (non-blocking) — new at HEAD `830909d`:** `HtmlTreePanel` is now instantiated unconditionally. The old branch guarded on `JBCefApp.isSupported()` and fell back to the native `TreePanel` when JCEF was unavailable. With the native tree removed, in JCEF-unsupported environments (headless / some remote-dev / certain JBR configs) `HtmlTreePanel` degrades to an error balloon ("Failed to show HTML tree view…") and renders no results view at all — the user has no way to see findings. It does not crash, and HTML-only is the stated intent of this PR, so this is a product/UX confirmation rather than a defect: please confirm losing the results view in JCEF-unsupported environments is acceptable.

— AI review

src/main/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanel.kt
```diff
@@ -440,117 +144,11 @@ class SnykToolWindowPanel(val project: Project) : JPanel(), Disposable {
...
-              scanListenerLS.displayIacResults(results)
-            }
-          }
-          LsProduct.Secrets -> Unit
-          LsProduct.Unknown -> Unit
-        }
-      }
-    }
-  }
+  // ponytail: JCEF self-renders; kept as no-op for SnykApplicationActivationListener callers
```
basti-snyk: **Suggestion (non-blocking) — new at HEAD `830909d`:** The `ponytail:` prefix in this comment is an agent/harness codename — it appears nowhere else in the repo and is unrelated to IDE-2166. Please strip the `ponytail:` marker and keep only the factual rationale (e.g. "JCEF self-renders; kept as a no-op for `SnykApplicationActivationListener` callers").

— AI review

src/main/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanel.kt
```diff
@@ -440,117 +144,11 @@ class SnykToolWindowPanel(val project: Project) : JPanel(), Disposable {
...
+  fun refreshUI() = Unit
```
basti-snyk: **Suggestion (non-blocking) — new at HEAD `830909d`:** `refreshUI()` is now a no-op with a single caller (`SnykApplicationActivationListener`) that still performs a panel lookup and logs "Calling refreshUI / refreshUI completed" on every app activation just to invoke a method that does nothing. Prefer deleting the method and its call site over leaving vestigial scaffolding.

— AI review

src/test/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanelIntegTest.kt
basti-snyk, line 1: Deleting all tests here is wrong. There's behaviour and functionality that should still apply and be tested, regardless of HTML tree view (e.g. showing of different panels on different status.

src/test/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanelTest.kt
basti-snyk, line 1: Same here

src/main/kotlin/snyk/common/lsp/LanguageServerWrapper.kt
basti-snyk: Verified against snyk-ls (`domain/ide/workspace/folder.go`, branch `feat/IDE-2052-remediation-followups`) — the contract change is **safe across all 7 callers**, because snyk-ls re-gates every scan on trust independently:

- `folder.scan()` (`folder.go:346`) is the only path to `scanner.Scan`; it early-returns on `!IsTrusted()` — *"skipping scan of untrusted path"*.
- `IsTrusted()` (`folder.go:974`) is fail-closed under `TRUST_ENABLED=true`: trusted only if the folder is contained in `TRUSTED_FOLDERS`.
- Every entrypoint funnels through it: `ScanWorkspace` iterates only `GetFolderTrust()`'s trusted set; the explicit `workspaceFolderScanCommand` → `ScanFolder` → `scan`.

So returning all local roots just means the LS registers them and scans the trusted subset. The 5 settings/config callers additionally intersect with `configuredWorkspaceFolders`; the project-close caller is removal-only (bounded by `updateWorkspaceFolders`'s `removed` filter). No caller introduces an untrusted scan — the contract change does not change the security posture. The real consent question is separate and auth-time-only; see the `SnykAuthPanel.kt` thread.

— AI review

src/main/kotlin/snyk/common/lsp/LanguageServerWrapper.kt
nick-y-snyk: Both addressed in commit f7481989: hoisted `addContentRoots(project)` above the loop and restored the `isInLocalFileSystem` guard in `getContentRoots`.

src/main/kotlin/io/snyk/plugin/ui/toolwindow/panels/SnykAuthPanel.kt
nick-y-snyk: Restored trust disclosure: button label changed to 'Trust project and authenticate' so the auto-trust action is explicit. Also added isInLocalFileSystem filter to the auto-trust loop, mirroring the same guard in getContentRoots.

src/main/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanel.kt
nick-y-snyk: Fixed: added currentSecretsResultsLS to the check, and replaced map.isEmpty() with map.isEmpty() || map.values.all { it.isEmpty() } so clean LS scans (keys→emptySet) correctly show the 'no vulnerabilities' state.

src/main/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanel.kt
nick-y-snyk: Confirmed JCEF-unsupported is acceptable. Improved the error message to be specific: 'Snyk results panel requires JCEF (Chromium Embedded Framework), which is not available in this environment. Please use a JetBrains Runtime (JBR) build of your IDE.'

src/main/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanel.kt
nick-y-snyk: Stripped the ponytail: prefix — kept only the factual rationale.

src/main/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanel.kt
nick-y-snyk: Deleted refreshUI() and its SnykApplicationActivationListener call site — it was a no-op with logging overhead on every app activation.

src/test/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanelIntegTest.kt
nick-y-snyk: Restored panel-state tests for auth/scan/empty states in a new SnykToolWindowPanelTest.kt (LightPlatform4TestCase). The IntegTest file was already @Ignore'd in CI — not restored. Native-tree-layout tests are legitimately gone with the tree.

src/main/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanel.kt
nick-y-snyk: Fixed — added currentSecretsResultsLS to the check alongside the map-emptiness fix.

src/test/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanelTest.kt
nick-y-snyk: Same fix applied — restored panel-state tests (auth/scan/empty) independent of tree view.

src/main/kotlin/snyk/common/lsp/LanguageServerWrapper.kt
```diff
@@ -346,41 +344,18 @@ class LanguageServerWrapper(private val project: Project) : Disposable {
...
-
-      if (!pathTrusted) {
-        logger.debug("Path not trusted: ${root.path}")
-        continue
-      }
-
+      if (!root.isInLocalFileSystem) continue
       var add = true
       for (normalizedRoot in normalizedRoots) {
         if (!root.path.startsWith(normalizedRoot.path)) continue
```
basti-snyk: **Should-fix (correctness) — re-review at HEAD `ad687db`:** This de-dup uses a raw `String.startsWith` with no path-separator boundary, so sibling content roots that share a name prefix collide. With roots `/proj/api` and `/proj/api-tests`, `"/proj/api-tests".startsWith("/proj/api")` is `true`, so `api-tests` is treated as nested under `api`, `add` is set `false`, and that root is silently never registered or scanned.

This line is unchanged from `master`, but this PR promotes it: `getTrustedContentRoots(...)` → `getContentRoots(...)` removed the trust pre-filter that used to sit in front of this loop, so this de-dup is now the *sole* gate deciding which roots reach the language server. A dropped sibling root means a whole module goes unscanned with no error.

Fix: compare on a path boundary rather than a raw prefix, e.g.
```kotlin
if (root.path != normalizedRoot.path && !root.path.startsWith(normalizedRoot.path + "/")) continue
```
or use `VfsUtilCore.isAncestor(normalizedRoot, root, false)`.

— AI review

nick-y-snyk: Fixed: replaced raw startsWith with VfsUtilCore.isAncestor(normalizedRoot, root, false) — correctly handles sibling roots like /proj/api vs /proj/api-tests.

src/main/kotlin/io/snyk/plugin/ui/toolwindow/panels/HtmlTreePanel.kt
```diff
@@ -84,10 +85,19 @@ class HtmlTreePanel(project: Project) : JPanel(), Disposable {
             }
           },
         )
+      // Request the current tree in case a scan completed before this panel was opened.
+      runAsync {
+        try {
+          LanguageServerWrapper.getInstance(project)
+            .executeCommandWithArgs("snyk.getTreeView", emptyList())
```
basti-snyk: I may remember it wrong, but isn't the actual tree view HTML returned by the command? Also, I'd use a constant for the command name. Could you please write a test that verifies the correct behavior?

basti-snyk (approved):
Approving, but please check the snyk.getTreeView command semantics and whether you need to trigger updating the browser component for the tree view after retrieving the HTML.

src/main/kotlin/io/snyk/plugin/ui/toolwindow/panels/HtmlTreePanel.kt
nick-y-snyk: Fixed in `3188e9de`:
- `snyk.getTreeView` returns the HTML string directly (confirmed in `snyk-ls/domain/ide/command/get_tree_view.go:67`). The return value was being discarded — cast to `String?` and loaded into the browser (same dedup guard as the push path).
- Added `COMMAND_GET_TREE_VIEW` constant in `Commands.kt`.
- Two tests in `HtmlTreePanelTest`: one verifying the returned HTML is loaded into the browser on init, one verifying no extra `loadHTML` when the command returns null.

PR #858, opened 2026-07-08
feat(auth): add "Snyk: Log out" action [IDE-2181]

rrama:
### Description

Adds a first-class **Snyk: Log out** action so users can clear a stale/expired OAuth token without needing the HTML settings page to load first.

Context ([IDE-2181](https://snyksec.atlassian.net/browse/IDE-2181), caused by [IDE-2178](https://snyksec.atlassian.net/browse/IDE-2178)): when the stored token is expired (`invalid_grant`), the language server thrashes on a doomed refresh and starves the `workspace/configuration` command, so the settings page — the only existing re-auth entry point — times out to the fallback. This action breaks that chicken-and-egg.

Changes:
- New `SnykCliAuthenticationService.logout()`, symmetric with `authenticate()`: clears the local token **first** (so a `workspace/configuration` pull during logout can't hand the stale token back and re-auth the LS), then clears it in the LS via `COMMAND_LOGOUT`, then resets the tool window to the auth panel.
- New `SnykLogoutAction` delegates to it. Registered standalone → surfaces in **Find Action** ("Snyk: Log out"), not the tool window toolbar.
- Renamed the existing settings action to **"Snyk: Settings"**.

### Checklist

- [x] Linted
- [ ] CHANGELOG.md updated
 - N/A
 - N/A

### Screenshots / GIFs

_Toolbar unchanged (settings gear retained); logout is available via Find Action (Cmd+Shift+A -> "Snyk: Log out")._

[IDE-2181]: https://snyksec.atlassian.net/browse/IDE-2181?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[IDE-2178]: https://snyksec.atlassian.net/browse/IDE-2178?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/kotlin/io/snyk/plugin/services/SnykCliAuthenticationService.kt
+++ b/src/main/kotlin/io/snyk/plugin/services/SnykCliAuthenticationService.kt
@@ -13,9 +13,12 @@ import com.intellij.openapi.ui.DialogWrapper
 import com.intellij.ui.components.JBScrollPane
 import com.intellij.ui.scale.JBUIScale
 import com.intellij.util.PlatformIcons
+import io.snyk.plugin.events.SnykSettingsListener
 import io.snyk.plugin.getCliFile
 import io.snyk.plugin.getSnykTaskQueueService
+import io.snyk.plugin.getSnykToolWindowPanel
 import io.snyk.plugin.pluginSettings
+import io.snyk.plugin.publishAsync
 import io.snyk.plugin.ui.SnykBalloonNotificationHelper
 import io.snyk.plugin.ui.getReadOnlyClickableHtmlJEditorPane
 import java.awt.BorderLayout
@@ -34,6 +37,30 @@ import snyk.common.lsp.LanguageServerWrapper
 @Service(Service.Level.PROJECT)
 class SnykCliAuthenticationService(val project: Project) {
   private val logger = logger<SnykCliAuthenticationService>()
+
+  /**
+   * Log out of Snyk: clears the stored token both in the Language Server and locally, then returns
+   * the tool window to the authentication panel. Runs on a pooled thread so callers (actions, UI)
+   * do not block. Clearing the local token is what stops it being re-sent to the LS on the next
+   * workspace/didChangeConfiguration push.
+   */
+  fun logout() {
+    ApplicationManager.getApplication().executeOnPooledThread {
+      if (project.isDisposed) return@executeOnPooledThread
+      // Clear the local token before logging the LS out. If the LS pulls workspace/configuration
+      // while handling logout, getSettings() would otherwise hand the stale token straight back
+      // and re-authenticate the server we just logged out.
+      pluginSettings().token = ""
+      LanguageServerWrapper.getInstance(project).logout()
+
+      ApplicationManager.getApplication().invokeLater {
+        if (project.isDisposed) return@invokeLater
+        getSnykToolWindowPanel(project)?.cleanUiAndCaches()
+      }
+      // Re-render the tool window; with an empty token it falls back to the authentication panel.
+      publishAsync(project, SnykSettingsListener.SNYK_SETTINGS_TOPIC) { settingsChanged() }
+    }
+  }
 
   fun authenticate() {
     fun showAuthDialog() {
--- a/src/main/kotlin/io/snyk/plugin/services/SnykCliAuthenticationService.kt
+++ b/src/main/kotlin/io/snyk/plugin/services/SnykCliAuthenticationService.kt
@@ -50,15 +50,20 @@ class SnykCliAuthenticationService(val project: Project) {
       // Clear the local token before logging the LS out. If the LS pulls workspace/configuration
       // while handling logout, getSettings() would otherwise hand the stale token straight back
       // and re-authenticate the server we just logged out.
-      pluginSettings().token = ""
-      LanguageServerWrapper.getInstance(project).logout()
-
-      ApplicationManager.getApplication().invokeLater {
-        if (project.isDisposed) return@invokeLater
-        getSnykToolWindowPanel(project)?.cleanUiAndCaches()
+      pluginSettings().token = null
+      try {
+        LanguageServerWrapper.getInstance(project).logout()
+      } finally {
+        // Always clean up locally, even if the LS logout call throws something other than the
+        // TimeoutException it already handles internally. Otherwise the previous user's cached
+        // scan findings would remain on screen after the token has been cleared.
+        ApplicationManager.getApplication().invokeLater {
+          if (project.isDisposed) return@invokeLater
+          getSnykToolWindowPanel(project)?.cleanUiAndCaches()
+        }
+        // Re-render the tool window; with an empty token it falls back to the authentication panel.
+        publishAsync(project, SnykSettingsListener.SNYK_SETTINGS_TOPIC) { settingsChanged() }
       }
-      // Re-render the tool window; with an empty token it falls back to the authentication panel.
-      publishAsync(project, SnykSettingsListener.SNYK_SETTINGS_TOPIC) { settingsChanged() }
     }
   }
 
--- a/src/test/kotlin/io/snyk/plugin/services/SnykCliAuthenticationServiceTest.kt
+++ b/src/test/kotlin/io/snyk/plugin/services/SnykCliAuthenticationServiceTest.kt
@@ -116,12 +116,10 @@ class SnykCliAuthenticationServiceTest : LightPlatform4TestCase() {
 
     SnykCliAuthenticationService(project).logout()
 
-    // logout() runs on a pooled thread; verify with a timeout, then poll for the token clear.
+    // logout() runs on a pooled thread; verify with a timeout. The local token clear happens on
+    // the same thread strictly before the LS logout() call, so once MockK observes that call the
+    // token is guaranteed to already be cleared - no need to sleep-poll for it separately.
     verify(timeout = 2000) { languageServerWrapperMock.logout() }
-    val deadline = System.currentTimeMillis() + 2000
-    while (System.currentTimeMillis() < deadline && !pluginSettings().token.isNullOrEmpty()) {
-      Thread.sleep(20)
-    }
     assertTrue(pluginSettings().token.isNullOrEmpty())
   }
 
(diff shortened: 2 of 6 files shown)


src/main/kotlin/io/snyk/plugin/services/SnykCliAuthenticationService.kt
```diff
@@ -35,6 +38,30 @@ import snyk.common.lsp.LanguageServerWrapper
...
+   * workspace/didChangeConfiguration push.
+   */
+  fun logout() {
+    ApplicationManager.getApplication().executeOnPooledThread {
+      if (project.isDisposed) return@executeOnPooledThread
+      // Clear the local token before logging the LS out. If the LS pulls workspace/configuration
+      // while handling logout, getSettings() would otherwise hand the stale token straight back
+      // and re-authenticate the server we just logged out.
+      pluginSettings().token = ""
+      LanguageServerWrapper.getInstance(project).logout()
```
basti-snyk: **Should-Fix: a failing LS logout skips local UI/cache cleanup (partial logout).**

`LanguageServerWrapper.logout()` only catches `TimeoutException`; any other throwable propagates out of `logout()` before the `invokeLater { cleanUiAndCaches() }` and `publishAsync { settingsChanged() }` below run. Result: the token is cleared but the previous user's cached scan findings stay on screen in the tool window — a partial, security-relevant logout.

Run the local cleanup regardless of LS outcome — wrap the LS call so cleanup executes in a `finally`, or scrub caches/UI before calling the LS. — AI review

src/test/kotlin/io/snyk/plugin/services/SnykCliAuthenticationServiceTest.kt
```diff
@@ -107,6 +109,22 @@ class SnykCliAuthenticationServiceTest : LightPlatform4TestCase() {
...
+    every { languageServerWrapperMock.logout() } returns Unit
+    pluginSettings().token = "a-stale-token"
+
+    SnykCliAuthenticationService(project).logout()
+
+    // logout() runs on a pooled thread; verify with a timeout, then poll for the token clear.
+    verify(timeout = 2000) { languageServerWrapperMock.logout() }
+    val deadline = System.currentTimeMillis() + 2000
+    while (System.currentTimeMillis() < deadline && !pluginSettings().token.isNullOrEmpty()) {
+      Thread.sleep(20)
```
basti-snyk: **Should-Fix: sleep-poll against a pooled thread can flake under CI load.**

The `verify(timeout = 2000)` plus a fixed 2000ms `Thread.sleep(20)` poll loop is a wall-clock race — on a slow CI box the budget can lapse before the pooled thread finishes, producing intermittent red. This repo is flake-sensitive. Prefer deterministic synchronization (e.g. a `CountDownLatch` the production path signals, as `UtilsIntegTest` already does, or draining the pooled executor) over a timeout poll. Same pattern in `SnykLogoutActionTest`. — AI review

src/test/kotlin/io/snyk/plugin/ui/actions/SnykLogoutActionTest.kt
```diff
@@ -0,0 +1,97 @@
...
+    pluginSettings().token = "a-stale-token"
+
+    action.actionPerformed(actionEvent())
+
+    // actionPerformed runs on a pooled thread; verify with a timeout.
+    verify(timeout = 2000) { languageServerWrapperMock.logout() }
+
+    val deadline = System.currentTimeMillis() + 2000
+    while (System.currentTimeMillis() < deadline && !pluginSettings().token.isNullOrEmpty()) {
+      Thread.sleep(20)
```
basti-snyk: **Should-Fix: same sleep-poll flake risk as `SnykCliAuthenticationServiceTest`.**

The 2000ms `Thread.sleep(20)` poll after `verify(timeout = 2000)` can lapse under CI load. Replace with deterministic synchronization (latch signalled by the production path, or drain the pooled executor). — AI review

src/main/kotlin/io/snyk/plugin/services/SnykCliAuthenticationService.kt
```diff
@@ -35,6 +38,30 @@ import snyk.common.lsp.LanguageServerWrapper
...
+
+      ApplicationManager.getApplication().invokeLater {
+        if (project.isDisposed) return@invokeLater
+        getSnykToolWindowPanel(project)?.cleanUiAndCaches()
+      }
+      // Re-render the tool window; with an empty token it falls back to the authentication panel.
+      publishAsync(project, SnykSettingsListener.SNYK_SETTINGS_TOPIC) { settingsChanged() }
```
basti-snyk: **Suggestion: the auth panel is rebuilt twice on logout.**

`cleanUiAndCaches()` -> `doCleanUi(true)` -> `displayEmptyDescription()` already sees the empty token and calls `displayAuthPanel()` (builds `SnykAuthPanel` #1). This `settingsChanged` publish then reaches `chooseMainPanelToDisplay()` -> `displayAuthPanel()` again, disposing #1 and building #2 — one wasted construct/dispose cycle per logout. It's not pure redundancy (`settingsChanged` also notifies other subscribers), but consider letting `settingsChanged` own the panel switch and `cleanUiAndCaches` own only cache scrubbing. — AI review

src/main/kotlin/io/snyk/plugin/services/SnykCliAuthenticationService.kt
```diff
@@ -35,6 +38,30 @@ import snyk.common.lsp.LanguageServerWrapper
...
+   * do not block. Clearing the local token is what stops it being re-sent to the LS on the next
+   * workspace/didChangeConfiguration push.
+   */
+  fun logout() {
+    ApplicationManager.getApplication().executeOnPooledThread {
+      if (project.isDisposed) return@executeOnPooledThread
+      // Clear the local token before logging the LS out. If the LS pulls workspace/configuration
+      // while handling logout, getSettings() would otherwise hand the stale token straight back
+      // and re-authenticate the server we just logged out.
+      pluginSettings().token = ""
```
basti-snyk: **Suggestion (minor): set the token to `null` rather than `""` for the "unset" state.**

Behaviorally fine here — every consumer on this path uses `isNullOrEmpty`/`isNullOrBlank` — but the codebase's convention for "no token" is `null` (e.g. the tests' `tearDown` sets `token = null`). Using `null` keeps the sentinel consistent. — AI review

src/main/kotlin/io/snyk/plugin/services/SnykCliAuthenticationService.kt
rrama: Won't fix — the double panel construct/dispose only causes a slight flash on logout, which is a rare, low-traffic UX path. Splitting the responsibilities of `cleanUiAndCaches()` and `settingsChanged` touches shared code used well beyond logout, so the refactor blast radius outweighs the benefit here.

rrama:
- **CHANGELOG:** no longer hand-maintained — release notes are auto-generated (`gh release create --generate-notes`); the dead checkbox is removed in #859. No entry needed here.
- **Find-Action-only:** intentional — that's the only intended place for this action. Separately I'm proposing we also surface it on the HTML settings *fallback* page (the fallback re-auth control from snyk/snyk-ls#1371): expose a plain "Log out" / "Clear credentials" button there rather than re-implementing the full IDE-side auth flow.
- **Multi-project staleness / plaintext token:** pre-existing, out of scope here.

src/main/kotlin/io/snyk/plugin/services/SnykCliAuthenticationService.kt
rrama: Fixed — local UI/cache cleanup and the `settingsChanged` publish now run in a `finally`, so a non-`TimeoutException` LS logout failure can no longer leave stale findings on screen. Local token clear still happens first. `1237e775`.

src/test/kotlin/io/snyk/plugin/services/SnykCliAuthenticationServiceTest.kt
rrama: Fixed — dropped the sleep-poll. The local token clear runs on the pooled thread strictly before the LS `logout()` call, so `verify(timeout = 2000) { logout() }` establishes the happens-before; the token assertion right after is race-free. `1237e775`.

src/test/kotlin/io/snyk/plugin/ui/actions/SnykLogoutActionTest.kt
rrama: Fixed — same deterministic fix as the service test, sleep-poll removed. `1237e775`.

src/main/kotlin/io/snyk/plugin/services/SnykCliAuthenticationService.kt
rrama: Done — `token = null` now. The existing `""` clear in `SnykProjectSettingsConfigurable` is left as-is (out of scope); both behave identically since all readers use `isNullOrEmpty`/`isNullOrBlank`. `1237e775`.

rrama:
Failing security scan will be done as part of: https://github.com/snyk/snyk-intellij-plugin/pull/862

rrama:
Addressed in `026dbd5c`:
- **Coverage:** added a test asserting `cleanUiAndCaches()` runs on logout (`SnykLogoutActionTest`, via a latch + `waitWithEventsDispatching` so it is race-free), and real coverage of the LS-uninitialized branch in `LanguageServerWrapperTest` (asserts `COMMAND_LOGOUT` is not sent when the LS is not initialized). The earlier service-level test that mocked the whole wrapper — never actually hitting that branch — was removed.
- **plugin.xml `Snyk:` rename:** intentional, matches the VS Code extension's action naming. Keeping it.
- **Silent server-side logout failure:** a non-`TimeoutException` from the LS logout propagates out of the pooled thread and is logged to `idea.log`, so it is recorded (just not user-facing). A toast is optional polish, out of scope here.

rrama:
Re: the "Missing Notification Topic" note — won't change. `SNYK_ENABLEMENT_TOPIC` has a single subscriber, `AnnotatorCommon`, which also subscribes to `SNYK_SETTINGS_TOPIC` and runs the identical `refreshAnnotationsForOpenFiles` handler for both. `logout()` already publishes `SNYK_SETTINGS_TOPIC`, so annotators already refresh; adding the enablement topic would just fire the same handler twice. No enablement-only listener exists.

nick-y-snyk (approved):

PR #860, opened 2026-07-09
fix: add re-auth control to settings fallback page (plugin mirror) [IDE-2181]

basti-snyk:
## Description

The IntelliJ plugin ships its own copy of the settings fallback page. This mirrors the language-server change ([snyk/snyk-ls#1371](https://github.com/snyk/snyk-ls/pull/1371)) into it **byte-identically** so an IntelliJ user dropped to the limited fallback page can re-authenticate.

The Authentication section gains a **"Connect"** control that dispatches `snyk.login` through the plugin command bridge with no args (reusing the configured method/endpoint), with the same safety-timeout recovery and single-shot cycle-aware callback latch as the LS copy.

## Changes

- `src/main/resources/html/settings-fallback.html` — add re-auth "Connect" control mirroring the LS fallback page.
- Kotlin test asserting the shipped fallback resource wires the Connect control to a `snyk.login` dispatch and that the command bridge forwards it with the login timeout.

The two fallback copies (plugin + LS) must remain byte-identical.

## Testing

Covered by `SettingsFallbackReauthDispatchTest`. Full verification runs in CI (no JDK in the local dev container).

Jira: IDE-2181

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -96,3 +96,4 @@ DESIGN.md
 .snyk-code-output.err
 .verification-result.*
 docs/plan.md
+.verify-report
--- a/src/main/resources/html/settings-fallback.html
+++ b/src/main/resources/html/settings-fallback.html
@@ -59,6 +59,10 @@
       --border-color: var(--vscode-panel-border, #454545);
       --focus-border: var(--vscode-focusBorder, #007acc);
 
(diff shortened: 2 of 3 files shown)


rrama:
Closing, since we have a "Clear credentials" button now.

PR #863, opened 2026-07-10
ci: bound the Test job on hang and capture thread dumps [IDE-2237]

rrama:
## What

Stop a hung `Test` job from stalling CI for hours, and make a rare hang diagnosable from the run's artifacts alone.

Three parts:

### 1. Bound the hang — layered timeouts

| Tier | Where | Value | Role |
|------|-------|-------|------|
| Gradle `test` task timeout | `build.gradle.kts` | **25m** | Fails an *interruptible* hung test with a precise "test timed out" message. Scoped to the `test` task (not `withType<Test>`) so heavier Test tasks like a future `testIdeUi` don't inherit a limit sized for the unit suite. |
| Test job `timeout-minutes` | `.github/workflows/build.yml` | **40m** | The real backstop — catches hangs the task timeout can't (wedged daemon, hung compile/spotless/kover, uninterruptible native I/O). |

No per-step timeout on purpose: a step timeout low enough to be useful would fire before the 25m Gradle timeout (compile+lint run first), degrading a precise "test timed out" failure into a generic "step timed out".

### 2. Diagnose the hang — thread-dump capture

`.github/scripts/hang-watchdog.sh` runs alongside `./gradlew` (backgrounded, torn down by a trap). It sleeps past the normal run time (`HANG_THRESHOLD_SECONDS`, default 20m — above the ~16m real-run max, below the 25m kill), then every 2m captures the Gradle test-worker JVM(s):
- `jstack -l` → a file per dump (primary; `-l` gives lock ownership)
- `kill -3` (SIGQUIT) → the JVM's own stdout (Unix best-effort)

Dumps upload as the `test-hang-dumps-<os>` artifact (`if: always()`). The `test-results` artifact also uploads on failure. On a healthy run the watchdog is killed while still sleeping and adds nothing.

### 3. Why this matters

The hang (IDE-2237) is intermittent and leaves no ordinary evidence — the worker goes silent and is killed. A thread dump taken while it is wedged is the one artifact that identifies the blocked thread. Capturing it in CI means a 1-in-N hang is diagnosable without a local repro.

## Validation (with links)

- **Timeout bounds the hang.** A real hang was killed by the 25m Gradle timeout instead of running to the 6h default — [Test (windows-latest), run 29255954069](https://github.com/snyk/snyk-intellij-plugin/actions/runs/29255954069/job/86844286732): `Execution failed for task ':test' > Timeout has been exceeded`, `BUILD FAILED in 29m 37s`. Earlier instances: [macOS run 29102204741](https://github.com/snyk/snyk-intellij-plugin/actions/runs/29102204741/job/86393680974), [ubuntu run 29110833595](https://github.com/snyk/snyk-intellij-plugin/actions/runs/29110833595/job/86422520894).
- **Watchdog detects a real hang.** On that same windows hang the watchdog fired and located the wedged test-worker JVM (log: `[hang-watchdog] iter N ... pid 7484`). That run also surfaced a bug — `./gradlew clean` had wiped the dump dir, so writes failed — fixed in this PR (recreate the dir each iteration + report write failures).
- **Capture works end-to-end.** [Run 29326207693](https://github.com/snyk/snyk-intellij-plugin/actions/runs/29326207693) (threshold temporarily lowered to force the watchdog on a healthy run) produced populated `test-hang-dumps-<os>` artifacts on all three OSes — each a real `Full thread dump` containing the `Test worker` and `AWT-EventQueue-0` stacks plus `jstack -l` lock ownership, the exact signal needed to diagnose the deadlock.

Normal `Test` durations (last 8 successful `master` runs): Linux ~6–7m, macOS ~8m, Windows ~13m (max 16m) — all well under the 25m timeout, so it won't false-trip.

## Out of scope

- The underlying flaky hang itself (IDE-2237): `UtilsKtTest` leaks `runAsync`/`invokeLater` work past teardown, deadlocking a later test's EDT dispatch. Fixed separately; this PR bounds and diagnoses it.
- `pluginVerifier` job also lacks a timeout (still the 6h default) — follow-up.

## Test plan

- [x] `./gradlew help` — config parses (`named<Test>("test")` resolves)
- [x] `ktlintCheck` / `spotlessCheck` — BUILD SUCCESSFUL
- [x] `hang-watchdog.sh` — `bash -n` + local smoke (no-worker + dir-recreate-after-clean paths)
- [x] Workflow YAML validated
- [x] Timeout fired on a real hang
- [x] Capture produced usable thread dumps in CI

### DIFF
--- a/.github/scripts/hang-watchdog.sh
+++ b/.github/scripts/hang-watchdog.sh
@@ -26,21 +26,26 @@ capture() { # $1 = pid, $2 = outfile
   "${JAVA_BIN}jstack" -l "$1" >"$2" 2>&1 || "${JAVA_BIN}jcmd" "$1" Thread.print >"$2" 2>&1
 }
 
-mkdir -p "$DUMP_DIR"
 sleep "$THRESHOLD"
 
 iter=0
 while true; do
   iter=$((iter + 1))
   ts="$(date +%Y%m%d-%H%M%S)"
+  # (Re)create the dump dir every iteration: `./gradlew clean` deletes build/ at the start of the
+  # step, so a dir created once up-front is gone by the time we dump.
+  mkdir -p "$DUMP_DIR"
   pids="$("${JAVA_BIN}jps" -l 2>/dev/null | grep 'GradleWorkerMain' | awk '{print $1}')"
   if [ -z "$pids" ]; then
     echo "[hang-watchdog] iter ${iter} @ ${ts}: no GradleWorkerMain process found"
   else
     for pid in $pids; do
       out="${DUMP_DIR}/threaddump-${ts}-pid${pid}-${iter}.txt"
-      capture "$pid" "$out"
-      echo "[hang-watchdog] iter ${iter} @ ${ts}: wrote ${out} for pid ${pid}"
+      if capture "$pid" "$out"; then
+        echo "[hang-watchdog] iter ${iter} @ ${ts}: wrote ${out} for pid ${pid}"
+      else
+        echo "[hang-watchdog] iter ${iter} @ ${ts}: FAILED to capture pid ${pid}"
+      fi
       kill -3 "$pid" 2>/dev/null || true
     done
   fi
--- a/.github/scripts/hang-watchdog.sh
+++ b/.github/scripts/hang-watchdog.sh
@@ -41,7 +41,13 @@ run_bounded() { # $1 = outfile; rest = command...
   local p=$! waited=0
   while kill -0 "$p" 2>/dev/null; do
     if [ "$waited" -ge "$CAPTURE_TIMEOUT" ]; then
+      # A capture wedged in JVM dynamic-attach may ignore SIGTERM; escalate to SIGKILL after a
+      # brief grace, then reap so stuck attach processes don't accumulate (zombies / a half-open
+      # attach socket) and degrade later captures in the series.
       kill "$p" 2>/dev/null || true
+      sleep 2
+      kill -0 "$p" 2>/dev/null && kill -9 "$p" 2>/dev/null || true
+      wait "$p" 2>/dev/null || true
       return 124
     fi
     sleep 2
@@ -51,11 +57,15 @@ run_bounded() { # $1 = outfile; rest = command...
 }
 
 capture() { # $1 = pid, $2 = outfile
-  run_bounded "$2" "${JAVA_BIN}jstack" -l "$1" && return 0
-  # jstack failed or timed out. If it left a partial dump, keep it — do NOT let the jcmd fallback
-  # truncate it (both use `>` on the same file). Only fall back when jstack produced nothing.
-  [ -s "$2" ] && return 1
+  run_bounded "$2" "${JAVA_BIN}jstack" -l "$1"
+  # Keep the jstack output only if it is a real thread dump (even a partial one, from a timeout)
+  # rather than an error string — `run_bounded` merges stderr into the file, so a fast attach
+  # failure ("Unable to open socket file", AttachNotSupportedException, wrong pid) also makes the
+  # file non-empty. A dump always contains a thread-state line; an error string does not. If it is
+  # a real dump, keep it and don't let the jcmd fallback truncate it; otherwise fall back to jcmd.
+  grep -q 'java.lang.Thread.State' "$2" 2>/dev/null && return 0
   run_bounded "$2" "${JAVA_BIN}jcmd" "$1" Thread.print
+  grep -q 'java.lang.Thread.State' "$2" 2>/dev/null
 }
 
 sleep "$THRESHOLD"
--- a/.github/scripts/hang-watchdog.sh
+++ b/.github/scripts/hang-watchdog.sh
@@ -7,23 +7,52 @@
 # a healthy run: that finishes before the threshold and the caller kills this
 # watchdog while it is still sleeping.
 #
-# Both capture mechanisms are used:
+# Scope: this targets the Gradle *test worker* (`GradleWorkerMain`), which is where
+# the known hang lives (IDE-2237: an EDT-dispatch deadlock in the test JVM). A hang
+# that instead wedges the Gradle daemon or a compile/lint phase has no test worker,
+# so this logs "no test-worker JVM found" and captures nothing — that means "not a
+# test-worker hang", NOT "nothing wrong"; such a run is still bounded by the job
+# `timeout-minutes` backstop, just without a thread dump.
+#
+# Capture uses two mechanisms:
 #   1. `jstack -l` -> a file per dump (primary; `-l` includes lock ownership).
 #   2. `kill -3` (SIGQUIT) -> the JVM's own stdout (Unix best-effort; on Windows a
 #      JVM dumps on Ctrl-Break rather than SIGQUIT, so this is a no-op there and
 #      jstack remains the source of truth).
+# Each capture is time-bounded: jstack/jcmd attach via the JVM dynamic-attach
+# socket, which can itself hang if the target is wedged in native code or a GC
+# stall. Bounding it keeps the watchdog loop alive so later iterations still fire.
 #
 # See .local/test-hang-capture-plan.md and the IDE-2237 investigation.
 set -uo pipefail
 
-THRESHOLD="${HANG_THRESHOLD_SECONDS:-1200}"      # 20m: above the ~16m real-run max, below the 25m Gradle test timeout
-INTERVAL="${HANG_DUMP_INTERVAL_SECONDS:-120}"    # dump every 2m thereafter
+THRESHOLD="${HANG_THRESHOLD_SECONDS:-1200}"          # 20m: above the ~16m real-run max, below the 25m Gradle test timeout
+INTERVAL="${HANG_DUMP_INTERVAL_SECONDS:-120}"        # dump every 2m thereafter
+CAPTURE_TIMEOUT="${HANG_CAPTURE_TIMEOUT_SECONDS:-60}" # bound a single jstack/jcmd attach
 DUMP_DIR="${HANG_DUMP_DIR:-build/hang-dumps}"
 
 JAVA_BIN="${JAVA_HOME:+${JAVA_HOME}/bin/}"
 
+# Run a command with its stdout+stderr to $out, killed if it exceeds CAPTURE_TIMEOUT.
+# Portable (no coreutils `timeout`, which is absent on macOS runners). Returns 124 on timeout.
+run_bounded() { # $1 = outfile; rest = command...
+  local out="$1"; shift
+  "$@" >"$out" 2>&1 &
+  local p=$! waited=0
+  while kill -0 "$p" 2>/dev/null; do
+    if [ "$waited" -ge "$CAPTURE_TIMEOUT" ]; then
+      kill "$p" 2>/dev/null || true
+      return 124
+    fi
+    sleep 2
+    waited=$((waited + 2))
+  done
+  wait "$p"
+}
+
 capture() { # $1 = pid, $2 = outfile
-  "${JAVA_BIN}jstack" -l "$1" >"$2" 2>&1 || "${JAVA_BIN}jcmd" "$1" Thread.print >"$2" 2>&1
+  run_bounded "$2" "${JAVA_BIN}jstack" -l "$1" && return 0
+  run_bounded "$2" "${JAVA_BIN}jcmd" "$1" Thread.print
 }
 
 sleep "$THRESHOLD"
@@ -37,14 +66,14 @@ while true; do
   mkdir -p "$DUMP_DIR"
   pids="$("${JAVA_BIN}jps" -l 2>/dev/null | grep 'GradleWorkerMain' | awk '{print $1}')"
   if [ -z "$pids" ]; then
-    echo "[hang-watchdog] iter ${iter} @ ${ts}: no GradleWorkerMain process found"
+    echo "[hang-watchdog] iter ${iter} @ ${ts}: no test-worker JVM found (a daemon/compile-phase hang is not captured here)"
   else
     for pid in $pids; do
       out="${DUMP_DIR}/threaddump-${ts}-pid${pid}-${iter}.txt"
       if capture "$pid" "$out"; then
         echo "[hang-watchdog] iter ${iter} @ ${ts}: wrote ${out} for pid ${pid}"
       else
-        echo "[hang-watchdog] iter ${iter} @ ${ts}: FAILED to capture pid ${pid}"
+        echo "[hang-watchdog] iter ${iter} @ ${ts}: capture FAILED/timed out for pid ${pid} (kept partial ${out})"
       fi
       kill -3 "$pid" 2>/dev/null || true
     done
--- a/.github/scripts/hang-watchdog.sh
+++ b/.github/scripts/hang-watchdog.sh
@@ -1,0 +1,48 @@
+#!/usr/bin/env bash
+# Hang watchdog for the Gradle `test` task.
+#
+# Sleeps past the normal run time, then repeatedly captures thread dumps of the
+# Gradle test-worker JVM(s) so an intermittent CI hang can be diagnosed from the
+# run's uploaded artifacts alone, with no local reproduction. It never triggers on
+# a healthy run: that finishes before the threshold and the caller kills this
+# watchdog while it is still sleeping.
+#
+# Both capture mechanisms are used:
+#   1. `jstack -l` -> a file per dump (primary; `-l` includes lock ownership).
+#   2. `kill -3` (SIGQUIT) -> the JVM's own stdout (Unix best-effort; on Windows a
+#      JVM dumps on Ctrl-Break rather than SIGQUIT, so this is a no-op there and
+#      jstack remains the source of truth).
+#
+# See .local/test-hang-capture-plan.md and the IDE-2237 investigation.
+set -uo pipefail
+
+THRESHOLD="${HANG_THRESHOLD_SECONDS:-1200}"      # 20m: above the ~16m real-run max, below the 25m Gradle test timeout
+INTERVAL="${HANG_DUMP_INTERVAL_SECONDS:-120}"    # dump every 2m thereafter
+DUMP_DIR="${HANG_DUMP_DIR:-build/hang-dumps}"
+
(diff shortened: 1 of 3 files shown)


build.gradle.kts
```diff
@@ -182,6 +183,15 @@ tasks {
...
   }
 
+  // Wall-clock cap on the unit-test task only (deliberately NOT withType<Test>, so heavier Test
+  // tasks such as a future testIdeUi don't inherit this 25m limit sized for the unit suite). It
+  // bounds an *interruptible* hung test with a precise "test timed out" failure, locally and in
+  // CI, and sits above the ~16m worst-case real run so it won't false-trip. It is not a guarantee:
+  // a test wedged in uninterruptible native I/O can ignore the interrupt, so the GitHub Actions
+  // `test` job `timeout-minutes` (40m) in .github/workflows/build.yml is the real backstop. Keep
+  // this below that job timeout.
+  named<Test>("test") { timeout.set(Duration.ofMinutes(25)) }
```
rrama: AI comment:

> **Task configuration ordering: named<Test> before withType<Test> is safe**
> 
> The `named<Test>("test")` block at line 193 sets `timeout` before the `withType<Test>` block at line 195. In Gradle, both are lazy configurations applied in declaration order. Since `withType<Test>` does not set `timeout`, the value from `named<Test>("test")` persists. However, if a future contributor adds a `timeout` assignment inside `withType<Test>`, it would silently override this 25m limit for the "test" task. The comment on line 186-192 documents the intent well, but the ordering dependency is subtle.

rrama (commented):
AI review:

> The comments in this PR are extremely verbose — they explain the reasoning, history, and cross-references far beyond what's useful inline. Here's my take:
> 
> build.gradle.kts
> The comment for named<Test>("test") { timeout.set(Duration.ofMinutes(25)) } is 6 lines for a 1-line change. Issues:
> 
> It duplicates rationale that lives better in the PR description / commit message.
> "sits above the ~16m worst-case real run" is a timestamped fact that will silently rot — runs get slower, this number becomes wrong, and nobody updates it.
> The cross-reference to build.yml is fine, but the reciprocal reference already exists in the YAML, so the two comments will drift out of sync.
> Suggest trimming to intent + the key constraint:
> 
> // 25m wall-clock cap on the unit-test task only (not withType<Test>, so heavier tasks
> // like a future testIdeUi don't inherit it). Catches interruptible hangs; the GitHub job
> // timeout in build.yml is the real backstop, so keep this below it.
> named<Test>("test") { timeout.set(Duration.ofMinutes(25)) }
> .github/workflows/build.yml
> Three comment blocks, all long:
> 
> The timeout-minutes: 40 comment (7 lines) and the step comment both re-explain the "no per-step timeout on purpose" reasoning. That's stated twice plus a third time in the PR description. One place is enough.
> The "Linters listed before check" comment mixes two unrelated topics — lint ordering and the timeout tiers — which makes it harder to follow.
> Both files essentially restate the same three-way relationship (task timeout ↔ job timeout ↔ no step timeout) in every location. Consolidating to a single authoritative explanation and short pointers elsewhere would cut the noise significantly.

rrama:
Re: https://github.com/snyk/snyk-intellij-plugin/pull/863#pullrequestreview-4673215004

Good call — trimmed. Dropped the "~16m worst-case" figure (agreed it rots), consolidated the "no per-step timeout" reasoning to one authoritative spot on the job `timeout-minutes`, and reduced the step comment to a pointer. Done in 9a086b3a.

build.gradle.kts
rrama: Fixed by removing the ordering dependency: moved `named<Test>("test")` to *after* the `withType<Test>` block, so declaration order makes the task-scoped 25m timeout win regardless of what `withType` sets. No more silent-override risk. (9a086b3a)

rrama:
Thanks — both are intentional / already-mitigated, no change needed:

1. **`kill -3` no-op on Windows** — correct. It's best-effort by design; the `jstack -l`/`jcmd` dump *files* are the primary path and upload as the `test-hang-dumps-<os>` artifact on all three OSes, so Windows still gets a full thread dump (just not an inline stdout copy in the log). Called out in the script comment.

2. **25m timeout doesn't cover `integTest`** — there's no `integTest` Gradle task in this build today (CI runs `check`, i.e. the unit `test` task; CONTRIBUTING mentions an `integTest` source root as a convention, not a wired task). The 25m is deliberately scoped to `test` so heavier suites aren't over-constrained. If an `integTest` task is added later it'd get its own sizing — and as you note, `hang-watchdog.sh` + the 40m job timeout already cover any `GradleWorkerMain` regardless.

.github/scripts/hang-watchdog.sh
```diff
@@ -0,0 +1,53 @@
...
+sleep "$THRESHOLD"
+
+iter=0
+while true; do
+  iter=$((iter + 1))
+  ts="$(date +%Y%m%d-%H%M%S)"
+  # (Re)create the dump dir every iteration: `./gradlew clean` deletes build/ at the start of the
+  # step, so a dir created once up-front is gone by the time we dump.
+  mkdir -p "$DUMP_DIR"
+  pids="$("${JAVA_BIN}jps" -l 2>/dev/null | grep 'GradleWorkerMain' | awk '{print $1}')"
```
basti-snyk, line 80: **Diagnostic-coverage gap:** the watchdog greps only `GradleWorkerMain`, so it captures dumps only for a hung *test worker*. But the job comment in `build.yml` names "wedged daemon, hung compile/lint" as scenarios the backstop exists for — those run in the Gradle daemon / launcher JVM, with no test worker present. In that case this prints `no GradleWorkerMain process found` every interval and captures nothing, giving zero diagnostic value for a subset of the very hangs IDE-2237 targets.

Consider also matching the Gradle daemon (`GradleDaemon`) / `KotlinCompileDaemon`, or dumping all JDK-owned JVMs (minus `jps` itself). At minimum, document the gap so a "no worker found" run isn't mistaken for "nothing to diagnose".

— AI review

.github/scripts/hang-watchdog.sh
```diff
@@ -0,0 +1,53 @@
...
+set -uo pipefail
+
+THRESHOLD="${HANG_THRESHOLD_SECONDS:-1200}"      # 20m: above the ~16m real-run max, below the 25m Gradle test timeout
+INTERVAL="${HANG_DUMP_INTERVAL_SECONDS:-120}"    # dump every 2m thereafter
+DUMP_DIR="${HANG_DUMP_DIR:-build/hang-dumps}"
+
+JAVA_BIN="${JAVA_HOME:+${JAVA_HOME}/bin/}"
+
+capture() { # $1 = pid, $2 = outfile
+  "${JAVA_BIN}jstack" -l "$1" >"$2" 2>&1 || "${JAVA_BIN}jcmd" "$1" Thread.print >"$2" 2>&1
```
basti-snyk: **`capture()` can itself hang on the exact failure mode it targets.** `jstack -l` and the `jcmd` fallback both attach via the JVM dynamic-attach socket, which needs the target's attach-listener thread to respond. A JVM wedged in a native lock, a GC stall, or a deadlock involving that thread will hang the attach — so `capture` blocks forever, the `while` loop never reaches the next `sleep "$INTERVAL"`, and you get a single truncated dump instead of the intended series.

Consider wrapping the capture in a bounded `timeout` (note: not present on macOS by default — `gtimeout`/coreutils or a background-kill fallback) and continuing the loop on capture-timeout so later iterations still fire.

— AI review

.github/workflows/build.yml
```diff
@@ -43,11 +47,44 @@ jobs:
...
+        # A background watchdog captures thread dumps of the test worker if it runs abnormally long,
+        # so an intermittent hang (see the Gradle `test` timeout in build.gradle.kts) is diagnosable
+        # from the uploaded dumps alone. `shell: bash` so the one script runs on all three OSes
+        # (git-bash on Windows). The trap tears the watchdog down whether gradle passes, fails, or is
+        # killed by its task timeout.
+        shell: bash
+        run: |
+          bash .github/scripts/hang-watchdog.sh &
+          WATCHDOG_PID=$!
+          trap 'kill "$WATCHDOG_PID" 2>/dev/null || true' EXIT
```
basti-snyk, line 63: **EXIT-trap teardown is ineffective on the healthy path and can stall the step.** `$WATCHDOG_PID` is the *bash* process; on a normal run gradle finishes while the watchdog is still blocked in its initial `sleep "$THRESHOLD"`. `kill "$WATCHDOG_PID"` terminates bash but leaves the `sleep` child orphaned, still holding the runner's log pipe fd. GitHub's log reader waits for that fd to hit EOF, so the step can stall after tests pass — bounded only by the new 40m job timeout. That's the exact failure class this PR aims to eliminate, reintroduced in the cleanup path, and it would affect every run rather than only hung ones.

Two options that fix it: redirect the watchdog's stdio to a file (`bash .github/scripts/hang-watchdog.sh >watchdog.log 2>&1 &`) so no orphan holds the log pipe, and/or kill the process tree rather than just the parent.

— AI review

.github/workflows/build.yml
```diff
@@ -19,6 +19,10 @@ jobs:
     name: Test
     needs: gradleValidation
     runs-on: ${{ matrix.os }}
+    # Real backstop for hangs the Gradle `test` task timeout (in build.gradle.kts) can't catch
+    # — wedged daemon, hung compile/lint, or uninterruptible native I/O. No per-step timeout: a
+    # step limit could fire before that task timeout and mask the precise failure.
+    timeout-minutes: 40
```
basti-snyk, line 25: **A job-level `timeout-minutes` may skip the `if: always()` uploads for the case it exists to cover.** The interruptible path is fine: the 25m Gradle `test` task timeout fails the *step*, so both upload steps run and dumps/results are captured. But for a *non-interruptible* hang (wedged daemon / native I/O) — precisely what this 40m job backstop targets — GitHub cancels the whole job, and `if: always()` steps are not guaranteed to run during a job-timeout cancellation. If they don't, the `build/hang-dumps` captured between 20–40m are never uploaded, defeating "diagnosable from the uploaded dumps alone".

Worth confirming on a real timed-out run. A step-level timeout on the Run Tests step would fail the step (not cancel the job) and guarantee the uploads run.

— AI review

.github/scripts/hang-watchdog.sh
rrama: Done (04c83f6b) — documented the scope rather than dumping every JVM. The watchdog intentionally targets the test worker, where the known hang (IDE-2237) lives; broadening to all JDK JVMs adds noise (idle daemon/compile workers) for hang classes we haven't observed. The "no worker found" log now reads "a daemon/compile-phase hang is not captured here" and the script header notes such a run is still bounded by the job timeout — so it isn't mistaken for "nothing to diagnose".

.github/scripts/hang-watchdog.sh
rrama: Done (04c83f6b) — each capture is now time-bounded via `run_bounded` (`HANG_CAPTURE_TIMEOUT_SECONDS`, default 60s): jstack/jcmd runs backgrounded and is killed if the attach hangs, so the loop keeps firing and produces the intended series instead of one truncated dump. Portable, no coreutils `timeout` (absent on macOS runners).

.github/workflows/build.yml
rrama: Done (04c83f6b) — the watchdog's stdio is now redirected to `hang-watchdog.log` (uploaded alongside the dumps) instead of inheriting the step's log pipe. With nothing holding that fd open, an orphaned child can no longer stall the step at EOF on the healthy path. Good catch.

.github/workflows/build.yml
rrama: Going to leave this one, for now. Our hang is interruptible: the 25m Gradle `test` task timeout fails the *step*, so the `if: always()` upload steps run — confirmed on a real run where the `test-hang-dumps` artifact uploaded from a step-failure. A non-interruptible hang riding to the 40m *job* timeout is a case we haven't observed, and `always()` does run on cancellation per GH docs. A step-level timeout would also re-reverse the "no per-step timeout" decision from the earlier review round (below 25m it masks the precise Gradle failure; above 25m it's redundant with the job timeout for this theoretical case). Happy to add one if a non-interruptible hang actually turns up.

rrama:
Thanks — addressed the major, the two minors were already handled:

- 🟠 **Diagnostic Data Loss (jcmd truncating a partial jstack dump)** — real bug, fixed in 69814e99. `capture()` now only falls back to `jcmd` when `jstack` produced nothing (`! -s "$out"`); if `jstack` timed out leaving a partial dump, that dump is kept instead of being truncated by the fallback. Verified the partial is preserved and the loop still continues.
- 🟡 **Portability (jps/grep 'GradleWorkerMain')** — intentional scope, documented: the watchdog targets the test worker (where IDE-2237 lives). A run with no worker logs "no test-worker JVM found (a daemon/compile-phase hang is not captured here)" and the header notes it's still bounded by the job timeout — so it's not mistaken for "nothing to diagnose". Broadening to all JVMs was declined to avoid dumping idle daemon/compile workers as noise.
- 🟡 **`kill -3` no-op on Windows** — expected and documented as Unix best-effort; `jstack` (file) is the source of truth on all OSes, and the `|| true` makes the no-op harmless.

rrama:
Re: Partial File Overwrite (🟡 minor) — this is already prevented by the `[ -s "$2" ]` guard; leaving the code as-is.

`run_bounded` **kills jstack before it returns** (SIGTERM on timeout, then returns 124), so by the time `capture()` checks `[ -s "$2" ]` the jstack process is already terminated. If jstack left *any* partial dump, `-s` is true → `capture()` returns and **never reaches the jcmd line**. jcmd runs only when the file is empty (jstack wrote nothing), so there is no partial output to overwrite.

The window described — a killed jstack writing its first byte in the microseconds *after* the `-s` check, racing jcmd's truncate — can't leave meaningful data: if jstack had written anything by the check, `-s` would already have short-circuited to jcmd being skipped. Happy to split the fallback into a separate temp file if we ever want it structurally impossible, but it's not a real data-loss path today.

.github/scripts/hang-watchdog.sh
```diff
@@ -0,0 +1,85 @@
...
+    waited=$((waited + 2))
+  done
+  wait "$p"
+}
+
+capture() { # $1 = pid, $2 = outfile
+  run_bounded "$2" "${JAVA_BIN}jstack" -l "$1" && return 0
+  # jstack failed or timed out. If it left a partial dump, keep it — do NOT let the jcmd fallback
+  # truncate it (both use `>` on the same file). Only fall back when jstack produced nothing.
+  [ -s "$2" ] && return 1
```
basti-snyk: **Should-Fix (new since last review) — the `jcmd` fallback is bypassed whenever `jstack` fails *fast*, not only when it leaves a partial dump.** `run_bounded` merges stderr into the outfile (`"$@" >"$out" 2>&1`), so a quick `jstack` attach failure (`Unable to open socket file: target process not responding`, `AttachNotSupportedException`, wrong PID) writes its *error text* to `$2`, making `[ -s "$2" ]` true — so `return 1` fires and line 58's `jcmd` fallback is never reached. The `-s` guard (added to stop `jcmd` truncating a genuine partial dump) conflates "partial thread dump" with "error message", so in the most common attach-failure mode the fallback that exists precisely for that case is skipped and the artifact is just a `jstack` error string.

Distinguish a real dump from an error before keeping it — e.g. gate the `return 1` on a dump marker (`grep -q 'java.lang.Thread.State' "$2"`) rather than mere non-emptiness, or only keep-and-return when `run_bounded` returned 124 (timeout) vs a fast non-zero exit.

— AI review

.github/scripts/hang-watchdog.sh
```diff
@@ -0,0 +1,85 @@
...
+# Run a command with its stdout+stderr to $out, killed if it exceeds CAPTURE_TIMEOUT.
+# Portable (no coreutils `timeout`, which is absent on macOS runners). Returns 124 on timeout.
+run_bounded() { # $1 = outfile; rest = command...
+  local out="$1"; shift
+  "$@" >"$out" 2>&1 &
+  local p=$! waited=0
+  while kill -0 "$p" 2>/dev/null; do
+    if [ "$waited" -ge "$CAPTURE_TIMEOUT" ]; then
+      kill "$p" 2>/dev/null || true
+      return 124
```
basti-snyk, line 51: **Should-Fix (new since last review) — `run_bounded`'s timeout path only sends SIGTERM; it never escalates or reaps.** On timeout it does `kill "$p"; return 124` with no `wait`. In the exact case the bound exists for — a `jstack`/`jcmd` wedged in JVM dynamic-attach — SIGTERM may not land promptly, and the child is neither `kill -9`'d nor `wait`ed. Across the ~10 dump iterations of a 20m→40m window, stuck attach processes can accumulate (each leaving a zombie and potentially holding the target's attach socket half-open), which can make *subsequent* captures fail to attach — degrading the very series of dumps this is meant to collect.

After `CAPTURE_TIMEOUT`: `kill "$p"`, brief grace, then `kill -9 "$p"` if still alive, and `wait "$p" 2>/dev/null || true` to reap.

— AI review

.github/scripts/hang-watchdog.sh
rrama: Fixed in be22c9dd — `capture()` now gates keep-vs-fallback on a real thread-dump marker (`grep -q 'java.lang.Thread.State'`) rather than `[ -s ]`. A fast jstack attach error (which run_bounded writes into the file via `2>&1`) no longer counts as a dump, so the jcmd fallback runs; a real partial dump (has the thread-state line) is kept.

.github/scripts/hang-watchdog.sh
rrama: Fixed in be22c9dd — on timeout `run_bounded` now escalates SIGTERM → (2s grace) → SIGKILL and `wait`s to reap. A capture wedged in JVM dynamic-attach can no longer accumulate zombies / hold the attach socket half-open and degrade later captures in the series.

rrama:
Re: Partial Dump Overwrite (🟡) — leaving as-is. The jcmd fallback runs *only* when jstack produced no `java.lang.Thread.State` marker — i.e. an error string or nothing, which has no diagnostic value — so letting jcmd overwrite it loses nothing. A real (even partial) dump contains the marker and is kept; jcmd is never reached in that case. Separate files would only ever preserve a jstack error string.

basti-snyk (approved):

PR #871, opened 2026-07-16
test: await Utils async helpers to reduce cross-test leakage [IDE-2237]

rrama:
## What

Make the fire-and-forget `runAsync { … invokeLater { … } }` helpers in `Utils.kt` awaitable, and await them in the tests that drive them, to remove a source of cross-test async leakage behind the intermittent `:test` hang (IDE-2237).

- `Utils.kt`: `navigateToSource`, `refreshAnnotationsForOpenFiles`, `refreshAnnotationsForFile(project, virtualFile)` now **return** the `runAsync` `Promise<Unit>` (guards return `resolvedPromise()`). Production callers ignore the return — no behaviour change. `refreshAnnotationsForOpenFiles` also collects and awaits its per-file promises so the returned promise reflects them.
- `UtilsKtTest`: the two `navigateToSource` tests `blockingGet` the returned promise so the pooled async work is drained before the test returns; the out-of-bounds test mocks `FileEditorManager` so the selection-highlight `invokeLater` no longer throws a `ClassCastException` on the pooled thread after teardown.
- `SnykLanguageClientTest` / `AnnotatorCommonIntegTest`: mocks updated for the new `Promise` return type.

## Scope — what this does and does not do

**Does:** removes the specific pooled-async leak these helpers caused in the tests that call them for real (notably the `navigateToSource` CCE), and makes the helpers awaitable so a test can drain their pooled portion.

**Does not:** by itself eliminate the IDE-2237 hang. These promises resolve once the background thread *schedules* the UI work via `invokeLater` — not when the EDT task runs. So awaiting them fully drains the work only where `invokeLater` executes inline (as the tests mock it). A leaked `invokeLater` from any test that does not can still corrupt the shared EDT/`LaterInvocator` queue and deadlock a later test's dispatch. Eliminating that class fully needs a systemic between-test drain (flush the EDT + wait for the app pooled executor to idle), tracked on IDE-2237 — this PR is a scoped hardening step, not the whole fix.

## Test plan

- [x] `ktlintCheck` / `spotlessCheck` — BUILD SUCCESSFUL
- [x] Compiles; production caller behaviour unchanged (return value ignored)
- [x] Reviewed (no blocking findings)
- [ ] Does not, on its own, eliminate the IDE-2237 hang — see Scope

### DIFF
--- a/src/test/kotlin/io/snyk/plugin/UtilsKtTest.kt
+++ b/src/test/kotlin/io/snyk/plugin/UtilsKtTest.kt
@@ -337,7 +337,6 @@ class UtilsKtTest {
     mockkStatic("io.snyk.plugin.UtilsKt")
     mockkStatic(ApplicationManager::class)
     mockkStatic(PsiNavigationSupport::class)
-    mockkStatic(FileEditorManager::class)
 
     val appMock = mockk<com.intellij.openapi.application.Application>(relaxed = true)
     every { ApplicationManager.getApplication() } returns appMock
@@ -367,12 +366,8 @@ class UtilsKtTest {
     every { PsiNavigationSupport.getInstance() } returns psiNavSupport
     every { psiNavSupport.createNavigatable(project, virtualFile, 5) } returns navigatable
 
-    // Selection-highlight path early-returns on a null selected editor, so this mock only
-    // needs to keep the (unmocked-by-default) static call from throwing.
-    val fileEditorManager = mockk<FileEditorManager>(relaxed = true)
-    every { FileEditorManager.getInstance(project) } returns fileEditorManager
-    every { fileEditorManager.selectedTextEditor } returns null
-
+    // document == null, so the selection-highlight block (guarded by `document != null`) is
+    // skipped and FileEditorManager is never touched here — no mock needed for it in this test.
     val promise = navigateToSource(project, virtualFile, 5, 10)
 
     assertTrue("Navigation should complete within timeout", latch.await(2, TimeUnit.SECONDS))
--- a/src/test/kotlin/io/snyk/plugin/UtilsKtTest.kt
+++ b/src/test/kotlin/io/snyk/plugin/UtilsKtTest.kt
@@ -2,6 +2,7 @@ package io.snyk.plugin
 
 import com.intellij.ide.util.PsiNavigationSupport
 import com.intellij.openapi.application.ApplicationManager
+import com.intellij.openapi.fileEditor.FileEditorManager
 import com.intellij.openapi.project.Project
 import com.intellij.openapi.vfs.VirtualFile
 import com.intellij.pom.Navigatable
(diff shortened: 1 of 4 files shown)


src/test/kotlin/io/snyk/plugin/UtilsKtTest.kt
```diff
@@ -365,10 +367,19 @@ class UtilsKtTest {
     every { PsiNavigationSupport.getInstance() } returns psiNavSupport
     every { psiNavSupport.createNavigatable(project, virtualFile, 5) } returns navigatable
 
-    navigateToSource(project, virtualFile, 5, 10)
+    // Selection-highlight path early-returns on a null selected editor, so this mock only
```
basti-snyk: Suggestion (non-blocking): in this `getDocument returns null` test, `document` is null, so the selection-highlight block in `navigateToSource` is guarded by `if (document != null && selectionEndOffset != null)` and is skipped entirely — `FileEditorManager.getInstance(project)` is never reached. So the `FileEditorManager` mock (and `mockkStatic(FileEditorManager::class)`) added for this test is inert, and this comment describes a path that does not execute (it never reaches the "null selected editor" check; the whole block is skipped on `document == null`). Consider dropping the `FileEditorManager` setup here and correcting/removing the comment. The mock + comment in the out-of-bounds test below (where `document != null`) are accurate and required — keep those.

— AI review

rrama: Done in 768f82ea — dropped the inert `FileEditorManager` mock + `mockkStatic` from this `getDocument returns null` test and corrected the comment (document==null skips the selection-highlight block, so FileEditorManager is never reached). The out-of-bounds test keeps its mock. Thanks for the catch.

rrama:
Thanks — addressed both majors:

- **Incomplete Awaitability** — fixed in 768f82ea: `refreshAnnotationsForOpenFiles` now collects and `blockingGet`s the per-file promises, so the returned promise reflects them instead of resolving as soon as they're started.
- **Premature Promise Resolution** — correct, and it's an inherent limit of fire-and-forget-to-EDT: the promise resolves when the UI work is *scheduled* via `invokeLater`, not when the EDT runs it, so awaiting fully drains only where `invokeLater` executes inline (as the tests mock it). Rather than make production synchronously block on the EDT, I've reframed the PR as a scoped hardening step (see the updated description) — fully eliminating the cross-test EDT leak is the systemic between-test drain tracked on IDE-2237.

rrama:
Re: API Contract Change (🟡) — these are internal `io.snyk.plugin` helpers with no external/published consumers; every in-repo caller uses them as statements (return value ignored) and is unchanged. The Unit→Promise change is precisely what makes the async work awaitable, which is the point of the PR. No action.

basti-snyk (approved):

PR #878, opened 2026-07-28
fix: jcef with 2026.2 requires optional plugin declaration & resilience [IDE-2336]

basti-snyk:
### Description

Provide optional dependency and resiliency messages.

### Checklist

- [ ] Linted

### Screenshots / GIFs

### DIFF
--- a/src/main/kotlin/io/snyk/plugin/Utils.kt
+++ b/src/main/kotlin/io/snyk/plugin/Utils.kt
@@ -152,6 +152,8 @@ private inline fun <reified T : Any> Project.serviceIfNotDisposed(): T? {
   return try {
     getService(T::class.java)
   } catch (t: Throwable) {
+    // Without this the real cause is lost and the caller only ever sees a downstream NPE.
+    logger.error("Could not instantiate service ${T::class.java.name}", t)
     null
   }
 }
--- a/src/main/kotlin/io/snyk/plugin/ui/jcef/JcefAvailability.kt
+++ b/src/main/kotlin/io/snyk/plugin/ui/jcef/JcefAvailability.kt
@@ -1,0 +1,88 @@
+package io.snyk.plugin.ui.jcef
+
+import com.intellij.openapi.diagnostic.logger
+
+/** Why the IDE can or cannot give this plugin an embedded browser. */
+enum class JcefStatus {
+  /** The embedded browser is reachable and the runtime can host it. */
+  AVAILABLE,
+
+  /**
+   * The embedded browser classes are not reachable from this plugin's classloader.
+   *
+   * This is the IntelliJ 2026.2 case: JCEF moved out of the core platform into the separately
+   * bundled "Web Browser (JCEF)" plugin (`com.intellij.modules.jcef`). It is visible only to
+   * plugins that declare a dependency on it, and only when that bundled plugin actually resolved
+   * for this OS and architecture.
+   */
+  CLASSES_MISSING,
+
+  /** The classes are reachable, but the runtime cannot host a browser (e.g. a non-JBR runtime). */
+  RUNTIME_UNSUPPORTED,
+}
+
+/**
+ * The single seam through which embedded-browser (JCEF) availability is decided.
+ *
+ * This object must never reference a `com.intellij.ui.jcef` type — not in a signature, a field, a
+ * body, or a catch clause. That is why `JBCefApp.isSupported()` cannot be the guard: it lives in
+ * the very package that goes missing, so calling it throws the exact error it exists to prevent.
+ * The one call that does need a JCEF type is isolated in [JcefSupportProbe], which is only ever
+ * reached after the classes have been confirmed loadable.
+ */
+object JcefAvailability {
+  private val logger = logger<JcefAvailability>()
+
+  private const val PROBE_CLASS = "com.intellij.ui.jcef.JBCefApp"
+
+  private val memoisedStatus: JcefStatus by lazy { probe() }
+
+  fun status(): JcefStatus = memoisedStatus
+
+  fun isAvailable(): Boolean = status() == JcefStatus.AVAILABLE
+
+  /** Human-readable explanation of why the browser is unavailable, or null when it is available. */
+  fun unavailableReason(): String? =
+    when (status()) {
+      JcefStatus.AVAILABLE -> null
+      JcefStatus.CLASSES_MISSING ->
+        "This IDE did not provide its embedded browser component to the Snyk plugin. " +
+          "Since IntelliJ 2026.2 the embedded browser ships as the bundled " +
+          "\"Web Browser (JCEF)\" plugin, which must be installed and enabled."
+      JcefStatus.RUNTIME_UNSUPPORTED ->
+        "This IDE runtime cannot host an embedded browser. " +
+          "Run your IDE on a JetBrains Runtime (JBR) build to enable it."
+    }
+
+  private fun probe(): JcefStatus {
+    val classesReachable =
+      try {
+        // Deliberately not initialising the class: we only need to know it resolves.
+        Class.forName(PROBE_CLASS, false, JcefAvailability::class.java.classLoader)
+        true
+      } catch (t: Throwable) {
+        // The real failure mode is NoClassDefFoundError, which is an Error and not an Exception,
+        // so Throwable is the only catch that actually covers it.
+        logger.info("Embedded browser classes are not reachable: ${t.javaClass.name}: ${t.message}")
+        false
+      }
+
+    if (!classesReachable) return JcefStatus.CLASSES_MISSING
+
+    return try {
+      if (JcefSupportProbe.isSupported()) JcefStatus.AVAILABLE else JcefStatus.RUNTIME_UNSUPPORTED
+    } catch (t: Throwable) {
+      logger.warn("Embedded browser support check failed unexpectedly", t)
+      JcefStatus.RUNTIME_UNSUPPORTED
+    }
+  }
+}
+
+/**
+ * Isolates the single call that genuinely needs a JCEF type, so that [JcefAvailability] itself
+ * stays free of them. Compiled to its own class file and only ever invoked once the classes have
+ * been confirmed loadable, so resolving it cannot fail on the path where JCEF is absent.
+ */
+internal object JcefSupportProbe {
+  fun isSupported(): Boolean = com.intellij.ui.jcef.JBCefApp.isSupported()
+}
--- a/src/main/kotlin/io/snyk/plugin/ui/settings/HTMLSettingsPanel.kt
+++ b/src/main/kotlin/io/snyk/plugin/ui/settings/HTMLSettingsPanel.kt
@@ -6,7 +6,6 @@ import com.intellij.openapi.application.ModalityState
 import com.intellij.openapi.diagnostic.Logger
 import com.intellij.openapi.project.Project
 import com.intellij.openapi.util.Disposer
-import com.intellij.ui.jcef.JBCefApp
 import com.intellij.ui.jcef.JBCefBrowser
 import com.intellij.ui.jcef.JBCefClient
 import com.intellij.util.ui.UIUtil
@@ -19,6 +18,7 @@ import io.snyk.plugin.settings.handleDeltaFindingsChange
 import io.snyk.plugin.settings.handleReleaseChannelChange
 import io.snyk.plugin.ui.SnykBalloonNotificationHelper
 import io.snyk.plugin.ui.jcef.JCEFUtils
+import io.snyk.plugin.ui.jcef.JcefAvailability
 import io.snyk.plugin.ui.jcef.SaveConfigHandler
 import io.snyk.plugin.ui.jcef.ThemeBasedStylingGenerator
 import io.snyk.plugin.ui.toolwindow.SnykPluginDisposable
@@ -133,16 +133,18 @@ class HTMLSettingsPanel(private val project: Project) : JPanel(BorderLayout()), 
 
     registerForProject(project, this)
     Disposer.register(SnykPluginDisposable.getInstance(project), this)
-    initializePanel()
-    subscribeToCliDownloadEvents()
+    // Every method whose body names a com.intellij.ui.jcef type must stay uninvoked when the
+    // embedded browser is unavailable: bytecode verification is per-method and happens on first
+    // invocation, so an early return inside such a method would not save us from a LinkageError.
+    if (JcefAvailability.isAvailable()) {
+      initializePanel()
+      subscribeToCliDownloadEvents()
+    } else {
+      showEmbeddedBrowserUnavailableMessage()
+    }
   }
 
   private fun initializePanel() {
-    if (!JBCefApp.isSupported()) {
-      showJcefNotSupportedMessage()
-      return
-    }
-
     // Show loading state immediately to avoid blocking
     showLoadingMessage()
 
@@ -338,6 +340,7 @@ class HTMLSettingsPanel(private val project: Project) : JPanel(BorderLayout()), 
   }
 
   fun reloadFromLanguageServer() {
+    if (!JcefAvailability.isAvailable()) return
     // Fetch LS HTML in background to avoid blocking EDT
     ApplicationManager.getApplication().executeOnPooledThread {
       if (isDisposed) return@executeOnPooledThread
@@ -371,26 +374,16 @@ class HTMLSettingsPanel(private val project: Project) : JPanel(BorderLayout()), 
     }
   }
 
-  /**
-   * Returns true when the given raw HTML, after theme injection, differs from what the browser
-   * currently displays. Used to skip needless re-renders when the LS returns identical content.
-   */
-  internal fun themedHtmlDiffersFromCurrent(rawHtml: String): Boolean =
-    themedHtmlDiffers(rawHtml, lastThemedHtml)
-
-  private fun showJcefNotSupportedMessage() {
+  private fun showEmbeddedBrowserUnavailableMessage() {
     removeAll()
+    val reason = JcefAvailability.unavailableReason() ?: ""
     val label =
       JLabel(
-        "<html><center>JCEF is not supported on this platform.<br>" +
-          "The HTML settings panel cannot be displayed.</center></html>",
+        "<html><center>Snyk settings cannot be displayed in this IDE.<br><br>$reason</center></html>",
         SwingConstants.CENTER,
       )
     add(label, BorderLayout.CENTER)
-    SnykBalloonNotificationHelper.showWarn(
-      "JCEF is not supported on this platform. Using legacy settings dialog.",
-      project,
-    )
+    SnykBalloonNotificationHelper.showWarn(reason, project)
   }
 
   private fun showSettingsLoadError() {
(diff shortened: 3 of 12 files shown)


rrama (approved):
Approving quickly to get it out. Manual testing showed it worked. Will review later.

src/main/kotlin/io/snyk/plugin/ui/jcef/JcefAvailability.kt
```diff
@@ -0,0 +1,88 @@
...
+
+  private fun probe(): JcefStatus {
+    val classesReachable =
+      try {
+        // Deliberately not initialising the class: we only need to know it resolves.
+        Class.forName(PROBE_CLASS, false, JcefAvailability::class.java.classLoader)
+        true
+      } catch (t: Throwable) {
+        // The real failure mode is NoClassDefFoundError, which is an Error and not an Exception,
+        // so Throwable is the only catch that actually covers it.
```
rrama, line 65: Minor nit on the comment: "so Throwable is the only catch that actually covers it" is a slight overstatement. `NoClassDefFoundError` is an `Error`, so `catch (t: Error)` (or `catch (t: NoClassDefFoundError)`) would cover it too — the real point is just that catch `(e: Exception)` is insufficient. Could reword to something like:

> The real failure mode is NoClassDefFoundError, which is an Error not an Exception, so catching Exception wouldn't cover it — Throwable is used here to catch any failure.

basti-snyk: Well. I trust people will understand the message ;)

src/main/kotlin/io/snyk/plugin/ui/jcef/JcefAvailability.kt
```diff
@@ -0,0 +1,88 @@
...
+  }
+}
+
+/**
+ * Isolates the single call that genuinely needs a JCEF type, so that [JcefAvailability] itself
+ * stays free of them. Compiled to its own class file and only ever invoked once the classes have
+ * been confirmed loadable, so resolving it cannot fail on the path where JCEF is absent.
+ */
+internal object JcefSupportProbe {
+  fun isSupported(): Boolean = com.intellij.ui.jcef.JBCefApp.isSupported()
```
rrama, line 87: I am not a fan of having to remember to check this everywhere. I did a bit of spitballing with AI on some alternatives, then had it do a write-up (which I obviously haven't read myself 😉):

## Follow-up proposal: structurally enforce the "no JCEF types on the absent path" invariant
### Context
This PR correctly identifies that bytecode verification is per-method, so any method whose body names a `com.intellij.ui.jcef` type throws `LinkageError` on first invocation when JCEF is absent (2026.2 without the bundled Web Browser plugin). The fix guards each such method behind `JcefAvailability.isAvailable()`.
That works, but it's a **discipline-based** invariant: correctness depends on every current and future call site remembering to guard, and on the transitive reachability assumption that JCEF-touching classes are only reached via already-guarded panel constructors. The bug-catcher flag on `src/main/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanel.kt:274-286` calls this out — a single unguarded entry point (a hover provider, intention action, annotator, etc.) reintroduces the crash.
There's a fair amount of surface relying on this discipline. Beyond the guarded panels, these all name JCEF types in signatures and are safe *only* by virtue of being reachable through guarded constructors:
- `src/main/kotlin/io/snyk/plugin/ui/jcef/Utils.kt` (`JCEFUtils`, the browser factory)
- `src/main/kotlin/io/snyk/plugin/ui/jcef/SaveConfigHandler.kt`
- `src/main/kotlin/io/snyk/plugin/ui/jcef/ExecuteCommandBridge.kt`
- `src/main/kotlin/io/snyk/plugin/ui/jcef/TreeViewBridgeHandler.kt`
- `src/main/kotlin/io/snyk/plugin/ui/jcef/OpenFileLoadHandlerGenerator.kt`
- `src/main/kotlin/io/snyk/plugin/ui/jcef/GenerateAIFixHandler.kt`
- `src/main/kotlin/io/snyk/plugin/ui/jcef/IgnoreInFileHandler.kt`
- `src/main/kotlin/io/snyk/plugin/ui/jcef/ApplyAiFixEditHandler.kt`
- `src/main/kotlin/io/snyk/plugin/ui/jcef/SubmitIgnoreRequestHandler.kt`
- `src/main/kotlin/io/snyk/plugin/ui/jcef/ToggleDeltaHandler.kt`
The goal of this follow-up: turn the invariant from "remember to guard" into "the compiler / an arch test won't let you break it".
### Option A (primary): a JCEF-free browser wrapper interface
Introduce an interface — say `IJBCefBrowserWrapper` — that names **no** `com.intellij.ui.jcef` type in any of its members. It exposes only what callers actually need in non-JCEF terms:
```kotlin
interface IJBCefBrowserWrapper : Disposable {
  val component: JComponent
  fun loadHtml(html: String)
  // ...register-handler style operations phrased without JCEF types
}
```
The single concrete implementation is the *only* class whose bytecode names `JBCefBrowser`, and it holds it privately. The factory returns the interface, nullable:
```kotlin
// replaces JCEFUtils.getJBCefBrowserIfSupported(...): JBCefBrowser?
fun createBrowserIfAvailable(html: String, ...): IJBCefBrowserWrapper?
```
Why this enforces the invariant rather than merely documenting it:
- **Callers name only your own type.** Today, `getJBCefBrowserIfSupported(...): JBCefBrowser?` at `src/main/kotlin/io/snyk/plugin/ui/jcef/Utils.kt` leaks the raw JCEF type into every caller's locals — `private var jbCefBrowser: JBCefBrowser?` at `src/main/kotlin/io/snyk/plugin/ui/toolwindow/panels/HtmlTreePanel.kt:26` (assigned at `src/main/kotlin/io/snyk/plugin/ui/toolwindow/panels/HtmlTreePanel.kt:57`), `src/main/kotlin/io/snyk/plugin/ui/toolwindow/panels/SuggestionDescriptionPanel.kt:31` (assigned at `src/main/kotlin/io/snyk/plugin/ui/toolwindow/panels/SuggestionDescriptionPanel.kt:126`), and the local at `src/main/kotlin/io/snyk/plugin/ui/toolwindow/panels/SummaryPanel.kt:63`. Those fields/locals put `JBCefBrowser` into each panel's bytecode. Returning `IJBCefBrowserWrapper?` deletes those references entirely.
- **The "unavailable" case is the null.** A `null` return (equivalently `Optional.empty()`) *is* the Maybe — callers already have to null-check, and on the null path they never construct or touch anything JCEF. No separate `isAvailable()` guard needed at the call site; unavailability is encoded in the type.
- **The impl loads only on the supported path.** The concrete wrapper class is instantiated exclusively inside the factory, after the availability check, so the JVM only ever *verifies* it when JCEF is present.
- **The handler classes fold inside the boundary.** The `*Handler` / bridge classes above should either take the wrapper (or a non-JCEF callback) instead of `JBCefBrowserBase`, or stay strictly `internal` to the wrapper's package so no caller names them. Since they're all reachable only through the browser today, moving them behind the wrapper is natural.
Net effect: the per-call-site `JcefAvailability.isAvailable()` guards currently scattered across `src/main/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanel.kt` and `src/main/kotlin/io/snyk/plugin/ui/settings/HTMLSettingsPanel.kt` collapse into a single structural boundary. `JcefAvailability` / `JcefSupportProbe` (`src/main/kotlin/io/snyk/plugin/ui/jcef/JcefAvailability.kt`) stays as the internal probe the factory consults.
One caveat to verify during implementation: closures passed to the wrapper (e.g. a lambda that manipulates the browser) compile to separate synthetic methods, so a lambda body naming a JCEF type is only verified when the wrapper actually invokes it — i.e. only on the supported path. That's fine, but it means the enforcement boundary is "callers hold only the interface", not "callers pass no JCEF-typed lambdas". Keep lambda signatures JCEF-free too.
### Option B (complementary): an architecture test
Regardless of Option A, add a Konsist or ArchUnit test that fails the build if any type **outside** the `io.snyk.plugin.ui.jcef` package references `com.intellij.ui.jcef.*`. This makes the package boundary a hard rule and would immediately catch a future unguarded hover provider / intention action / annotator — exactly the follow-up the flag asks a reviewer to check manually. It's cheap and pairs well with Option A (the test guards the boundary the wrapper establishes). Worth checking first whether the repo already pulls in Konsist/ArchUnit.
### What I'd explicitly *not* do
A runtime "check-before-call" DI/middleware interface where the provider's own methods take or return raw JCEF types. That reintroduces the `com.intellij.ui.jcef` reference into the abstraction meant to remove it, and gives none of the per-method-verification guarantee — it's indirection without enforcement. The value of Option A comes specifically from the interface being JCEF-type-free, not from it being injected.
### Scope note
This is a refactor, not a behavior change — the current PR's guards are correct and ship-ready. This follow-up trades scattered runtime guards for a single compiler-checked boundary, and is best done as its own PR with the arch test landing first so the refactor can't regress.

src/main/kotlin/io/snyk/plugin/Utils.kt
```diff
@@ -152,6 +152,8 @@ private inline fun <reified T : Any> Project.serviceIfNotDisposed(): T? {
   return try {
     getService(T::class.java)
   } catch (t: Throwable) {
+    // Without this the real cause is lost and the caller only ever sees a downstream NPE.
+    logger.error("Could not instantiate service ${T::class.java.name}", t)
```
rrama, line 156: AI reported bug (unchecked by me):

> **Routine service lookups during project shutdown now raise error reports instead of failing quietly**
> 
> Every failed attempt to obtain one of the plugin's internal services is now reported at error level (`logger.error(...)` at `src/main/kotlin/io/snyk/plugin/Utils.kt:156`) inside a catch that was designed to quietly return nothing, so failures that happen routinely (for example while a project is closing) get surfaced instead of being ignored.
> Impact: During normal project close/disposal users can see spurious Snyk error notifications, and cancellation-type exceptions that the platform rethrows when handed to error-level logging can propagate to callers that expected a null result.
> 
> <details>
> <summary>Why catching-all-and-logging changes the graceful-fallback contract</summary>
> 
> `serviceIfNotDisposed` (`src/main/kotlin/io/snyk/plugin/Utils.kt:150-159`) is the shared helper behind `getSnykToolWindowPanel`, `getSnykCachedResults`, `getSnykTaskQueueService`, etc., and virtually every caller uses `?.` relying on a graceful `null` when the service cannot be produced. `getService` commonly throws while the container is being disposed (e.g. `AlreadyDisposedException`, which extends `CancellationException`). The broad `catch (t: Throwable)` now passes such throwables to `logger.error`. IntelliJ's logging framework treats control-flow / cancellation exceptions specially (it does not permit logging them and rethrows), and in unit-test mode `logger.error` throws outright. As a result the previously silent `return null` path can now either emit a user-visible error report (for ordinary exceptions) or rethrow (for cancellation/control-flow exceptions), changing the long-standing null-returning contract many call sites depend on during shutdown races.
> 
> </details>

src/main/resources/META-INF/plugin.xml
```diff
@@ -5,6 +5,17 @@
...
+    Since IntelliJ 2026.2 the embedded browser (com.intellij.ui.jcef.*) is no longer part of the
+    core platform: it ships as the bundled "Web Browser (JCEF)" plugin. Without this declaration
+    our PluginClassLoader cannot see those classes and the tool window fails to open.
+
+    It MUST stay optional: the com.intellij.modules.jcef id does not exist before 2025.3.1, and
+    pluginSinceBuild is 242, so a required depends would disable the plugin on 2024.2-2025.3.0.
+    The bundled plugin is also OS/arch specific, so it can be absent even on 2026.2 — every JCEF
+    call site therefore goes through JcefAvailability rather than trusting this declaration.
+  -->
+  <depends optional="true">com.intellij.modules.jcef</depends>
```
rrama, line 18: **The declaration works — but not for the reason this comment gives, and a v2 migration would silently break it.**

First, confirming the premise, because it's load-bearing and the docs actively contradict it. `<depends optional="true">` with no `config-file` **does** grant classloader visibility. In the `262` sources, `collectDirectDependenciesInOldFormat` (`platform/core-impl/src/com/intellij/ide/plugins/ModulesWithDependencies.kt`) builds the classloader parent graph from every `<depends>` element and never reads `isOptional`. `isOptional` is consulted only to decide whether a *missing* dependency disables the plugin (`PluginManagerCore.kt`), and `config-file` only controls the optional sub-descriptor's extension registration. Neither touches classloader parentage. Worth knowing that the docs disagree: [Plugin Configuration File § depends](https://plugins.jetbrains.com/docs/intellij/plugin-configuration-file.html) lists `config-file` as *required* when `optional="true"`, so we're relying on real-but-undocumented behaviour.

Now the part that matters. `com.intellij.ui.jcef.*` is **not** in the JCEF plugin's main module. From `plugins/jcef/plugin/resources/META-INF/plugin.xml` on branch `262`:

```xml
<id>com.intellij.modules.jcef</id>
<content namespace="jetbrains">
  <module name="intellij.libraries.jcef" loading="required"/>
  <module name="intellij.platform.ui.jcef"/>
  ...
</content>
```

`intellij.platform.ui.jcef` is a non-embedded content module shipped in its own jar (`lib/modules/intellij.platform.ui.jcef.jar`) with its **own** classloader. Under the v2 plugin model a consumer has to name it explicitly. This declaration works only because the **v1 `<depends>` path** carries a backward-compatibility branch that pulls in every content module of the target plugin:

```kotlin
// e.g. `.env` plugin in an old format and doesn't explicitly specify dependency on new extracted modules
if (dep is PluginMainDescriptor) {
  dependenciesCollector.addAll(dep.contentModules)
}
dependenciesCollector.add(dep)
```

The consequence: if anyone later migrates this descriptor to the v2 `<dependencies><plugin id="com.intellij.modules.jcef"/></dependencies>` block — a natural-looking modernisation, and the form JetBrains' own migration note shows — the implicit pull-in disappears and the `NoClassDefFoundError` comes straight back. The v2 form needs `<module name="intellij.platform.ui.jcef"/>` alongside the plugin reference.

Suggest the comment records that, since the trap is invisible otherwise.

Two smaller confirmations while I was in there: the "id does not exist before 2025.3.1" claim checks out — JetBrains state it in the [2026.2 compatibility announcement](https://platform.jetbrains.com/t/2026-2-is-coming-time-to-check-your-plugin-compatibility/4618), and `plugins/jcef/plugin/` genuinely does not exist on the `253` branch. And the OS/arch caveat is real: the bundled plugin's `<dependencies>` block carries an `<!-- OS/ARCH-DEPENDENCY-PLACEHOLDER -->` substituted at build time, which `PluginManagerCore.getUnfulfilledOsRequirement` / `getUnfulfilledCpuArchRequirement` then enforce. JetBrains also flag that JCEF does not load in `useInstaller = false` builds, which is likely to show up in our own Gradle/CI runs and won't mean anything is broken.

— AI generated comment

src/main/kotlin/io/snyk/plugin/ui/settings/HTMLSettingsPanel.kt
```diff
@@ -133,16 +133,18 @@ class HTMLSettingsPanel(private val project: Project) : JPanel(BorderLayout()),
...
-    initializePanel()
-    subscribeToCliDownloadEvents()
+    // Every method whose body names a com.intellij.ui.jcef type must stay uninvoked when the
+    // embedded browser is unavailable: bytecode verification is per-method and happens on first
+    // invocation, so an early return inside such a method would not save us from a LinkageError.
+    if (JcefAvailability.isAvailable()) {
+      initializePanel()
+      subscribeToCliDownloadEvents()
+    } else {
+      showEmbeddedBrowserUnavailableMessage()
```
rrama, line 143: **On a JCEF-less IDE the plugin becomes unconfigurable — even though a working legacy settings dialog is sitting right there.**

This is the one degradation path where a real fallback exists and we don't take it.

`SnykProjectSettingsConfigurable.createComponent()` picks between the HTML panel and the legacy Swing dialog on the registry flag alone:

```kotlin
useNewConfigDialog = isNewConfigDialogEnabled()
return if (useNewConfigDialog) {
  htmlSettingsPanel = HTMLSettingsPanel(project)
  htmlSettingsPanel!!
} else {
  snykSettingsDialog.getRootPanel()
}
```

`snyk.useNewConfigDialog` defaults to `true` (`plugin.xml:52-56`), so on 2026.2 without the bundled JCEF plugin the user is handed this panel, which now renders nothing but the "Snyk settings cannot be displayed in this IDE" message. Worse, `apply()` short-circuits with `modified.set(false)` and `reset()` returns early, so nothing can ever be saved — no token, no endpoint, no organisation, no CLI path. That isn't degraded functionality, it's an unusable plugin: a user in that state cannot even authenticate.

`snykSettingsDialog` is fully functional and already constructed in the same class, so the fix is roughly one line in the Configurable rather than anything in this file:

```kotlin
useNewConfigDialog = isNewConfigDialogEnabled() && JcefAvailability.isAvailable()
```

For fairness: the pre-PR message read "JCEF is not supported on this platform. **Using legacy settings dialog.**" — that promise was never actually implemented, so the gap predates this PR rather than being a regression. But this change removed the sentence instead of the gap, and since this PR is where the whole degradation strategy was defined, it's the natural place to close it.

— AI generated comment

src/main/kotlin/io/snyk/plugin/ui/toolwindow/SnykToolWindowPanel.kt
```diff
@@ -267,14 +273,30 @@ class SnykToolWindowPanel(val project: Project) : JPanel(), Disposable {
...
     summaryPanel.removeAll()
-    Disposer.register(this, summaryPanelContent)
-    summaryPanel.add(summaryPanelContent)
+    if (JcefAvailability.isAvailable()) {
+      val summaryPanelContent = SummaryPanel(project)
+      this.summaryPanelContent = summaryPanelContent
+      Disposer.register(this, summaryPanelContent)
+      summaryPanel.add(summaryPanelContent)
+    } else {
+      this.summaryPanelContent = null
```
rrama, line 283: **Summary strip goes silently blank here, unlike the other two regions.**

The tree area and the description area both get `embeddedBrowserUnavailablePanel()` when the browser is missing, but this branch runs `summaryPanel.removeAll()` and puts nothing back, so the user sees an unexplained empty strip sitting above a panel that *does* explain itself. Reusing `embeddedBrowserUnavailablePanel()` here would make all three consistent.

Cosmetic only — flagging it because the rest of the PR is deliberately careful about exactly this.

— AI generated comment

rrama (commented):
Post-merge review. I went through this against the IntelliJ `262` platform sources to check that the central premise holds — that `<depends optional="true">` really does grant classloader visibility to `com.intellij.ui.jcef.*`. It does, but via a mechanism worth writing down; details in the `plugin.xml` comment below.

Three findings I didn't see covered already. Deliberately **not** repeating: the `logger.error` / `serviceIfNotDisposed` severity issue (@rrama and the review bot both have it), the `mailto:`/custom-scheme restriction in `openInExternalBrowser`, and the `adjustHexBrightness` `require` throwing on unexpected colour formats.

Also checked and found **fine**, so nobody re-treads them:

- `htmlTreePanel` / `summaryPanelContent` can now be null, but every call site already uses `?.`
- the deleted `@TestOnly` helpers and `themedHtmlDiffersFromCurrent` have no remaining references; the unit-tested `themedHtmlDiffers` companion survives
- `onBeforePopup`'s four-argument signature is still current in `JetBrains/jcef` master, so the new popup handler really will fire against the `jcef-144` build bundled in 2026.2
- `pluginUntilBuild=262.*`, so the fix actually reaches 2026.2 users

— AI generated review

PR #886, opened 2026-08-18
fix: apply the selected issues-to-display value in legacy settings dialog [IDE-2477]

rrama:
### Description

`apply()` in `SnykProjectSettingsConfigurable.kt` compared `settingsStateService.issuesToDisplay` against `newDisplayIssuesSelection` but then assigned `newCliReleaseChannel` into it — a copy-paste slip from the sibling CLI-release-channel block just above. This meant the legacy Settings dialog's New/Total radio buttons never actually took effect.

Extracted the compare-and-assign into a standalone `applyDisplayIssuesSelection(settingsStateService, newDisplayIssuesSelection): Boolean` so the logic is unit-testable without instantiating the platform-heavy `Configurable`, and added coverage for both the changed and unchanged cases (there was previously zero test coverage on this code path, which is how the bug shipped unnoticed).

### Checklist

- [x] Read and understood the Code of Conduct and Contributing Guidelines.
- [x] Linted
- [ ] README.md updated, if user-facing (not user-facing)

### Screenshots / GIFs

N/A — internal logic fix, no UI change.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -100,3 +100,4 @@ docs/plan.md
 # Implementation plan working copies and generated diagrams (session scratchpads — source of truth is Confluence + brain)
 *_implementation_plan.md
 docs/diagrams/
+.verify-report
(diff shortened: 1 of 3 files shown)


rrama:
Legacy settings page, which is no longer supported and will be removed in IDE-2171.

PR #887, opened 2026-08-18
fix: move sendScanCommand LSP calls off the EDT [IDE-2478]

rrama:
### Description

Fixes IDE-2478 — the Snyk IntelliJ plugin freezing the whole IDE for 5-10s at a time on workspaces with many content roots.

Root cause: `sendScanCommand()` ran a chain of blocking LSP calls (`addContentRoots` → `ensureLanguageServerInitialized` (untimed lock + blocking network/CLI init) → `updateWorkspaceFolders` → synchronous LSP calls) inside `DumbService.runWhenSmart{}`, which executes on the EDT.

**Fix:** wrap the update in `runInBackground`, mirroring the existing `restart()` pattern.

Two follow-on issues surfaced during review and are fixed in the same commit:
- Moving `addContentRoots` off-EDT would make its internal content-root lookup race (only resolves synchronously when called from the EDT) and silently collapse to just `project.baseDir`. Fixed by threading the already EDT-resolved roots through a new `addContentRoots(project, roots)` overload, normalized the same way the original recompute path was.
- Backgrounding removed the EDT's implicit serialization between concurrent `sendScanCommand()` calls, and its "can't run mid-project-close" safety. Fixed with a dedicated lock and a disposal guard.

Known, deliberately out-of-scope limitations (documented in code comments): `restart()` and `SnykTaskQueueService.connectProjectToLanguageServer` retain the identical pre-existing off-EDT recompute race on much rarer paths than `sendScanCommand`; `sendScanCommand`'s pre-existing `notAuthenticated()` guard can still block the EDT if invoked while the LS isn't yet initialized (e.g. via `PreCommitHookHandler`) — unchanged by this diff, needs separate investigation.

### Checklist

- [x] Linted
- [ ] README.md updated, if user-facing (not user-facing)

### Screenshots / GIFs

N/A — internal threading fix, no UI change.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -100,3 +100,10 @@ docs/plan.md
 # Implementation plan working copies and generated diagrams (session scratchpads — source of truth is Confluence + brain)
 *_implementation_plan.md
 docs/diagrams/
+.verify-report
+.security-scan-progress
+.gitleaks-output.json
+.gitleaks-staged.json
+.security-scan-scope
+.snyk-secrets-output.err
+.snyk-secrets-output.json
--- a/src/main/kotlin/snyk/common/lsp/LanguageServerWrapper.kt
+++ b/src/main/kotlin/snyk/common/lsp/LanguageServerWrapper.kt
@@ -40,6 +40,7 @@ import java.util.concurrent.TimeoutException
 import java.util.concurrent.locks.ReentrantLock
 import java.util.logging.Level
 import java.util.logging.Logger.getLogger
+import kotlin.concurrent.withLock
 import org.apache.commons.lang3.SystemUtils
 import org.eclipse.lsp4j.ClientCapabilities
 import org.eclipse.lsp4j.ClientInfo
@@ -159,6 +160,14 @@ class LanguageServerWrapper(private val project: Project) : Disposable {
   lateinit var process: Process
 
   var isInitializing: ReentrantLock = ReentrantLock()
+
+  // Before this class backgrounded sendScanCommand's workspace-folder update, the whole call ran
+  // inside DumbService.runWhenSmart, which always executes on the single-threaded EDT — so
+  // concurrent sendScanCommand() calls (e.g. a token refresh and a user-triggered scan landing
+  // close together) were naturally serialized. Running that update in a background task loses that
+  // for free; this lock restores it so two overlapping calls can't interleave their
+  // configuredWorkspaceFolders mutation and didChangeWorkspaceFolders/executeCommand dispatch.
+  private val workspaceFolderUpdateLock = ReentrantLock()
 
   // Written by initialize() / the LS-termination listener on background threads and read
   // cross-thread (e.g. the settings panel poll and the non-forcing getConfigHtml check) outside the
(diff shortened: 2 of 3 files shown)


rrama:
## Context for reviewers: why this PR is bigger than the original triage suggestion

The automated triage comment on IDE-2478 proposed a one-line fix: wrap `sendScanCommand()`'s body in `runInBackground(...)` to get it off the EDT. That diagnosis was correct, but implementing it exposed three more issues, each found by verification and fixed in this same commit. Writing this up in full so reviewers aren't left guessing why a "wrap it in runInBackground" ticket turned into a 188-line diff.

### 1. The original one-liner would have silently dropped workspace folders

Moving `addContentRoots(project)` off the EDT means its internal `getContentRootVirtualFiles()` call — which resolves via `DumbService.runWhenSmart` — only executes synchronously when the *caller* is already on the EDT and in smart mode. Off the EDT, that assignment is deferred to a later EDT tick, so the code that reads the result immediately after would see the initial empty array and fall back to `arrayOf(project.baseDir)`. In other words: naively backgrounding this call would have silently collapsed a workspace with N content roots down to just the project's base directory on every scan trigger.

I confirmed this against the actual `intellij-community` platform source (`DumbServiceImpl`/`SmartModeScheduler`) rather than guessing.

**Fix:** added `addContentRoots(project, roots: Set<VirtualFile>? = null)`. `sendScanCommand` now resolves `roots` once on the EDT (before backgrounding) and passes them straight through, so the recompute — and the race — never happens on this path.

`restart()` and `SnykTaskQueueService.connectProjectToLanguageServer` have the identical pre-existing exposure (they already call `addContentRoots(project)` from a background thread today, unrelated to this change) — left untouched, since fixing them means touching call sites outside this ticket's scope. Called out in code comments.

### 2. The new overload bypassed content-root normalization

Adversarial review caught that the new `roots` parameter skipped the ancestor-folding / null-path filtering the original recompute path applied (`getContentRoots`'s inner logic). Zero live impact today (the only caller passes already-normalized roots), but it was a latent trap for any future caller.

**Fix:** extracted that logic into `normalizeContentRoots()`, applied it to both the original path and the new `roots` parameter.

### 3. Backgrounding silently dropped two safety properties the EDT gave for free

Before this fix, `sendScanCommand()`'s entire body ran inside `DumbService.runWhenSmart`, which always executes on the single-threaded EDT. That gave two properties nobody had to think about:
- Two overlapping `sendScanCommand()` calls (e.g. a token refresh and a user-triggered scan landing close together) were naturally serialized.
- The whole call ran atomically with respect to project disposal — it couldn't be interrupted by the IDE window closing mid-execution.

Backgrounding drops both. Code review caught the first; a follow-up adversarial pass caught the second.

**Fix:** a dedicated `workspaceFolderUpdateLock` restores the serialization (scoped only to `sendScanCommand`'s own background block — not extended to `restart()`/`connectProjectToLanguageServer`, which is a separate, pre-existing exposure). A `disposed || project.isDisposed` guard restores the disposal safety, matching the exact guard pattern already used elsewhere in this class (an earlier draft of this guard only checked `project.isDisposed`, missing the wrapper's own `disposed` flag — set independently on plugin unload — which two reviewers independently caught).

### 4. Test infrastructure needed real fixes too, not just new tests

Getting real end-to-end coverage of `sendScanCommand`'s `runWhenSmart → runInBackground` wiring meant mocking `com.intellij.openapi.progress.ProgressManager`. First attempt stubbed it via `applicationMock.getService(ProgressManager::class.java)`, following the existing pattern for `DumbService` in this test file — but `ProgressManager.getInstance()` caches its resolved instance in a static field, so that stub only worked for whichever test happened to run first in the JVM fork. I reproduced this empirically (ran two `ProgressManager`-touching tests together, got a `MockKException` from the wrong test's mock), then fixed it with `mockkStatic(ProgressManager::class)`, which intercepts the static call directly instead of depending on the real method's internals.

Separately, the first version of the disposal-guard test was vacuous — it asserted `executeCommand` was never called, but `sendFolderScanCommand`'s own unrelated membership check on an unseeded `configuredWorkspaceFolders` was already suppressing that call regardless of whether the new guard existed. Two independent review passes caught this. Fixed by seeding `configuredWorkspaceFolders` with a matching entry so the assertion is actually attributable to the guard — verified via TDD (removed the guard, confirmed exactly this one test fails; restored it, confirmed 88/88 green).

### Verification

Ran 5 rounds of 4-agent verification (semantic-analysis, adversarial review, security scan, code review) as changes accumulated — every finding above came from that process, not from me second-guessing myself after the fact. Full history is in the `.verify-report` this repo's commit tooling would have archived to brain, summarized here for anyone who doesn't have access to that.

Security: Snyk CLI auth wasn't available in the sandboxed verification agents' environment, so those rounds fell back to manual review + gitleaks (clean throughout). I authenticated separately afterward and ran a real `snyk code test` against the final `LanguageServerWrapper.kt` — 0 issues.

### What's deliberately *not* fixed here

- `restart()` and `SnykTaskQueueService.connectProjectToLanguageServer` still have the pre-existing off-EDT content-root recompute race (much rarer paths than `sendScanCommand`, would need a separate change to their call sites).
- `sendScanCommand()`'s `notAuthenticated()` guard still calls `ensureLanguageServerInitialized()` synchronously *before* reaching the backgrounded block — unchanged by this diff. If `sendScanCommand()` is ever invoked from the EDT while the LS isn't yet initialized (e.g. via `PreCommitHookHandler.beforeCheckin()`), that init call could still block the EDT. This needs someone to confirm `CheckinHandler.beforeCheckin()`'s actual threading contract before it's worth touching — flagging for a follow-up ticket rather than guessing and expanding this PR further.

